PR #19, opened 2018-08-02
ecp5: Write tiletype names in correct order

daveshah1:
Accidentally relied on Python dictionary ordering, which happened to work on some distros but not on others. This explicitly orders correctly instead and should fix all these issues.


q3k:
How about using https://docs.python.org/3.5/library/collections.html#collections.OrderedDict instead?

daveshah1:
It is a map from tile type names to sequential integer IDs. We want the dictionary to end up in the order of those IDs when generating the database, so imo the easiest solution is just to sort the dictionary by value at the end. (the IDs are also the order of insertion, but I'd rather not even rely on that)

PR #23, opened 2018-08-03
Making use of relative constraints 

daveshah1:
Not ready to be merged yet, still a bit rough around the edges and needs debugging to be taken out.

This PR replaces the current legalisation API with relative placement constraints being set in the packer and a generic legaliser for these constraints being used instead during placement. We also use "constraint distance" - the distance between current position of cells in a chain and satisfied relative constraints in that chain - as a SA metric


ice40/chains.h
```diff
@@ -1,7 +1,7 @@
 /*
  *  nextpnr -- Next Generation Place and Route
  *
- *  Copyright (C) 2018  David Shah <david@symbioticeda.com>
+ *  Copyright (C) 2018  David Shah <clifford@symbioticeda.com>
```
eddiehung: Make up your mind!

ice40/chains.cc
```diff
@@ -0,0 +1,287 @@
+/*
+ *  nextpnr -- Next Generation Place and Route
+ *
+ *  Copyright (C) 2018  David Shah <clifford@symbioticeda.com>
```
eddiehung: Copy-pasta fail here too

common/placer1.cc
```diff
@@ -225,7 +225,8 @@ class SAPlacer
             // Once cooled below legalise threshold, run legalisation and start requiring
             // legal moves only
             if (temp < legalise_temp && !require_legal) {
-                legalise_design(ctx);
+                legalise_relative_constraints(ctx);
+                // FIXME
```
eddiehung: What does this refer to?

common/placer1.cc
```diff
@@ -60,11 +60,11 @@ class SAPlacer
                 type_idx = bel_types.at(type);
             }
             if (int(fast_bels.size()) < type_idx + 1)
-                fast_bels.resize(type_idx + 1);
+            fast_bels.resize(type_idx + 1);
```
eddiehung: clangformat necessary?

common/placer1.cc
daveshah1: Obsolete, from before I put the legalise step in there again

PR #38, opened 2018-08-06
Restore original assign_budget behaviour (to distribute slack evenly)

eddiehung:
During my experimentation with slack redistribution, I inadvertently juggled around where the first `assign_budget` is called for ice40 from after packing (where there are no net delays) to after initial placement (where a net delay estimate exists). 

This effect would cause the port budget to be initialised according to estimate net delays based on the (poor) initial placement, even when slack redistribution is not enabled.

This PR reverts this behaviour.


daveshah1:
After testing this, I realise this breaks the timing estimates printed when slack_redist is disabled (because net delay is then not considered)

eddiehung:
Thanks for checking! I set off a benchmarking run which doesn't show the timing report.

PR #40, opened 2018-08-06
Fix budget overrides

eddiehung:
It appears that `Arch::getBudgetOverride()` wasn't being used correctly during STA. Builds on #38.

Also changed signature of `Arch::getBudgetOverride()` to `bool Arch::getBudgetOverride(const Netinfo*, const PortRef&, delay_t&)` to modify budget in place and return bool if this was done.

This appears to return a small QoR improvement:
On #38:
```
  | min | avg | max | success (/32)
master 100MHz | 50.88 | 55.04 | 60.28 | 32
master 70MHz | 50.4 | 54.97 | 59.13 | 31
master 60MHz | 51.16 | 54.97 | 59.06 | 32
master (default=12MHz) | 51.54 | 54.42 | 59.73 | 32
```

This:
```
  | min | avg | max | success (/32)
master 100MHz | 49.87 | 55.45 | 61 | 32
master 70MHz | 51.14 | 55.94 | 63.33 | 32
master 60MHz | 51.71 | 57.32 | 62.53 | 32
master (default=12MHz) | 50.17 | 56.25 | 62.26 | 31
```

### DIFF
--- a/common/timing.cc
+++ b/common/timing.cc
@@ -32,14 +32,15 @@ typedef std::map<int, unsigned> DelayFrequency;
 struct Timing
 {
     Context *ctx;
+    bool net_delays;
     bool update;
     delay_t min_slack;
     PortRefVector current_path;
     PortRefVector *crit_path;
     DelayFrequency *slack_histogram;
 
-    Timing(Context *ctx, bool update, PortRefVector *crit_path = nullptr, DelayFrequency *slack_histogram = nullptr)
-            : ctx(ctx), update(update), min_slack(1.0e12 / ctx->target_freq), crit_path(crit_path),
+    Timing(Context *ctx, bool net_delays, bool update, PortRefVector *crit_path = nullptr, DelayFrequency *slack_histogram = nullptr)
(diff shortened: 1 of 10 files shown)


ice40/arch.cc
```diff
@@ -637,17 +637,19 @@ std::vector<GroupId> Arch::getGroupGroups(GroupId group) const
...
-    if (driver.port == id_cout) {
+    if (driver.port == id_cout && sink.port == id_cin) {
         auto driver_loc = getBelLocation(driver.cell->bel);
         auto sink_loc = getBelLocation(sink.cell->bel);
         if (driver_loc.y == sink_loc.y)
-            return 0;
-        return 250;
+            budget = 0;
+        else
+            budget = 190;
```
daveshah1: If you need this value to be exactly right, it needs to vary per device - 190ps for HX1K/HX8K, 290ps for LP384/LP1K/LP8K and 560ps for UP5K.

common/timing.cc
```diff
@@ -32,36 +32,42 @@ typedef std::map<int, unsigned> DelayFrequency;
 struct Timing
```
daveshah1: You should add your name to the copyright header in timing.cc, given you've now written most of it!

Issue #32, opened 2018-08-04
Support ports forming both register and combinational inputs in cells


daveshah1:
The current timing visitor will stop at a port once it is determined to be a register input (`getPortClock != IdString()`). This doesn't work for ports that may be both, such as ECP5 slice inputs, or iCE40 LUT inputs with LUTCascade/carries.

Instead, it should keep looking for further paths if there are any more outputs with a direct timing path `getCellDelay` returning true.

Relevant code
```
    if (ctx->getPortClock(user.cell, user.port) != IdString()) {
        // At the end of a timing path (arguably, should check setup time
        // here too)
        value = slack / path_length;
        if (slack < min_slack) {
            min_slack = slack;
            if (crit_path)
                *crit_path = *current_path;
        }
    } else {
        // Default to the path ending here, if no further paths found
        value = slack / path_length;
        // Follow outputs of the user
        for (auto port : user.cell->ports) {
            if (port.second.type == PORT_OUT) {
                DelayInfo comb_delay;
                // Look up delay through this path
                bool is_path = ctx->getCellDelay(user.cell, user.port, port.first, comb_delay);
                if (is_path) {
                    NetInfo *net = port.second.net;
                    if (net) {
                        delay_t path_budget = follow_net(ctx, net, path_length, slack - comb_delay.maxDelay(), update,
                                                         min_slack, current_path, crit_path);
                        value = std::min(value, path_budget);
                    }
                }
            }
        }
    }
```

daveshah1:
This seems now to be fixed following various improvements to the timing code.

PR #41, opened 2018-08-07
Speedup assign_budget by walking design in topographical order

eddiehung:
Previously, `assign_budget` would walk recursively from each clocked output, through all its fanout paths, in order to determine the minimum slack of each path, before returning to distribute this path slack evenly.

This PR changes this by first generating a topographical ordering to walk the design (making the assumption it is a DAG, and there exists no combinatorial loops). Then it walks forward through the design in order to find the minimum slack of each path, and then backwards through the design to distribute this slack.

This should make `assign_budget` linear efficiency. The litmus test is that design03-up5k now completes in 4 minutes on my 1.4GHz machine.

```
Info: estimated Fmax = 7.72 MHz

Info: Slack histogram:
Info:  legend: * represents 48 endpoint(s)
Info:          + represents [1,48) endpoint(s)
Info: [-46196, -39808) |+
Info: [-39808, -33420) |+
Info: [-33420, -27032) |*+
Info: [-27032, -20644) |***+
Info: [-20644, -14256) |**+
Info: [-14256,  -7868) |+
Info: [ -7868,  -1480) |+
Info: [ -1480,   4908) |*+
Info: [  4908,  11296) |+
Info: [ 11296,  17684) |*+
Info: [ 17684,  24072) |***+
Info: [ 24072,  30460) |**+
Info: [ 30460,  36848) |**+
Info: [ 36848,  43236) |**+
Info: [ 43236,  49624) |***+
Info: [ 49624,  56012) |*****+
Info: [ 56012,  62400) |**********+
Info: [ 62400,  68788) |****************+
Info: [ 68788,  75176) |************+
Info: [ 75176,  81564) |************************************************************
Info: [ 81564,  87952) |+

real    3m58.400s
user    3m58.154s
sys     0m0.442s
```

### DIFF
--- a/common/timing.cc
+++ b/common/timing.cc
@@ -23,6 +23,7 @@
 #include <utility>
 #include "log.h"
 #include "util.h"
+#include <boost/range/adaptor/reversed.hpp>
 
 NEXTPNR_NAMESPACE_BEGIN
 
@@ -37,6 +38,14 @@ struct Timing
     PortRefVector current_path;
     PortRefVector *crit_path;
     DelayFrequency *slack_histogram;
+
+    struct TimingData {
+        TimingData() : max_arrival(), max_path_length(), min_remaining_budget() {}
+        TimingData(delay_t max_arrival) : max_arrival(max_arrival), max_path_length(), min_remaining_budget() {}
+        delay_t max_arrival;
+        unsigned max_path_length = 0;
+        delay_t min_remaining_budget;
+    };
 
     Timing(Context *ctx, bool update, PortRefVector *crit_path = nullptr, DelayFrequency *slack_histogram = nullptr)
             : ctx(ctx), update(update), min_slack(1.0e12 / ctx->target_freq), crit_path(crit_path),
@@ -53,8 +62,8 @@ struct Timing
             // If budget override is less than existing budget, then do not increment
             // path length
             int pl = path_length + 1;
-            auto budget = ctx->getBudgetOverride(net, usr, net_budget);
-            if (budget < net_budget) {
+            auto budget = net_budget;
+            if (ctx->getBudgetOverride(net, usr, budget) && budget < net_budget) {
                 net_budget = budget;
                 pl = std::max(1, path_length);
             }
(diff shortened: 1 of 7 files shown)


common/timing.cc
```diff
@@ -36,115 +37,263 @@ struct Timing
...
+                if (clock_domain != IdString()) {
+                    DelayInfo clkToQ;
+                    ctx->getCellDelay(cell.second.get(), clock_domain, o->name, clkToQ);
+                    topographical_order.emplace_back(o->net);
+                    net_data.emplace(o->net, TimingData{ clkToQ.maxDelay() });
+                }
+                else {
+                    // Also add I/O cells too
+                                                      // TODO(eddieh): More generic way of detecting PLLs
+                    if (is_io || cell.second->type == ctx->id("ICESTORM_PLL")) {
```
eddiehung: Need a more generic way of detecting PLLs -- shall I add another Arch function?

common/timing.cc
```diff
@@ -36,115 +37,263 @@ struct Timing
...
+            const auto net = queue.front();
+            queue.pop_front();
+
+            DelayInfo clkToQ;
+            for (auto &usr : net->users) {
+                auto clock_domain = ctx->getPortClock(usr.cell, usr.port);
+                for (auto& port : usr.cell->ports) {
+                    if (port.second.type == PORT_OUT && port.second.net) {
+                        // Skip if this is a clocked output (but allow non-clocked ones)
+                        if (clock_domain != IdString() && ctx->getCellDelay(usr.cell, clock_domain, port.first, clkToQ))
```
eddiehung: @daveshah1 This _should_ allow cells which have both clocked and unclocked outputs to be analysed which might fulfill #32 --- this was necessary because I found a case where an I3 -> COUT path wasn't being analysed since the FF was also present.

daveshah1:
Right now this is failing on designs that use internal oscillators (which are similar to PLLs), and also the DSPs (which don't have timing analysis implemented yet because it's a bit tricky, but I will fix that at some point), with "Assertion failure: port_fanin.empty()".

I am wondering if not having timing defined through a cell should just be allowed?

eddiehung:
That assertion is a sanity check and can be commented out, at the risk of obviously missing that path. I use timing arcs to determine if there is a dependency from that input pin to the output pin.

cliffordwolf:
Performance is looking good: https://nbviewer.jupyter.org/url/scratch.clifford.at/report_assign_budget_speedup.ipynb

daveshah1:
Closing as superseded by  #44

PR #44, opened 2018-08-08
Speed up budget allocator using topographical ordering and update cell timing API

daveshah1:


ice40/arch.cc
```diff
@@ -883,27 +891,74 @@ bool Arch::getCellDelay(const CellInfo *cell, IdString fromPort, IdString toPort
...
+    } else if (cell->type == id("ICESTORM_HFOSC")) {
+        if (port == id("CLKHF"))
+            return TMG_GEN_CLOCK;
+        return TMG_IGNORE;
+    } else if (cell->type == id_sb_gb) {
+        if (port == id_glb_buf_out)
+            return TMG_COMB_OUTPUT;
+        return TMG_COMB_INPUT;
+    }
+    return TMG_IGNORE;
```
eddiehung: It doesn't like you fall through to here from any known cell type.. should we call log_error() as a catch all?

daveshah1: Yeah, would just need to add SB_WARMBOOT then

daveshah1: Done

daveshah1:
Yes, this has been tested on ECP5 too

ice40/arch.cc
eddiehung: Do we need entries for all items in:
```
Info:            ICESTORM_LC:  4217/ 5280    79%
Info:           ICESTORM_RAM:    28/   30    93%
Info:                  SB_IO:    30/   96    31%
Info:                  SB_GB:     8/    8   100%
Info:           ICESTORM_PLL:     1/    1   100%
Info:            SB_WARMBOOT:     0/    1     0%
Info:           ICESTORM_DSP:     0/    8     0%
Info:         ICESTORM_HFOSC:     0/    1     0%
Info:         ICESTORM_LFOSC:     0/    1     0%
Info:                 SB_I2C:     0/    2     0%
Info:                 SB_SPI:     0/    2     0%
Info:                 IO_I3C:     0/    2     0%
Info:            SB_LEDDA_IP:     0/    1     0%
Info:            SB_RGBA_DRV:     0/    1     0%
Info:         ICESTORM_SPRAM:     4/    4   100%
```

daveshah1: No, some of those cells aren't really supported and I'd rather trap on them for now

PR #47, opened 2018-08-09
Use settings for placer1 and router1

mmicko:


common/settings.h
```diff
@@ -0,0 +1,57 @@
...
+class Settings
+{
+  public:
+    explicit Settings(Context *ctx) : ctx(ctx) {}
+
+    template <typename T> T get(const char *name, T defaultValue)
+    {
+        IdString id = ctx->id(name);
+        if (ctx->settings.find(id) != ctx->settings.end())
+            return boost::lexical_cast<T>(ctx->settings[id]);
```
eddiehung: Consider caching the iterator from the `find` above and reusing it here, rather than doing a new lookup

common/settings.h
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    template <typename T> void set(const char *name, T value)
+    {
+        IdString id = ctx->id(name);
+        if (ctx->settings.find(id) != ctx->settings.end()) {
+            ctx->settings[id] = value;
+            return;
+        }
+        ctx->settings.emplace(id, std::to_string(value));
```
eddiehung: Consider doing an `emplace` without checking for whether it was already present. `emplace` returns a pair with an iterator and a bool. If bool is false (meaning value wasn't inserted because it already exists) use the iterator to overwrite its value.

mmicko:
@eddiehung used emplace result. this is not a critical path but it is nicer this way

PR #48, opened 2018-08-10
Placer speedup

eddiehung:
A few low hanging optimisations to speed up the placer. On my machine down from 18K samples to 14K...
[design03.pdf](https://github.com/YosysHQ/nextpnr/files/2276516/design03.pdf)

### DIFF
--- a/ice40/arch_place.cc
+++ b/ice40/arch_place.cc
@@ -23,15 +23,17 @@
 #include "nextpnr.h"
 #include "util.h"
 
+#include <boost/range/iterator_range.hpp>
+
 NEXTPNR_NAMESPACE_BEGIN
 
-bool Arch::logicCellsCompatible(const std::vector<const CellInfo *> &cells) const
+bool Arch::logicCellsCompatible(const CellInfo** it, const size_t size) const
 {
     bool dffs_exist = false, dffs_neg = false;
     const NetInfo *cen = nullptr, *clk = nullptr, *sr = nullptr;
     int locals_count = 0;
 
-    for (auto cell : cells) {
+    for (auto cell : boost::make_iterator_range(it, it+size)) {
         NPNR_ASSERT(cell->belType == id_ICESTORM_LC);
         if (cell->lcInfo.dffEnable) {
             if (!dffs_exist) {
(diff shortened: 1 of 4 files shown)


ice40/arch_place.cc
```diff
@@ -71,12 +71,13 @@ bool Arch::logicCellsCompatible(const std::vector<const CellInfo *> &cells) cons
 bool Arch::isBelLocationValid(BelId bel) const
 {
     if (getBelType(bel) == id_ICESTORM_LC) {
-        std::vector<const CellInfo *> bel_cells;
+        static std::vector<const CellInfo *> bel_cells;
+        bel_cells.clear();
```
cliffordwolf: Why does this provide a speedup? Is it significant? I'm asking because this will prevent us from running the placer multiple times in parallel in multiple threads for different ctx, which is something that we might (for example) want to do when using ML or GA/EA to find the best settings.

daveshah1: Thinking some more about this, given there is an upper limit of 8 cells here a simple array on the stack plus a count might be even better

eddiehung: Sounds good. 

Non static vector means that it has to grow its storage (using whatever growth factor the compiler has, but certainly the first allocation won't be straight to 8) on the heap, and then freed, every single time this function is called from the inner loop of the placer. The previous profile I shared internally before this and other optimisations has `Arch::isBelLocationValid()` and `Arch::isValidBelForCell()` taking up 10+3% of total runtime, contributing to most of the 6% runtime spent inside `std::vector::push_back`. Now the latest profile above shows about the same proportion of time spent inside the two `Arch` functions (though overall runtime has dropped) but a much smaller proportion going to `emplace_back`.

eddiehung: This is just really low hanging fruit to get us a boost. I think `isBelLocationValid` should be rethought for speed, perhaps using incremental updates rather than recomputing the whole thing.

eddiehung:
I've reworked it to now use std::array.

ice40/arch_place.cc
eddiehung: I've reworked it to now use std::array.

PR #52, opened 2018-08-16
gui: improved FPGAViewWidget::paintGL() performance

msgctl:
Profiling revealed that memcpy() in QOpenGLBuffer::allocate() had been taking
the most time during paintGL() calls. I've been able to take the CPU usage
down to about 1/4 of its previous values by caching elements in VBOs and
updating them only after subsequent calls to renderGraphicElement().

Signed-off-by: Mateusz Zalega <mateusz@appliedsourcery.com>

### DIFF
--- a/common/nextpnr.h
+++ b/common/nextpnr.h
@@ -157,10 +157,21 @@ struct GraphicElement
 
     enum style_t
     {
+        STYLE_GRID,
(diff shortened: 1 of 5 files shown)


msgctl:
I'll need to work on common/nextpnr.h.

PR #56, opened 2018-08-19
ice40: make PLL packing more robust

### LINKED ISSUE #55: nextpnr-ice40 crashes when using pll block
I am using a PLL block on iCE40up5k. This causes nextpnr to crash with the following error:
```
Info: Packing constants..
Info: Promoting globals..
Info: Packing IOs..
Info: Packing LUT-FFs..
Info: Packing non-LUT FFs..
Info: Packing carries..
Info: Packing RAMs..
Info: Packing special functions..
terminate called after throwing an instance of 'std::out_of_range'
  what():  _Map_base::at
../main.mk:15: recipe for target 'pll_uart_mirror.asc' failed
make: *** [pll_uart_mirror.asc] Aborted (core dumped)
```

I have published the design that is triggering the error here: https://github.com/icebreaker-fpga/icebreaker-examples/tree/master/pll_uart

To reproduce the error just run `make USE_NEXTPNR=1` in the linked directory. If I am "just doing something wrong" please let me know too. :smile:

q3k:
- Fix the SDO port definition on the iCE40 PLL packed cell.
 - Warn instead of crashing when the user design contains a port that is not present in the PLL packed cell.
 - Style fixes in PLL packing logging.

Fixes #55 .

### DIFF
--- a/ice40/cells.cc
+++ b/ice40/cells.cc
@@ -28,6 +28,7 @@ NEXTPNR_NAMESPACE_BEGIN
 void add_port(const Context *ctx, CellInfo *cell, std::string name, PortType dir)
 {
     IdString id = ctx->id(name);
+    NPNR_ASSERT(cell->ports.count(id) == 0);
     cell->ports[id] = PortInfo{id, nullptr, dir};
 }
 
@@ -237,7 +238,7 @@ std::unique_ptr<CellInfo> create_ice_cell(Context *ctx, IdString type, std::stri
 
         add_port(ctx, new_cell.get(), "SCLK", PORT_IN);
         add_port(ctx, new_cell.get(), "SDI", PORT_IN);
-        add_port(ctx, new_cell.get(), "SDI", PORT_OUT);
+        add_port(ctx, new_cell.get(), "SDO", PORT_OUT);
 
         add_port(ctx, new_cell.get(), "LOCK", PORT_OUT);
         add_port(ctx, new_cell.get(), "PLLOUT_A", PORT_OUT);
--- a/ice40/pack.cc
+++ b/ice40/pack.cc
@@ -733,7 +733,7 @@ static void pack_special(Context *ctx)
 
                 if (pi.name == ctx->id("PACKAGEPIN")) {
                     if (!is_pad) {
-                        log_error("  PLL '%s' has a PACKAGEPIN but is not a PAD PLL", ci->name.c_str(ctx));
+                        log_error("PLL '%s' has a PACKAGEPIN but is not a PAD PLL", ci->name.c_str(ctx));
                     } else {
                         // We drop this port and instead place the PLL adequately below.
                         pad_packagepin_net = port.second.net;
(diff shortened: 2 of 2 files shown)


ice40/pack.cc
```diff
@@ -793,13 +797,13 @@ static void pack_special(Context *ctx)
                         packagepin_cell->ports.erase(pll_packagepin_driver.port);
                     }
 
-                    log_info("  constrained '%s' to %s\n", packed->name.c_str(ctx), ctx->getBelName(bel).c_str(ctx));
+                    log_info("Constrained '%s' to %s\n", packed->name.c_str(ctx), ctx->getBelName(bel).c_str(ctx));
```
cliffordwolf: A lot of code in nextpnr uses the convention of log "headers" with capitalized initial letter and ending in `..`, followed by detail messages indented and with lowercase initial letter. This is such a detail message. `Packing special functions..` above is the header. This change breaks this convention. Why do that?

(Same with the other `log_info` changes below. Un-indenting the erros is okay imo.)

ice40/pack.cc
```diff
@@ -743,9 +743,13 @@ static void pack_special(Context *ctx)
                 }
                 if (pi.name == ctx->id("REFERENCECLK")) {
                     if (!is_core)
-                        log_error("  PLL '%s' has a REFERENCECLK but is not a CORE PLL", ci->name.c_str(ctx));
+                        log_error("PLL '%s' has a REFERENCECLK but is not a CORE PLL", ci->name.c_str(ctx));
                 }
 
+                if (packed->ports.count(ctx->id(newname)) == 0) {
+                    log_warning("PLL '%s' has unknown port '%s'\n", ci->name.c_str(ctx), pi.name.c_str(ctx));
```
daveshah1: This doesn't seem right to me at all - given this is guaranteed to lead to a broken design if the port is used, it should be a warning only if the port is unused (connected net is null). Otherwise it should remain an error (perhaps unless --force is set?)

ice40/cells.cc
```diff
@@ -237,7 +237,7 @@ std::unique_ptr<CellInfo> create_ice_cell(Context *ctx, IdString type, std::stri
 
         add_port(ctx, new_cell.get(), "SCLK", PORT_IN);
         add_port(ctx, new_cell.get(), "SDI", PORT_IN);
-        add_port(ctx, new_cell.get(), "SDI", PORT_OUT);
```
daveshah1: Maybe you could add an assertion to `add_port` to catch this typo in the future?

ice40/pack.cc
q3k: Hm, that's more or less the style I was trying to achieve, but I noticed that these 'detail' messages had no 'header' to hang off of, apart from the 'Packing specials' one. Let me fix this up.

ice40/cells.cc
q3k: Done.

ice40/pack.cc
q3k: Done.

PR #76, opened 2018-09-18
Add needed PLLOUTGLOBAL ports and mapped it

mmicko:


q3k:
1) Please add some warning that if the user uses PLLOUTGLOBAL

q3k:
1) Please add a warning when a user uses PLLOUTGLOBALx saying that the functionality is stop-gap and does not actually use the global clock output of the PLL.
2) Please add an error if the user uses both PLLOUTx and PLLOUTGLOBALx at the same time.

PR #93, opened 2018-10-21
Gui changes

mmicko:
- make grid only cover device itself
- fix for ice40 grid dimensions
- split device tree to multiple tabs/trees
- made bel tree available on all arch, since it use standard api

### DIFF
--- a/gui/designwidget.cc
+++ b/gui/designwidget.cc
@@ -51,8 +51,7 @@ DesignWidget::DesignWidget(QWidget *parent) : QWidget(parent), ctx(nullptr)
     tabWidget = new QTabWidget();
 
     // Add tree view
-    for(int i=0;i<6;i++)
-    {
+    for (int i = 0; i < 6; i++) {
         treeView[i] = new TreeView();
         treeModel[i] = new TreeModel::Model();
         treeView[i]->setModel(treeModel[i]);
@@ -96,7 +95,8 @@ DesignWidget::DesignWidget(QWidget *parent) : QWidget(parent), ctx(nullptr)
         history_index = 0;
         auto h = history.at(history_index);
         selectionModel[h.first]->setCurrentIndex(h.second, QItemSelectionModel::ClearAndSelect);
-        if (tabWidget->currentIndex()!=h.first) tabWidget->setCurrentIndex(h.first);
+        if (tabWidget->currentIndex() != h.first)
+            tabWidget->setCurrentIndex(h.first);
         updateButtons();
     });
 
@@ -108,7 +108,8 @@ DesignWidget::DesignWidget(QWidget *parent) : QWidget(parent), ctx(nullptr)
         history_index--;
         auto h = history.at(history_index);
         selectionModel[h.first]->setCurrentIndex(h.second, QItemSelectionModel::ClearAndSelect);
-        if (tabWidget->currentIndex()!=h.first) tabWidget->setCurrentIndex(h.first);
+        if (tabWidget->currentIndex() != h.first)
+            tabWidget->setCurrentIndex(h.first);
         updateButtons();
     });
 
@@ -120,7 +121,8 @@ DesignWidget::DesignWidget(QWidget *parent) : QWidget(parent), ctx(nullptr)
         history_index++;
         auto h = history.at(history_index);
         selectionModel[h.first]->setCurrentIndex(h.second, QItemSelectionModel::ClearAndSelect);
-        if (tabWidget->currentIndex()!=h.first) tabWidget->setCurrentIndex(h.first);
+        if (tabWidget->currentIndex() != h.first)
+            tabWidget->setCurrentIndex(h.first);
         updateButtons();
     });
 
@@ -132,7 +134,8 @@ DesignWidget::DesignWidget(QWidget *parent) : QWidget(parent), ctx(nullptr)
         history_index = int(history.size() - 1);
         auto h = history.at(history_index);
         selectionModel[h.first]->setCurrentIndex(h.second, QItemSelectionModel::ClearAndSelect);
-        if (tabWidget->currentIndex()!=h.first) tabWidget->setCurrentIndex(h.first);
+        if (tabWidget->currentIndex() != h.first)
+            tabWidget->setCurrentIndex(h.first);
         updateButtons();
     });
 
@@ -143,8 +146,7 @@ DesignWidget::DesignWidget(QWidget *parent) : QWidget(parent), ctx(nullptr)
         history_index = -1;
         history.clear();
         int num = tabWidget->currentIndex();
-        if (selectionModel[num]->selectedIndexes().size()> 0)
-        {
+        if (selectionModel[num]->selectedIndexes().size() > 0) {
             QModelIndex index = selectionModel[num]->selectedIndexes().at(0);
             if (index.isValid()) {
                 ElementType type = treeModel[num]->nodeFromIndex(index)->type();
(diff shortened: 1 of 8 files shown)


gui/designwidget.cc
```diff
@@ -487,8 +487,14 @@ void DesignWidget::onClickedBel(BelId bel, bool keep)
         Q_EMIT selected(getDecals(ElementType::BEL, ctx->getBelName(bel)), keep);
     }
     int index = getIndexByElementType(ElementType::BEL);
-    if (tabWidget->currentIndex() != index)
+    if (!keep)
+    {
+        for(int i=0;i<6;i++)
```
msgctl: What's 6? I'd rather have a constant with an appropriate scope and a name that would help me understand what's happening here.

msgctl: …can it be inferred from *selectionModel*?

gui/designwidget.cc
```diff
@@ -528,6 +539,11 @@ void DesignWidget::onClickedPip(PipId pip, bool keep)
     }
     
     int index = getIndexByElementType(ElementType::PIP);
+    if (!keep)
+    {
+        for(int i=0;i<6;i++)
```
msgctl: Coding style is inconsistent with rest of the codebase (space after ';' and before '(').

msgctl: Oh, there's 'clangformat' later in this patchset. Nevermind…

gui/designwidget.cc
```diff
@@ -331,8 +329,18 @@ void DesignWidget::updateTree()
     {
         std::lock_guard<std::mutex> lock_ui(ctx->ui_mutex);
         std::lock_guard<std::mutex> lock(ctx->mutex);
-        getTreeByElementType(ElementType::CELL)->updateCells(ctx);
-        getTreeByElementType(ElementType::NET)->updateNets(ctx);
+
+        std::vector<IdString> cells;
```
msgctl: std::transform() could yield better results due to reallocations in std::vector, but I'm not sure about this. Moreover it might be better to build these vectors before taking a mutex if there's any stuttering visible that's due to updateTree() calls.

msgctl: profile first, optimize later - there might be no need for that

gui/treemodel.cc
```diff
@@ -161,35 +161,13 @@ void Model::loadData(std::unique_ptr<Item> data)
     endResetModel();
 }
 
-void Model::updateCells(Context *ctx)
+void Model::updateElements(Context *ctx, std::vector<IdString> elements)
 {
     if (!ctx)
```
msgctl: Unrelated: I might never get used to having pointers instead of references here… I don't see the point.

gui/designwidget.cc
```diff
@@ -265,11 +265,49 @@ void DesignWidget::newContext(Context *ctx)
         std::lock_guard<std::mutex> lock_ui(ctx->ui_mutex);
         std::lock_guard<std::mutex> lock(ctx->mutex);
         
-        getTreeByElementType(ElementType::BEL)->loadContext(ElementType::BEL, ctx);
-        getTreeByElementType(ElementType::WIRE)->loadContext(ElementType::WIRE, ctx);
-        getTreeByElementType(ElementType::PIP)->loadContext(ElementType::PIP, ctx);
-        getTreeByElementType(ElementType::CELL)->loadContext(ElementType::CELL, ctx);
-        getTreeByElementType(ElementType::NET)->loadContext(ElementType::NET, ctx);
+        {
+            std::map<std::pair<int, int>, std::vector<BelId>> belMap;
```
msgctl: I think we could use a typedef for the pair here. I don't immediately follow this code.

msgctl: It doesn't even have to be a global scope typedef.

gui/designwidget.cc
```diff
@@ -265,11 +265,49 @@ void DesignWidget::newContext(Context *ctx)
...
         std::lock_guard<std::mutex> lock(ctx->mutex);
         
-        getTreeByElementType(ElementType::BEL)->loadContext(ElementType::BEL, ctx);
-        getTreeByElementType(ElementType::WIRE)->loadContext(ElementType::WIRE, ctx);
-        getTreeByElementType(ElementType::PIP)->loadContext(ElementType::PIP, ctx);
-        getTreeByElementType(ElementType::CELL)->loadContext(ElementType::CELL, ctx);
-        getTreeByElementType(ElementType::NET)->loadContext(ElementType::NET, ctx);
+        {
+            std::map<std::pair<int, int>, std::vector<BelId>> belMap;
+            for (auto bel : ctx->getBels()) {
```
msgctl: *auto &bel* ?

gui/designwidget.cc
```diff
@@ -265,11 +265,49 @@ void DesignWidget::newContext(Context *ctx)
...
+                auto loc = ctx->getBelLocation(bel);
+                belMap[std::pair<int, int>(loc.x, loc.y)].push_back(bel);
+            }
+            auto belGetter = [](Context *ctx, BelId id) { return ctx->getBelName(id); };
+            
+            getTreeByElementType(ElementType::BEL)->loadData(std::unique_ptr<TreeModel::ElementXYRoot<BelId>>(
+                    new TreeModel::ElementXYRoot<BelId>(ctx, belMap, belGetter, ElementType::BEL)));
+        }
+
+#ifdef ARCH_ICE40
```
msgctl: I'd rather have a proper interface for architecture-specific UI implementation, but I understand that this would involve refactoring.

gui/designwidget.cc
```diff
@@ -46,15 +46,29 @@ void TreeView::mouseMoveEvent(QMouseEvent *event)
...
     // Add tree view
-    treeView = new TreeView();
-    treeModel = new TreeModel::Model();
-    treeView->setModel(treeModel);
-    treeView->setContextMenuPolicy(Qt::CustomContextMenu);
-    treeView->setSelectionMode(QAbstractItemView::ExtendedSelection);
-    treeView->viewport()->setMouseTracking(true);
+    for(int i=0;i<6;i++)
+    {
+        treeView[i] = new TreeView();
```
msgctl: Instead of managing *treeView* by hand I'd rather have a vector of unique_ptrs here - that's less code and fewer mistakes. TreeView instances could be created with *std::make_unique<TreeView>()*;

msgctl: It looks like the same applies to *treeModel*.

msgctl: I might be wrong - I'm not sure how Qt manages these objects…

gui/treemodel.cc
```diff
@@ -166,50 +166,59 @@ void Model::loadContext(Context *ctx)
...
-            belMap[std::pair<int, int>(loc.x, loc.y)].push_back(bel);
+        if (type==ElementType::BEL)
+        {
+            std::map<std::pair<int, int>, std::vector<BelId>> belMap;
+            for (auto bel : ctx->getBels()) {
+                auto loc = ctx->getBelLocation(bel);
+                belMap[std::pair<int, int>(loc.x, loc.y)].push_back(bel);
+            }
+            auto belGetter = [](Context *ctx, BelId id) { return ctx->getBelName(id); };
+            root_ = std::unique_ptr<ElementXYRoot<BelId>>(
```
msgctl: std::make_unique would make it concise

gui/designwidget.cc
mmicko: yeah will use proper typedef for this

gui/designwidget.cc
msgctl: there's also *const auto &*

daveshah1: BelId is supposed to always be small enough to be used as a value not a reference (32 bits for ice40 and 64 bits for ecp5)

gui/treemodel.cc
q3k: ++

PR #103, opened 2018-11-03
WIP: Timing constraints API, multiple clock domains

daveshah1:
Implemented
 - [X] Timing API changes for common and ice40
 - [X] Separate analysis of different clock domains

Not yet implemented
 - [ ] Changes to ECP5 or generic architectures
 - [ ] Python bindings for constraints
 - [ ] Consideration of clock _constraints_ in multiple clock domain analysis
 - [ ] Implementation of other constraints in analysis

### DIFF
--- a/docs/archapi.md
+++ b/docs/archapi.md
@@ -455,11 +455,17 @@ Cell Delay Methods
 Returns the delay for the specified path through a cell in the `&delay` argument. The method returns
 false if there is no timing relationship from `fromPort` to `toPort`.
 
-### TimingPortClass getPortTimingClass(const CellInfo *cell, IdString port, IdString &clockPort) const
+### TimingPortClass getPortTimingClass(const CellInfo *cell, IdString port, int &clockInfoCount) const
 
 Return the _timing port class_ of a port. This can be a register or combinational input or output; clock input or
-output; general startpoint or endpoint; or a port ignored for timing purposes. For register ports, clockPort is set
-to the associated clock port.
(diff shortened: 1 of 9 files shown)


docs/archapi.md
```diff
@@ -398,6 +398,10 @@ actual penalty used is a multiple of this value (i.e. a weighted version of this
 
 Convert an `delay_t` to an actual real-world delay in nanoseconds.
 
+### DelayInfo getDelayFromNS(float v) const
+
+Convert a real-world delay in nanoseconds to a DelayInfo with equal min/max rising/falling values.
```
cliffordwolf: Why not simply return a delay_t?

daveshah1: We use DelayInfo in clock constraints, etc - but architectures currently have no API to create one. The alternative would be to use something other than DelayInfo in all those cases.

cliffordwolf: Ah, I see, that makes sense. Maybe we should add some std constructors to DelayInfo then. (But no need to delay this PR for that imo.)

eddiehung:
What's the interaction between clock constraints and the `--freq` command? It doesn't look like they are unified, and IMO `--freq` should set a clock constraint underneath the hood, so that this is so

daveshah1:
As it stands, clock constraints are applied to specific nets whereas `--freq` is global to a context and used for nets without a clock constraint.

eddiehung:
> As it stands, clock constraints are applied to specific nets whereas `--freq` is global to a context and used for nets without a clock constraint.
I don't think this is the right thing to do, though I'm okay with merging this PR without resolving this immediately.

PR #107, opened 2018-11-11
Major rewrite of "router1"

cliffordwolf:
Summary of changes:
- Add two new arch APIs: `getConflictingWireWire` and `getConflictingPipWire`
- Major rewrite of `router1`

This new router now uses finer-grain ripups and has no non-ripup phase anymore. It uses arcs as unit of work and always routes the unrouted arc with the worst budget-to-estimate ratio first.

This does not yet do timing-based ripups and restarts, but that will be easy to add once this PR is merged.


common/router1.cc
```diff
@@ -694,302 +736,265 @@ Router1Cfg::Router1Cfg(Context *ctx) : Settings(ctx)
...
-            PipId pip = router.visited.at(cursor).pip;
-            (*route)[cursor] = pip;
-            if (pip == PipId())
-                break;
-            cursor = getPipSrcWire(pip);
-        }
-    }
-
-    return true;
+    // FIXME
```
eddiehung: Context: it appears that this is only called by the fuzzer (e.g. `ice40DelayFuzzerMain`) meaning it is not critical for this PR

docs/faq.md
```diff
@@ -38,7 +38,94 @@ For nextpnr we are using the following terminology.
...
+
+### Delay Estimates
+
+Each architecture must implement a `estimateDelay()` method that estimates the expected delay for a path from given `src` to `dst` wires.
+*It is very important that this method slightly overestimates the expected delay.* Furthermore, it should overestimate the expected delay
+by a slightly larger margin for longer paths than for shorter paths. Otherwise there will be performance issues with the router.
+
+The delays estimates returned by that method should also be as fine-grain as possible. It definitely pays off to spend some time improving the `estimateDelay()`
+for your architecture once implementing small designs work.
+
```
eddiehung: It would be good to describe the difference between `predictDelay` and `estimateDelay` -- my understanding is the former is used in placement, the latter in routing. I would recommend changing their names, as they are too similar, perhaps: `fastEstimateDelay()` and `slowEstimateDelay()`?
Again, not critical for this PR, but dropping this comment here before I forget about it...

cliffordwolf: I was thinking maybe `predictDelay() -> estimateArcDelay()`.

cliffordwolf: Arc is one of the words on our nomenclature list with exactly this meaning. https://github.com/YosysHQ/nextpnr/blob/master/docs/faq.md If the name clash is a problem then we should change our nomenclature. In either case I think it would be a good idea to use whatever word we assign to this concept.

eddiehung: Sorry, I'm mistaking Arc for "Pip". `estimateArcDelay` is good.

PR #123, opened 2018-11-16
ice40/bitstream: Convert to UNIX line endings

smunaut:


mmicko:
Thanks for pointing out will check if git caused issue or editor did.

PR #130, opened 2018-11-19
common/placer1: In random pick, only use grid if there is more than 64 BELs

### LINKED ISSUE #127: infinite loop in random_bel_for_cell() while placing rare resources
Attached json is a small design (VGA character display and SPI interface to VRAM) that causes nextpnr-ice40 to get stuck forever.

[stuck.tar.gz](https://github.com/YosysHQ/nextpnr/files/2592066/stuck.tar.gz)

`$ nextpnr-ice40 --package sg48 --up5k --pcf top.pcf --asc top.asc --json top.json`

Discovered after resolving https://github.com/YosysHQ/nextpnr/issues/126 

May need fixes from @smunaut to get far enough to get stuck.

smunaut:
If you have a large grid and very few BELs of a given type, picking a
random grid location yields very little odds of finding a BEL of that
type.

So for those, just put all of them at (0,0) and do a true random pick.

Fixes #127 

### DIFF
--- a/common/placer1.cc
+++ b/common/placer1.cc
@@ -51,15 +51,20 @@ class SAPlacer
     {
         int num_bel_types = 0;
         for (auto bel : ctx->getBels()) {
+            IdString type = ctx->getBelType(bel);
+            if (bel_types.find(type) == bel_types.end()) {
+                bel_types[type] = std::tuple<int, int>(num_bel_types++, 1);
+            } else {
+                std::get<1>(bel_types.at(type))++;
+            }
+        }
+        for (auto bel : ctx->getBels()) {
             Loc loc = ctx->getBelLocation(bel);
             IdString type = ctx->getBelType(bel);
-            int type_idx;
-            if (bel_types.find(type) == bel_types.end()) {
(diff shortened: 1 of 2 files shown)


common/placer1.cc
```diff
@@ -51,15 +51,20 @@ class SAPlacer
...
-                type_idx = bel_types.at(type);
+                std::get<1>(bel_types.at(type))++;
             }
+        }
+        for (auto bel : ctx->getBels()) {
+            Loc loc = ctx->getBelLocation(bel);
+            IdString type = ctx->getBelType(bel);
+            int type_idx = std::get<0>(bel_types.at(type));
+            int type_cnt = std::get<1>(bel_types.at(type));
+            if (type_cnt < 64)
```
daveshah1: Can you make this magic number a constant somewhere? This might even be something that the arch API should supply. 64 seems a bit large to me, as this will also include block RAM - and DSPs on the ECP5.

smunaut: Fixed with `Placer1Cfg.minBelsForGridPick`

Issue #137, opened 2018-11-21
Not meeting timing should be at least a warning


whitequark:
Right now `--freq` seems to be essentially advisory, and nextpnr would gladly make a design that's half as slow as requested, and not even display a warning. This is particularly an issue with `--quiet`.

mithro:
I would say this should be an error!?

mithro:
A `--allow-timing-failure` could make this a warning.

cliffordwolf:
It's not an error in any of the vendor tools I know so no, it should not be an error.

daveshah1:
Following further discussion with @whitequark, it seems like the best approach is for failing timing to be a non-fatal error that will proceed to bitstream generation, so PnR time is not wasted; but produce a non-zero return code at the end.

PR #138, opened 2018-11-21
Tidy up logging code, add log file support, make timing failures non-fatal errors

### LINKED ISSUE #136: A command-line option to save a log file
In Glasgow, `--quiet` is especially useful, because running the test suite will synthesize and place dozens of designs, and having all that on stderr is just not convenient. On the other hand, it is still often useful to see the P&R output.

An option like `--log`, analogous to Yosys' `-L`, would be ideal for this. Migen already runs Yosys with `-q -L` for similar reasons.

daveshah1:
Fixes #136 and #137 

Adds the `--log`/`-l` option, similar to Yosys, to write a log file which contains all messages even if `-q` is set to limit what is printed to the console.

This also attempts to make the log code more orthogonal, adding the concept of log levels and removing some ugliness such as vectors of both C file pointers and C++ streams.

Timing failures are now printed as warnings in post-route timing analysis.

### DIFF
--- a/common/command.cc
+++ b/common/command.cc
@@ -40,7 +40,7 @@
 
 NEXTPNR_NAMESPACE_BEGIN
 
-CommandHandler::CommandHandler(int argc, char **argv) : argc(argc), argv(argv) { log_files.push_back(stdout); }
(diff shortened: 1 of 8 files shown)


whitequark:
Also tested here with Migen, works great.

PR #149, opened 2018-11-26
Fixes for global promotion

### LINKED ISSUE #148: problems with SB_PLL40_2_PAD and promote_globals()
Observeable on https://github.com/swetland/cpu16/ as of commit ffc013add5208852276d815ab3af5dca26684764

foss toolchain(s) are ToT as of today:
```
icestorm 9671b760f84ca4006f0ef101a3e3b201df4eabb5
nextpnr ff978570b162844de61dfa4ea9e30db6904b0b1f
arachnepnr 840bdfdeb38809f9f6af4d89dd7b22959b176fdd
yosys ab97eddee9b0ea0a772660731fe4c3270d2564e5
```

build w/ `make USE_NEXTPNR=true`, output is in `out/ice40.asc` `out/ice40.bin` etc

If the full design is built, the 12m clock from the PLL's GLOBALA (passthru) is visible on out1 (pin) but GLOBALB on out2 is flatline.  Exposing the lock signal showed it high in both cases.  Building the design with `hdl/ice40.v`'s `define WITH_CPU commented out (leaving only the SPI debug interface and VGA blck) results in a functional sub-design.

Commenting out `promote_globals()` in `ice40/pack.cc` per tnt's suggestion resolves the issue (but only uses 2 globals -- for the two PLL outputs).

My guess is `promote_globals()` is managing to clobber the GLOBALB PLL output...

smunaut:
Fixes #148 

@swetland If you could confirm that :)

### DIFF
--- a/common/placer1.cc
+++ b/common/placer1.cc
@@ -337,9 +337,10 @@ class SAPlacer
                         }
                     } else {
                         uint64_t score = ctx->rng64();
-                        if (score <= best_ripup_score) {
+                        CellInfo *bound_cell = ctx->getBoundBelCell(bel);
+                        if (score <= best_ripup_score && bound_cell->belStrength < STRENGTH_STRONG) {
                             best_ripup_score = score;
-                            ripup_target = ctx->getBoundBelCell(bel);
+                            ripup_target = bound_cell;
                             ripup_bel = bel;
                         }
                     }
--- a/ice40/arch.h
+++ b/ice40/arch.h
@@ -884,6 +884,13 @@ struct Arch : BaseCtx
         }
         NPNR_ASSERT_FALSE("Expected PLL pin to share an output with an SB_IO D_IN_{0,1}");
     }
+
+    int getDrivenGlobalNetwork(BelId bel) const
+    {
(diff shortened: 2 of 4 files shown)


common/placer1.cc
```diff
@@ -350,6 +351,8 @@ class SAPlacer
                     log_error("failed to place cell '%s' of type '%s'\n", cell->name.c_str(ctx), cell->type.c_str(ctx));
                 --iters;
                 ctx->unbindBel(ripup_target->bel);
+                if (ripup_target->type == id_SB_GB)
```
daveshah1: This will break the ECP5 build, where SB_GB doesn't exist (this seems just to be debugging anyway)

daveshah1:
LGTM bar one change

common/placer1.cc
smunaut: Doh .. yeah, not sure how I missed that :/

smunaut:
Just forced pushed an updated branch without that debug print.

swetland:
Confirmed.  This PR resolves #148 for me.  Thanks!

daveshah1:
Thanks again for your work @smunaut!

PR #153, opened 2018-11-27
ice40: Finer-grained control of global promotion

daveshah1:
- Disable promotion of 'logic' globals for now, at least until we have a better heuristic for this, and require `--promote-logic` to enable it
 - Add `--no-promote-globals` to disable all automatic global promotion

### DIFF
--- a/ice40/main.cc
+++ b/ice40/main.cc
@@ -65,6 +65,9 @@ po::options_description Ice40CommandHandler::getArchOptions()
     specific.add_options()("pcf", po::value<std::string>(), "PCF constraints file to ingest");
     specific.add_options()("asc", po::value<std::string>(), "asc bitstream file to write");
     specific.add_options()("read", po::value<std::string>(), "asc bitstream file to read");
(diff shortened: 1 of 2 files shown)


whitequark:
Context: on Glasgow, I have a design for UP5K with a 33.33 MHz domain where fully 1/4 of the critical path (6 ns) is spent in a global buffer.

whitequark:
Without logic global promotion, the troublesome design passes timing. Also, nextpnr now selects some CE nets for promotion instead of Logic nets, which may possibly be an improvement elsewhere.

daveshah1:
@mmicko can you check that the settings stuff looks alright?

mmicko:
Usage is fine, thing that is missing is persistence layer in ice40/project.cc  but think we need to rethink project files anyway.

whitequark:
I did a few more tests on Glasgow, and it looks like in nearly all cases, with logic global promotion disabled, I get *significantly* better timing on UP5K.

PR #158, opened 2018-11-29
Error reporting improvements

daveshah1:
- Replace assertion failure for multiple drivers with a precise error
- Replace the "no timing data" error (the first time that unknown cell types present a problem) with an "unsupported cell type" error that provides a clearer path to a solution in most cases
 - Remove the inaccurate "pre-placed design" in rulecheck
 - Add a warning for IO pins without a constraint, as this is often a bug in the design


q3k:
three reviews at the same time, man

PR #160, opened 2018-12-02
ice40: Add support for placing SB_LEDDA_IP block.

dmsc:
Signed-off-by: Daniel Serpell <daniel.serpell@gmail.com>


daveshah1:
Thanks for your improvement!

PR #163, opened 2018-12-04
Adding criticality calculation and experimental timing optimisation pass

daveshah1:
This adds `criticality` calculation (ie how close an arc is from becoming the critical path, in [0, 1]) to the timing analysis; and a highly experimental pass for optimising the critical path based on https://www.cerc.utexas.edu/utda/publications/C205.pdf


smunaut:
Just ran this on a few designs I have, average over 10 seed was consistently higher (sometimes not much, but still). But more importantly the max over 10 seed was really higher by 5 to 10%.

PR #164, opened 2018-12-04
ice40: Carry chain optimisations

daveshah1:
- @smunaut 's suggestion to include 'I3-only' connectivity during chain finding
 - Pass carry up the chain when inserting a feed-through LC to avoid the need to split the chain


smunaut:
Checked and Works for me, the carry chain placement look much better.

PR #165, opened 2018-12-04
build: Make use of the pipe option to avoid temporary files

smunaut:
This is really useful when building the ice40 with the gigantic .cc
files that generate multi gigabyte .s temporary files ... this way the
assembler just processed it in streaming way.


daveshah1:
@mmicko can you check that the Windows build is OK with this too?

mmicko:
Will do, it would affect mingw only in case of windows. But please check
memory allocation when doing db compile since when using pipe all is in
memory

smunaut:
I did watch top / memory during the compile and it didn't see to grow that much actually. I think the assembler can "injest" that file in a streaming fashion without having to have it fully in memory.

mmicko:
@daveshah1 Works fine with mingw

PR #171, opened 2018-12-07
WIP: Reworking placer1

daveshah1:
This makes several changes to placer1; bringing it closer to the algorithm described at http://www.eecg.toronto.edu/~vaughn/papers/fpga2000_arm.pdf and also working on some optimisations

In particular:
 - Timing weighting is now based on criticality rather than budget
 - "Normalised" costs are used for wirelength and timing
 - Bounding boxes are stored to optimise wirelength calculation (as suggested by @eddiehung )
 - Carry chains are now able to be moved after legalisation

The SA heuristic parameters are totally thrown off by these changes, they need fixing in order to see a QoR improvement. There may also be some performance improvements to investigate such as caching topological ordering for timing.


daveshah1:
Going to close this PR. Further placer development, combining an analytic placer with these changes to placer1 for better QoR without the runtime cost, in #219. This also solves the ECP5 QoR issues that were in this PR.

PR #182, opened 2018-12-16
json: Prefer higher level net names when a net has multiple names

daveshah1:
Previously ties were being broken arbitrarily (last name wins). For purposes such as timing constraints/reports this was not useful. Instead this applies a simple heuristic to pick a probably more useful, and at least consistent, name.


swetland:
This definitely improves quality of life for reading log output and for making it easier to reliably specify clock constraints.

Quality Patch. Would Apply Again. L@@K A++++

PR #212, opened 2019-01-18
gui: Allow shift+drag to move around the view

smunaut:
right & middle clight are really inconvenient when working with a touchpad
in tap-to-click mode

### DIFF
--- a/gui/fpgaviewwidget.cc
+++ b/gui/fpgaviewwidget.cc
@@ -644,12 +644,16 @@ void FPGAViewWidget::mousePressEvent(QMouseEvent *event)
     if (io.WantCaptureMouse)
         return;
 
-    if (event->buttons() & Qt::RightButton || event->buttons() & Qt::MidButton) {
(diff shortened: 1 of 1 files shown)


gui/fpgaviewwidget.cc
```diff
@@ -644,7 +644,10 @@ void FPGAViewWidget::mousePressEvent(QMouseEvent *event)
     if (io.WantCaptureMouse)
         return;
 
-    if (event->buttons() & Qt::RightButton || event->buttons() & Qt::MidButton) {
+    bool shift = QApplication::keyboardModifiers().testFlag(Qt::ShiftModifier);
```
q3k: How about refactoring this slightly to reduce the mess in the following `if()` statements? Something like:
```
bool shift = ...;
bool left = ...;
bool mid = ...;
bool right = ...;

if (left || mid || (left && shift) ...
if (left && !shift) ...
```

Also I think the second if should be `left && !shft`, otherwise you're executing a click action on shift-drag.

smunaut:
Just pushed a version with requested changes.

PR #216, opened 2019-01-21
[timing] Ignore output ports that start another path

eddiehung:


eddiehung:
Superseded by PR #217

Issue #116, opened 2018-11-14
Smarter initial placement for ice40


smunaut:
I have a design that fails initial placement. This is mostly because of incompatible CE lines for the flipflops and it is for a LP384 which doesn't have a lot of resources and I'm trying to pack 330 LCs in there.

icecube2 does produce a valid and working bitstream, so that design _can_ definitely fit. But looking at the output, it seems that the synthesizer of icecube 2 also produces a slightly smaller result, so this might help as well.

Anyway, what I've been toying with is to add a `int Arch::scoreBelForCell(CellInfo *cell, BelId bel) const
` method that returns a "score" of how good a BEL is for a CELL. and then I use that score in the initial placement instead of random.

Currently my attempt looks like this : 

```
int Arch::scoreBelForCell(CellInfo *cell, BelId bel) const
{       
    /* Only process LC */
    if (cell->type != id_ICESTORM_LC)
        return 0;
    
    NPNR_ASSERT(getBelType(bel) == id_ICESTORM_LC);
        
    /* If the cell doesn't have any FF, no preferences */
    if (!cell->lcInfo.dffEnable)
        return 8;

    /* If the cell has a FF, then count how many bels would be used in that slice */
    size_t num_cells = 0; 

    Loc bel_loc = getBelLocation(bel);
    for (auto bel_other : getBelsByTile(bel_loc.x, bel_loc.y)) {
        CellInfo *ci_other = getBoundBelCell(bel_other);
        if (ci_other != nullptr && bel_other != bel)
            num_cells++;
    }

    return 8 - num_cells;
}
```

Basically if the cell has flipflops, it will try to pack it in the same slice as other cells that have the same constraints. If the cell doesn't have a FF, then doesn't matter, it can get ripped up and replaced somewhere else if needed.

This does help. I went from finding placement for only ~ 230 LCs to about 320 LCs. Still not enough for my design though.

eddiehung:
@smunaut Can you share a testcase please?

smunaut:
This is the code I'm working with : https://github.com/smunaut/nextpnr/tree/ice40_initial_placement

And here's the json for a design that's exhibiting issues :

[phase_addon.json.gz](https://github.com/YosysHQ/nextpnr/files/2583413/phase_addon.json.gz)
[phase_addon.pcf.gz](https://github.com/YosysHQ/nextpnr/files/2583417/phase_addon.pcf.gz)

```
[...]

PR #219, opened 2019-01-26
HeAP-based analytical placer

daveshah1:
Note that is is based on top of #171 (commits up to and including ba6a991) because it uses that parts of that placer for iterative refinement. A decision on merge order remains to be made.

This is a simple analytical placer based on [HeAP](https://janders.eecg.utoronto.ca/pdfs/marcelfpl12.pdf) and [SimPL](http://www.ece.umich.edu/cse/awards/pdfs/iccad10-simpl.pdf). It uses an iterative approach to minimise HPWL, mixed with recursive cut-based spreading followed by an additional strict legalisation pass to  deal with bel validity requirements. In lieu of the iterative refinement described in the paper, after AP finishes regular SA is entered at constant low temperature and diameter (which implements a near-identical algorithm in practice).

On a mid-size iCE40 design (hx8kdemo):
 - master: 85s place time, 7s route time, 33.5MHz Fmax
 - HeAP: 12s place time, 7s route time, 33.4MHz Fmax

On a large ECP5 design (7x picorv32s - 40k LUT4, 26k SLICE) on 85k ECP5:
 - master: 5968s place time, wirelength 199912, 600s route time, 73.99MHz Fmax
 - HeAP: 154s place time, wirlelength 225260, 838s route time, 79.35MHz Fmax


3rdparty/taucs/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+endif()
+
+add_executable(taucs_configurator configurator/taucs_config.c)
+set(ENV_CMD ${CMAKE_COMMAND} -E env "OSTYPE=${TAUCS_OSTYPE}")
+set(TAUCS_CONFIG_DIR ${CMAKE_CURRENT_BINARY_DIR}/build/linux)
+set(TAUCS_CONFIG_BUILD ${TAUCS_CONFIG_DIR}/taucs_config_build.h)
+set(TAUCS_CONFIG_TESTS ${TAUCS_CONFIG_DIR}/taucs_config_tests.h)
+
+
+file(WRITE ${TAUCS_CONFIG_TESTS} "#define TAUCS_CORE\n#define TAUCS_BLAS_NOUNDERSCORE\n#include <cblas.h>\n")
```
gsomlo: at least on Fedora, cblas.h is found under /usr/include/cblas/cblas.h, so this should be either "include <cblas/cblas.h>" or "-I /usr/include/cblas" added to CFLAGS somewhere else. Maybe the latter, since Fedora might be mangling the install path of various include files w.r.t. the original cmakefile shipping with lapack/cblas...

daveshah1: Thanks for pointing this out. It seems like Ubuntu and Arch both install it straight into /usr/include; so the latter seems like a better option.

Due to various problems with TAUCS, I am currently investigating alternative solver options, which might remove the BLAS dependency altogether in any case.

gsomlo:
The latest/current version of this PR (applied over commit 4a2560b) got my rv64imadc-based blinky demo finished in 17:26 minutes (down from 4:49:42 hours with commit db990d2). I haven't actually tested it by pushing to the FPGA, but will do so early on Monday and report back.

3rdparty/taucs/CMakeLists.txt
daveshah1: TAUCS has been replaced by Eigen3 now.

gsomlo:
eigen3 based version tested on fedora28 and https://github.com/gsomlo/yoloRISC; nextpnr run takes cca 15 minutes, resulting bitstream still working great.

daveshah1:
This is now (almost) ready for merge from my PoV. The last remaining issue is the failing regression test for #116 - even though HeAP isn't used by default, the general changes to placer1 cause it to now fail. However, it was also failing on master with different seeds so anything that peturbed the PRNG could also cause it to fail. So I think we should neutralise this test until we genuinely fix initial placement  and post-legalisation re-placement.

common/nextpnr.cc
```diff
@@ -221,6 +221,9 @@ delay_t Context::getNetinfoRouteDelay(const NetInfo *net_info, const PortRef &us
         return 0;
 #endif
 
+    if (net_info->wires.empty())
```
eddiehung: What's the impact of this change? Just a speedup?

daveshah1: Yes, makes timing analysis significantly faster for unrouted designs (it would always call the bottom predictDelay in this case, just avoids the extra getNetinfoSourceWire etc calls)

ecp5/main.cc
```diff
@@ -59,6 +59,8 @@ po::options_description ECP5CommandHandler::getArchOptions()
     specific.add_options()("um5g-45k", "set device type to LFE5UM5G-45F");
     specific.add_options()("um5g-85k", "set device type to LFE5UM5G-85F");
 
+    specific.add_options()("sa-placer", "use pure simulated annealing placer instead of HeAP analytic placer");
```
eddiehung: So ecp5 has a new `"sa-placer"` option, whereas ice40 has a new `"heap-placer"` option. Intentional?

daveshah1: SA remains the default for iCE40, which is a more stable architecture and small enough for speed to be less of a problem for most people. SA tends to win on iCE40 QoR too.

HeAP becomes the default for ECP5, which has major runtime (into several hours) issues with SA, but has mostly experimental users. HeAP tens to win for ECP5 QoR because it is better at moving chains which the ECP5 has lots of due to distributed RAM and cascade muxes. A SA fallback is still available and will be used automatically if HeAP is disabled in CMake (e.g. because Eigen3 isn't easily available for the target).

eddiehung: Makes sense. Should write that down somewhere. I would encourage you to fixate on one argument string, and then set it to default to one value for ice40, and another value for ecp5 though...

common/placer_heap.cc
```diff
@@ -0,0 +1,1543 @@
...
+                } else {
+                    std::thread xaxis([&]() { build_solve_direction(false, (iter == 0) ? -1 : iter); });
+                    build_solve_direction(true, (iter == 0) ? -1 : iter);
+                    xaxis.join();
+                }
+                auto solve_endt = std::chrono::high_resolution_clock::now();
+                solve_time += std::chrono::duration<double>(solve_endt - solve_startt).count();
+                update_all_chains();
+                solved_hpwl = total_hpwl();
+                log_info("Solved HPWL = %d\n", int(solved_hpwl));
```
eddiehung: In line with previous SA placer, can we be less spammy by only coughing these up under `--verbose`?

eddiehung:
On a mid-size iCE40 design (hx8kdemo) with `--freq 40`:

81a7b5d without `--heap-placer`: 79s place time, 114s route time, 37.47MHz Fmax
81a7b5d with `--heap-placer`: 16s place time, 12s route time, 35.59MHz Fmax

Also with `--seed 0`
81a7b5d without `--heap-placer`: 81s place time, 140s route time, 32.77MHz Fmax
81a7b5d with `--heap-placer`: 14s place time, 9s route time, 38.68MHz Fmax

And with `--seed 1`
81a7b5d without `--heap-placer`: 84s place time, 82s route time, 35.95MHz Fmax
81a7b5d with `--heap-placer`: 17s place time, 10s route time, 34.48MHz Fmax

After this very unrigorous exercise, I think it's safe to say there doesn't seem to be a huge change in final nextpnr-reported Fmax, but it is significant that not just placer time is down, routing time is down too...

eddiehung:
Okay, looks like we're back to normal now:

52e05f4 without --heap-placer: 118s place time, 10s route time, 31.63MHz Fmax
Also with --seed 0:
52e05f4 without --heap-placer: 115s place time, 12s route time, 31.62MHz Fmax
And with --seed 1:
52e05f4 without --heap-placer: 118s place time, 10s route time, 32.99MHz Fmax

daveshah1:
Added the `--placer` option and moved stats to one line. Looking at nextpnr-bench now.

daveshah1:
nextpnr-bench results

arachne-pnr baseline:
```
design01-hx8k average Fmax: 25.98MHz runtime: 69.9s place: 30.6s route: 39.2s
design01-hx8k stddev Fmax: 2.01MHz runtime: 25.4s place: 1.0s route: 24.9s

design02-hx1k average Fmax: 56.13MHz runtime: 4.2s place: 3.1s route: 1.1s
design02-hx1k stddev Fmax: 4.42MHz runtime: 0.4s place: 0.3s route: 0.3s

design02-hx8k average Fmax: 56.32MHz runtime: 9.2s place: 3.7s route: 5.5s
design02-hx8k stddev Fmax: 2.48MHz runtime: 0.4s place: 0.5s route: 0.5s

design03-up5k average Fmax: 6.39MHz runtime: 43.2s place: 21.2s route: 22.0s
design03-up5k stddev Fmax: 0.20MHz runtime: 2.7s place: 1.2s route: 2.3s
```

master:
```
design01-hx8k average Fmax: 35.80MHz runtime: 97.5s place: 89.3s route: 8.2s
design01-hx8k stddev Fmax: 0.82MHz runtime: 2.7s place: 2.3s route: 1.2s

design02-hx1k average Fmax: 71.50MHz runtime: 8.6s place: 7.8s route: 0.8s
design02-hx1k stddev Fmax: 2.84MHz runtime: 0.8s place: 0.7s route: 0.4s

design02-hx8k average Fmax: 72.24MHz runtime: 12.6s place: 11.5s route: 1.1s
design02-hx8k stddev Fmax: 2.28MHz runtime: 0.9s place: 0.9s route: 0.3s

design03-up5k average Fmax: 8.25MHz runtime: 65.7s place: 50.2s route: 15.5s
design03-up5k stddev Fmax: 0.22MHz runtime: 1.9s place: 0.6s route: 1.8s
```

new SA:
```
design01-hx8k average Fmax: 40.48MHz runtime: 142.9s place: 125.0s route: 17.9s
design01-hx8k stddev Fmax: 1.23MHz runtime: 7.1s place: 5.3s route: 7.1s

design02-hx1k average Fmax: 76.48MHz runtime: 7.3s place: 6.6s route: 0.7s
design02-hx1k stddev Fmax: 1.64MHz runtime: 0.5s place: 0.5s route: 0.5s

design02-hx8k average Fmax: 76.69MHz runtime: 8.0s place: 7.0s route: 1.0s
design02-hx8k stddev Fmax: 4.16MHz runtime: 0.4s place: 0.4s route: 0.0s

design03-up5k average Fmax: 8.66MHz runtime: 91.5s place: 66.0s route: 25.5s
design03-up5k stddev Fmax: 0.22MHz runtime: 4.1s place: 1.5s route: 3.7s
```

HeAP:
```
design01-hx8k average Fmax: 40.08MHz runtime: 24.6s place: 15.7s route: 8.9s
design01-hx8k stddev Fmax: 0.71MHz runtime: 2.4s place: 1.3s route: 1.4s

design02-hx1k average Fmax: 72.71MHz runtime: 5.4s place: 4.5s route: 0.9s
design02-hx1k stddev Fmax: 2.14MHz runtime: 1.0s place: 1.2s route: 0.3s

design02-hx8k average Fmax: 75.37MHz runtime: 3.3s place: 2.2s route: 1.1s
design02-hx8k stddev Fmax: 2.07MHz runtime: 0.8s place: 0.9s route: 0.3s

design03-up5k average Fmax: 8.56MHz runtime: 33.5s place: 18.9s route: 14.6s
design03-up5k stddev Fmax: 0.17MHz runtime: 4.0s place: 3.0s route: 2.3s
```

tl;dr the criticality changes, etc, do make the SA option slower for some designs - but with an increase in Fmax on the other hand. I don't know if the existence of the now-much-faster HeAP option for those who care about runtime is enough to justify this. Hopefully in a month or two we can make HeAP the default anyway.

daveshah1:
nextpnr-bench with new SA and `--placer-budgets`

```
design01-hx8k average Fmax: 36.29MHz runtime: 99.8s place: 92.9s route: 6.9s
design01-hx8k stddev Fmax: 0.78MHz runtime: 4.0s place: 4.3s route: 1.4s

design02-hx1k average Fmax: 74.16MHz runtime: 6.0s place: 5.6s route: 0.4s
design02-hx1k stddev Fmax: 1.75MHz runtime: 0.4s place: 0.5s route: 0.5s

design02-hx8k average Fmax: 72.12MHz runtime: 6.6s place: 6.0s route: 0.6s
design02-hx8k stddev Fmax: 1.83MHz runtime: 0.5s place: 0.4s route: 0.5s

design03-up5k average Fmax: 7.55MHz runtime: 72.5s place: 52.0s route: 20.5s
design03-up5k stddev Fmax: 0.12MHz runtime: 6.0s place: 2.1s route: 5.8s
```

janrinze:
Is there a specific reason why the speedup on up5k is so much smaller? I expected a bigger difference between master and HeAP.
With SA on hx8k compared to master we get a 36% higher Fmax compared to master.
With SA on up5k the resulting Fmax is only 5% more compared to master.

A possible explanation that i was thinking of is that the delay contributed by the placement is much shorter than the delay of the LUTs and other component. Which in turn would mean that only a small overall reduction in signal propagation can be achieved through better placement on up5k.
The difference with arachne-pnr does however seem to paint a different picture.

The reduced processing time of HeAP is a very welcome result.

daveshah1:
@janrinze FYI, the design03-up5k benchmark is a totally different design (a NES core) to the design01-hx8k one (picosoc). Indeed you are correct, the design03-up5k test has a high number of logic levels (and is pushing the device to the limits in terms of utilisation), so there is little a placer can do.

janrinze:
one a side note: does any of the timing rules take into account if there are mixed uses of posedge and negedge of the same clock?

daveshah1:
Yes, that is considered inside nextpnr's timing analysis (unlike icetime) and Fmax will be calculated correctly where the critical path for a clock is between different edges.

common/placer_heap.h
```diff
@@ -0,0 +1,47 @@
...
+    float alpha;
+    float criticalityExponent;
+    float timingWeight;
+
+    std::unordered_set<IdString> ioBufTypes;
+};
+
+extern bool placer_heap(Context *ctx, PlacerHeapCfg cfg);
+NEXTPNR_NAMESPACE_END
+#endif
```
cliffordwolf: Pedantic nitpick: missing newline at end of file

daveshah1:
HeAP log output now:
```
Info: Placed 10 cells based on constraints.
Info: Creating initial analytic placement for 1034 cells, random placement wirelen = 129705.
Info:     at initial placer iter 0, wirelen = 791
Info:     at initial placer iter 1, wirelen = 1140
Info:     at initial placer iter 2, wirelen = 729
Info:     at initial placer iter 3, wirelen = 1134
Info:     at iteration #1, type TRELLIS_SLICE: wirelen solved = 733, spread = 22039, legal = 22452; time = 0.05s
Info:     at iteration #2, type TRELLIS_SLICE: wirelen solved = 2315, spread = 16049, legal = 16239; time = 0.05s
Info:     at iteration #3, type TRELLIS_SLICE: wirelen solved = 2403, spread = 14645, legal = 15010; time = 0.05s
Info:     at iteration #4, type TRELLIS_SLICE: wirelen solved = 2744, spread = 14616, legal = 14945; time = 0.05s
Info:     at iteration #5, type TRELLIS_SLICE: wirelen solved = 3585, spread = 14719, legal = 15122; time = 0.05s
Info:     at iteration #6, type TRELLIS_SLICE: wirelen solved = 4664, spread = 14166, legal = 14604; time = 0.05s
Info:     at iteration #7, type TRELLIS_SLICE: wirelen solved = 5167, spread = 14901, legal = 15407; time = 0.04s
Info:     at iteration #8, type TRELLIS_SLICE: wirelen solved = 6390, spread = 14143, legal = 14644; time = 0.04s
Info:     at iteration #9, type TRELLIS_SLICE: wirelen solved = 7436, spread = 14545, legal = 15043; time = 0.05s
Info:     at iteration #10, type TRELLIS_SLICE: wirelen solved = 8051, spread = 14224, legal = 14780; time = 0.05s
Info:     at iteration #11, type TRELLIS_SLICE: wirelen solved = 8627, spread = 15196, legal = 15713; time = 0.05s
Info: HeAP Placer Time: 1.03s
Info:   of which solving equations: 0.63s
Info:   of which spreading cells: 0.03s
Info:   of which strict legalisation: 0.02s

Info: Running simulated annealing placer for refinement.
Info:   at iteration #1: temp = 0.000000, timing cost = 661, wirelen = 14604
Info:   at iteration #5: temp = 0.000000, timing cost = 462, wirelen = 12759
Info:   at iteration #10: temp = 0.000000, timing cost = 389, wirelen = 12359
Info:   at iteration #15: temp = 0.000000, timing cost = 393, wirelen = 12104
Info:   at iteration #20: temp = 0.000000, timing cost = 388, wirelen = 12035
Info:   at iteration #20: temp = 0.000000, timing cost = 388, wirelen = 12037 
Info: SA placement time 1.68s
```

Initial placement for SA tries to find a random placement that is legal in all senses other than relative constraints (the current implementation could probably be sped up, I think it is a bit obsessive about how random the placement is).

For HeAP it is a bit more involved:
 - For IO buffer cells (a set of IO buffer cell types is given to HeAP), a legal placement is found if not constrained already, then that position is locked - this prevents the matrix from being singular if pins aren't constrained
 - For all other cells that are not relatively constrained (but may be the parent of relatively constrained cells i.e. root of a chain), we come up with a fast, vaguely random, not-necessarily-legal placement. Cells that are relatively constrained are "implicitly considered" throughout HeAP, only the root is considered in the equations with children being a fixed offset from the root.
 - The iterative part of initial placement then runs the solver, without any spreading or legalisation, a fixed number of times to converge on an illegal solution that is the input to the main solve/spread/legalise loop.

eddiehung:
Top notch stuff. APPROVED!

P.S. What's the significance of `type TRELLIS_SLICE:`? On ice40 I see `type ALL:`?

daveshah1:
There are three possible options wrt the types, this is to do with the heterogeneous part of the HeAP algorithm.

 - If there is only one cell type in the design (excl. IO buffers and other constrained cell types), only that cell type ever needs to be considered. That is the `TRELLIS_SLICE` case I showed.
 - If there are multiple types of cell to be considered, but one cell type significantly dominates, all cell types are always considered each iteration (threshold currently 98%, could be experimented with further), this is the `ALL` case you have. This is an optimisation I have added on top of original HeAP.
 - If there are multiple types of cell and the design is truly heterogeneous (not triggering the optimisation above) - particularly designs with significant amount of DSP and RAM usage - then it will round-robin between solving for one cell type in particular with other types locked, and solving all cells at once. Then you will see different types.

PR #220, opened 2019-01-27
ice40: Add budget override for CO->I3 path

daveshah1:
Signed-off-by: David Shah <dave@ds0.me>

### DIFF
--- a/ice40/arch.cc
+++ b/ice40/arch.cc
@@ -621,6 +621,29 @@ bool Arch::getBudgetOverride(const NetInfo *net_info, const PortRef &sink, delay
             }
         }
         return true;
+    } else if (driver.port == id_COUT && sink.port == id_I3) {
+        bool same_y = driver.cell->constr_abs_z && driver.cell->constr_z < 7;
+        switch (args.type) {
+#ifndef ICE40_HX1K_ONLY
+        case ArchArgs::HX8K:
+#endif
+        case ArchArgs::HX1K:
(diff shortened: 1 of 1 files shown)


eddiehung:
Suggest refactor entire function:
```
bool Arch::getBudgetOverride(const NetInfo *net_info, const PortRef &sink, delay_t &budget) const
{
    const auto &driver = net_info->driver;
    if (driver.port == id_COUT) {
        NPNR_ASSERT(sink.port == id_CIN || sink.port == id_I3);
        NPNR_ASSERT(driver.cell->constr_abs_z);
        bool cin = sink.port == id_CIN;
        bool same_y = driver.cell->constr_z < 7;
        if (cin && same_y)
            budget = 0;
        else {
            switch (args.type) {
#ifndef ICE40_HX1K_ONLY
            case ArchArgs::HX8K:
#endif
            case ArchArgs::HX1K:
                budget = cin ? 190 : (same_y ? 260 : 560);
                break;
#ifndef ICE40_HX1K_ONLY
            case ArchArgs::LP384:
            case ArchArgs::LP1K:
            case ArchArgs::LP8K:
                budget = cin ? 290 : (same_y ? 380 : 670);
                break;
            case ArchArgs::UP5K:
                budget = cin ? 560 : (same_y ? 660 : 1220);
                break;
#endif
            default:
                log_error("Unsupported iCE40 chip type.\n");
            }
        }
        return true;
    }
    return false;
}
```

PR #221, opened 2019-01-27
[WIP] Allow timing graph to be visualised by outputting as DOT file

eddiehung:
If `--debug`, (over)write out a "timing.dot" on each invocation of timing analysis.
Advisable to only use dot for small circuits.
[blinky example](https://github.com/YosysHQ/nextpnr/files/2800453/timing.pdf)

Needed a new Arch API: `Arch::getBelIOB(BelId)`

### DIFF
--- a/common/timing.cc
+++ b/common/timing.cc
@@ -639,18 +639,20 @@ struct Timing
         }
 
         if (ctx->debug) {
+            log_info("Writing timing.dot\n");
             std::ofstream f("timing.dot");
             f << "digraph T {" << std::endl;
             f << "\tlabel=\"clk_period=" << clk_period << "\";" << std::endl;
             f << "\tlabelloc=t;" << std::endl;
+            f << "\tnewrank=true;" << std::endl;
             for (auto &net : ctx->nets) {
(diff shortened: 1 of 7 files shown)


common/timing.cc
```diff
@@ -636,6 +637,147 @@ struct Timing
             }
 #endif
         }
+
+        if (ctx->debug) {
```
q3k: Would it make sense to have a different runtime/commandline flag for this? I would imagine someone would want to run `--debug` on large designs without generating the timing graphviz, especially when you say it's CPU intensive.

eddiehung: Sorry for being unclear: writing out timing.dot is not particularly CPU intensive, but running dot to render it is.

daveshah1: Personally I don't like the idea of a PnR tool deciding to create files in the current directory even in debug mode, it's the kind of thing that annoys me when using vendor tools. I think this as  separate option would be better.

gatecat:
with all the conflicts and changes inbetween, I think it's time to close this old friend

PR #223, opened 2019-02-03
ecp5: cmake: Search for pytrellis.so in multiple locations

gsomlo:
If a distro-specific "prjtrellis-devel" packaged dependency is used,
the pytrellis.so library might be located in a dedicated directory,
rather than under ${TRELLIS_ROOT}.

Search for pytrellis.so in a list of directories, then subsequently
use the first match as part of ${PYTHONPATH}.

Signed-off-by: Gabriel Somlo <gsomlo@gmail.com>


gsomlo:
see also: https://github.com/SymbiFlow/prjtrellis/pull/59

PR #226, opened 2019-02-07
Fix slack_histogram computation accessing num_bins+1

eddiehung:
Also added failing (ice40) testcase into new `regressions` subdir inside `ice40`


ice40/regressions/Makefile
```diff
@@ -0,0 +1,8 @@
+NPNR = ../../nextpnr-ice40
+JSON = $(wildcard */*.json)
+OUTPUT = $(patsubst %.json,%.log,$(JSON))
+
+all: $(NPNR) $(OUTPUT)
+
+%.log: %.json $(wildcard %.pcf)
```
q3k: This doesn't do what it's supposed to - on my machine (Gentoo, GNU Make 4.2.1), it ends up not calling nextpnr with `--pcf` for the submitted test case.

ice40/regressions/Makefile
```diff
@@ -0,0 +1,8 @@
+NPNR = ../../nextpnr-ice40
+JSON = $(wildcard */*.json)
+OUTPUT = $(patsubst %.json,%.log,$(JSON))
+
+all: $(NPNR) $(OUTPUT)
```
q3k: I don't think NPNR should be a prereq here - it's not like this Makefile knows how to build it, and this prevents you from doing `make NPNR=nextpnr-ice40`, you have to do `make NPNR=$(which nextpnr-ice40)` instead.

q3k:
Also, please add running those tests to the CI scripts (`.cirrus.yml`).

ice40/regressions/Makefile
eddiehung: Thanks! I spotted some flaws when I was building more testcases.

Perhaps this is better for you?
```
%.log: %.json $(wildcard %.pcf) $(wildcard %.npnr) $(NPNR)
        $(NPNR) --json $(word 1,$^) $(if $(wildcard $*.pcf),--pcf $*.pcf,)) > $@ 2>&1
```
I'm not entirely sure why it would fail on your make...

ice40/regressions/Makefile
eddiehung: You're right: it doesn't know how to make it.
I do however want it to rebuild if the NPNR in my sandbox changed. Whilst I think of how to achieve both, can you think of a solution?

eddiehung:
I'm going to remove the test from this PR/repo so that I can put it in the new nextpnr-tests repo.

eddiehung:
> Also, please add running those tests to the CI scripts (`.cirrus.yml`).

I plan on doing so in a future PR once I've merged in #227

Issue #209, opened 2019-01-12
LVDS clock input results in assertion failure on iCE40


bgamari:
Minimal reproducer:

```verilog
// test.v
module test (input  PIN_5, output wire PIN_18);
   lvds_clock_input in (PIN_5, PIN_18);
endmodule

module lvds_clock_input ( input pin, output wire rd );
   SB_GB_IO #(.PIN_TYPE(6'b0000_00),
           .IO_STANDARD("SB_LVDS_INPUT")) lvds_in
     (.PACKAGE_PIN(pin), .GLOBAL_BUFFER_OUTPUT(rd));
endmodule
```

```
# test.pcf
set_io --warn-no-port PIN_5 C1
set_io --warn-no-port PIN_18 A9
```

```
$ yosys -p "synth_ice40 -top test; json -o test.json" test.v
$ nextpnr-ice40 --lp8k --package cm81 --pcf test.pcf --json test.json
...
Info: Packing constants..
Info: Packing IOs..
Info: PIN_5 feeds SB_IO in.lvds_in, removing $nextpnr_ibuf PIN_5.
terminate called after throwing an instance of 'nextpnr_ice40::assertion_failure'
  what():  Assertion failure: wire != WireId() (/build/source/ice40/arch.h:610)
```

smunaut:
https://github.com/YosysHQ/nextpnr/pull/213

This will show a slightly more understandable error message. (especially since the site id is printed and linked to the pin a couple lines earlier, so it should be clear enough).

For your case, if you want to still use this as a clock input, the only solution is to use a SB_IO followed by a separate SB_GB.  You loose the shorter path / deterministic timing of the SB_GB_IO but if you don't use that clock to capture any IO linked to it, that doesn't matter at all.

bgamari:
Ahh, silly me. Of course.

Thanks @smunaut!

Issue #203, opened 2019-01-04
Ring oscillator fails with NextPNR


Mecrisp:
I wish to switch from Arachne-PNR to NextPNR, but my design contains a ring oscillator for random number generation.

Arachne-PNR is fine with it, but NextPNR fails with "ERROR: timing analysis failed due to presence of combinatorial loops, incomplete specification of timing ports, etc." even when launching with the --no-tmdriv option.

How can I place and route ring oscillators with NextPNR ? I include a minimal example.

[ringosc.v.txt](https://github.com/YosysHQ/nextpnr/files/2727389/ringosc.v.txt)
[ringosc.pcf.txt](https://github.com/YosysHQ/nextpnr/files/2727392/ringosc.pcf.txt)
[compile.txt](https://github.com/YosysHQ/nextpnr/files/2727393/compile.txt)

I would prefer if I could use timing driven routing for most of the design and just denote the ring oscillator as "do not care".

daveshah1:
You can use `--force` to override this warning (

Mecrisp:
Thank you, up and running now.

Issue #188, opened 2018-12-20
-pullup yes causes pin defined in pcf to be unconstrained and not connected


janrinze:
having a pin defined in a pcf file with the " -pullup yes " flag fails to allocate the pin with nextpnr.

see attached file,

untar attached file in an empty directory
 yosys -q -p 'synth_ice40 -top top -json test.json' test.v
nextpnr-ice40 --pcf hx8k.pcf --json test.json --asc test.asc --hx8k

output:
Info: Importing module top
Info: Rule checker, verifying imported design
Info: Checksum: 0xade084ac

Info: constrained 'led1' to bel 'X17/Y33/io0'
Info: constrained 'led2' to bel 'X16/Y33/io1'
Info: constrained 'led3' to bel 'X17/Y0/io0'
Info: constrained 'pclk' to bel 'X16/Y0/io1'
Warning: unmatched constraint 'yes' (on line 10)

Info: Packing constants..
Info: Packing IOs..
Warning: IO 'button' is not constrained to a pin and will be automatically placed
Info: Packing LUT-FFs..
.......

The warning clearly shows that button is not properly assigned to the pin.
[nextpnrtest.tar.gz](https://github.com/YosysHQ/nextpnr/files/2700230/nextpnrtest.tar.gz)

daveshah1:
See #189

Issue #186, opened 2018-12-20
using pll lock causes assert error in nextpnr


janrinze:
When using a lock signal from the pll module (created with icepll) triggers an assert error.

to replicate:
untar attached file in an empty directory,

yosys -q -p 'synth_ice40 -top top -json test.json' test.v
nextpnr-ice40 --pcf hx8k.pcf --json test.json --asc test.asc --hx8k

error given is:

Info: Importing module top
Info: Rule checker, verifying imported design
Info: Checksum: 0x00b23127

Info: constrained 'led1' to bel 'X17/Y33/io0'
Info: constrained 'led2' to bel 'X16/Y33/io1'
Info: constrained 'led3' to bel 'X17/Y0/io0'
Info: constrained 'pclk' to bel 'X16/Y0/io1'

Info: Packing constants..
Info: Packing IOs..
Info: Packing LUT-FFs..
Info: Packing non-LUT FFs..
Info: Packing carries..
Info: Packing RAMs..
Info: Placing PLLs..
Info:   constrained PLL 'mypll.uut' to X16/Y33/pll_3
Info:   (given its connections, this PLL could have been a PAD PLL)
Info: Packing special functions..
Info:   PLL 'mypll.uut' has LOCK output, need to pass all outputs via LUT
Info:   constrained '$abc$231$auto$blifparse.cc:492:parse_blif$232_LC' to X1/Y32/lc0
Info:   constrained '$auto$alumacc.cc:474:replace_alu$8.slice[15].adder_LC' to X1/Y32/lc1
Info:   constrained '$auto$alumacc.cc:474:replace_alu$8.slice[16].adder_LC' to X1/Y32/lc2
Info:   constrained '$auto$alumacc.cc:474:replace_alu$8.slice[17].adder_LC' to X1/Y32/lc3
Info:   constrained '$auto$alumacc.cc:474:replace_alu$8.slice[18].adder_LC' to X1/Y32/lc4
Info:   constrained '$auto$alumacc.cc:474:replace_alu$8.slice[19].adder_LC' to X1/Y32/lc5
Info:   constrained '$auto$alumacc.cc:474:replace_alu$8.slice[20].adder_LC' to X1/Y32/lc6
Info:   constrained '$auto$alumacc.cc:474:replace_alu$8.slice[0].adder_LC' to X1/Y32/lc7
terminate called after throwing an instance of 'nextpnr_ice40::assertion_failure'
[...]

PR #227, opened 2019-02-07
[tests] Move tests subdir to submodule, also add regression tests of fixed ice40 issues

eddiehung:
New submodule: https://github.com/YosysHQ/nextpnr-tests

In which [ice40/regressions](https://github.com/YosysHQ/nextpnr-tests/tree/master/ice40/regressions) contains testcases extracted from #69 #70 #71 #72 #84 #98 #106 #117 #120 #121 #126 #151 #170 #176 #186 #188 #203 #209 
EDIT: and more that are referenced from the commit message.

Right now, these regression tests are really just a sanity check -- nextpnr the JSON, icebox_vlog the resulting ASC, and make sure neither of them returns with a bad exit code.
There are a few exceptions where an error is expected in nextpnr, and this is when we grep for a particular error message having appeared and then fake the icebox_vlog output.

Further down the line, I think we would want to use Yosys' `test_autotb` feature to run end-to-end Verilog->Yosys->nextpnr->Verilog tests with auto generated testbenches... (for those testcases we have source Verilog for)

Currently, just the ice40 testcases have been added. These tests are run by doing `make -C tests/ice40/regressions` and are ~not~ now run by the CI ~yet~.


daveshah1:
Probably sensible to add something explaining how to ensure the submodule is fetched to the README. Otherwise LGTM.

eddiehung:
> Probably sensible to add something explaining how to ensure the submodule is fetched to the README. Otherwise LGTM.

https://github.com/YosysHQ/nextpnr-tests/blob/master/README.md

PR #230, opened 2019-02-09
ice40: Don't create PLLOUT_B buffer for single-output PLL variants

### LINKED ISSUE #229: ERROR: Cell '$gbuf_SB_GB_IO_io' cannot be bound to bel 'X0/Y17/gb' since it is already bound to cell '$gbuf_SB_PLL40_CORE_pllout_b'
To reproduce, use this archive: [vga-output.zip](https://github.com/YosysHQ/nextpnr/files/2847603/vga-output.zip)

I don't entirely understand the nature of the bug. I think it's related to global promotion. The error message refers to (a) a bel that has never been mentioned in the log before, and (b) an internal name of a global buffer that has never been mentioned in the log before, so at least the error message is wrong. 

But, I also don't know how to make the design work; switching to PLLOUTGLOBAL instead of PLLOUTCORE has no effect.

daveshah1:
Fixes #229 

For single-output PLL variants, we were only checking for conflicts with the first output; but still creating SB_GBs for both outputs.

### DIFF
--- a/ice40/pack.cc
+++ b/ice40/pack.cc
@@ -1063,7 +1063,12 @@ static void pack_special(Context *ctx)
                     create_ice_cell(ctx, ctx->id("ICESTORM_PLL"), ci->name.str(ctx) + "_PLL");
             packed->attrs[ctx->id("TYPE")] = ci->type.str(ctx);
             packed_cells.insert(ci->name);
-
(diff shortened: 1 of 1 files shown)


whitequark:
I can indeed confirm that it works great on real hardware:

PR #241, opened 2019-02-22
ice40: support u4k

corecode:


ice40/main.cc
```diff
@@ -60,6 +60,7 @@ po::options_description Ice40CommandHandler::getArchOptions()
     specific.add_options()("hx1k", "set device type to iCE40HX1K");
     specific.add_options()("hx8k", "set device type to iCE40HX8K");
     specific.add_options()("up5k", "set device type to iCE40UP5K");
+    specific.add_options()("u4k", "set device type to iCE40LP4K");
```
daveshah1: iCE5LP4K (iCE40LP4K is a different device, rebranded 8k)

ice40/chipdb.py
```diff
@@ -76,6 +76,7 @@
 constids["MAC16"] = constids["ICESTORM_DSP"]
 constids["HFOSC"] = constids["ICESTORM_HFOSC"]
 constids["LFOSC"] = constids["ICESTORM_LFOSC"]
+constids["SMCCLK"] = constids["ICESTORM_SMCCLK"]
```
daveshah1: The ICESTORM_ prefix is added when we modify a primitive for any reason (such as adding the _FABRIC ports and trimming to oscillators, or converting a bus to a list of signals). I don't think it's needed for SMCCLK, which might as well stay as just SMCCLK.

daveshah1:
LGTM with a few tiny changes.

This needs cliffordwolf/icestorm#202 to be merged first (I also expect CI to fail until this is merged and the icestorm commit in the Dockerfile bumped)

PR #245, opened 2019-03-03
Use cmake's find_library to search for pytrellis.

twam:
This is the matching pull request for https://github.com/SymbiFlow/prjtrellis/pull/72 to fix building on OS X.

PR #246, opened 2019-03-03
ecp5/cmake: Add TRELLIS_INSTALL_ROOT variable

daveshah1:
Signed-off-by: David Shah <dave@ds0.me>


twam:
We should align this with https://github.com/YosysHQ/nextpnr/pull/245

PR #260, opened 2019-04-02
README: update windows build instructions for eigen, python

xobs:
The recent addition of "eigen3" to nextpnr necessitates the addition of
"eigen" to the list of vcpkg packages to be installed on Windows.

Add the 32- and 64-bit versions of eigen to the command issued to
install various dependencies.

Also document one possible build command, and note about unpacking
python36.zip.

Signed-off-by: Sean Cross <sean@xobs.io>


daveshah1:
Thanks!

PR #261, opened 2019-04-02
Python API for generic architecture

daveshah1:
This adds the python bindings to the generic arch API to allow programmatic construction of arbitrary architectures. It also adds documentation for these, a simple generic packer for LUTs, DFFs and IO buffer insertion, and a small example of this. 

A more real-world example using this API to place-and-route for a small FPGA remains to be developed.


daveshah1:
Documentation: https://github.com/YosysHQ/nextpnr/blob/pygeneric/docs/generic.md
Example: https://github.com/YosysHQ/nextpnr/tree/pygeneric/generic/examples
Generic primitives: https://github.com/YosysHQ/nextpnr/blob/pygeneric/generic/synth/prims.v

mithro:
@daveshah1 Great work, looking forward to trying this out sometime!

generic/examples/simple.py
```diff
@@ -0,0 +1,87 @@
+# Grid size including IOBs at edges
+X = 12
+Y = 12
+# SLICEs per tile
+Z = 8
```
eddiehung: `N` is traditionally the number of LUTs in a logic block/tile

generic/examples/simple.py
```diff
@@ -0,0 +1,87 @@
...
+# LUT input count
+K = 4
+# Number of local wires
+L = Z*(K+1) + 8
```
eddiehung: You might want to consider using the term `W` to represent some form of channel width -- i.e. the number of routing tracks in a routing channel; `L` is typically used to describe the length of a routing wire

PR #263, opened 2019-04-07
placer1: Improve incremental bounding box updates

daveshah1:
This further improves incremental bounding box updates by keeping track of the number of cells on each edge of the bounding box, so a full recompute is needed only when the last cell at an edge moves inwards (moves outwards never need a full recompute).

Although there is no significant runtime improvement on small designs, on a large design it gives about a 3x runtime improvement.

### DIFF
--- a/common/placer1.cc
+++ b/common/placer1.cc
@@ -64,9 +64,10 @@ class SAPlacer
   private:
     struct BoundingBox
     {
+        // Actual bounding box
         int x0 = 0, x1 = 0, y0 = 0, y1 = 0;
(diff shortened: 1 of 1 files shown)


eddiehung:
> Although there is no significant runtime improvement on small designs, on a large design it gives about a 3x runtime improvement.

Yowsers. Sorry for doubting you! I guess the impact is big when the bounding box shrinks to low single digits in which case the chances are many things are touching the perimeter.

The other concern I would have is that we doubling the memory footprint of the bounding box data structure -- which would be 4*4 extra bytes per net.

daveshah1:
Good thinking! Added in ad9bc62

PR #269, opened 2019-04-17
PLL GB usage and placement fix

smunaut:


daveshah1:
Can you have a look at the regression test failure for issue0127? Otherwise LGTM

smunaut:
Yup looking at it now

smunaut:
Yeah there is definitely something wrong with the second patch :/

smunaut:
Fixed now. If there is a unused net connected to one of the global output, it now gets removed.

PR #270, opened 2019-04-17
SB IO conflict checks

### LINKED ISSUE #268: Silent bug for two SB_IOs in one tile, one with clock enable, one without
**To reproduce**

clke_bug.v:

```
module clke_bug (
	input wire clk,
	output wire a,
	output wire b
);

reg d = 0;
reg e = 0;

always @ (posedge clk) begin
	e <= !e;
	d <= d ^ e;
end

// Without enable (a)
SB_IO #(
	.PIN_TYPE (6'b01_01_00), // normal registered output, OE always on
	.PULLUP (1'b 0)
) io_a (
	.OUTPUT_CLK   (clk),
	.D_OUT_0      (d),
	.PACKAGE_PIN  (a),
);

// With enable (b)
SB_IO #(
	.PIN_TYPE (6'b01_01_00),
	.PULLUP (1'b 0)
) io_b (
	.OUTPUT_CLK   (clk),
	.D_OUT_0      (d),
	.CLOCK_ENABLE (e),
	.PACKAGE_PIN  (b)
);

endmodule
```

clke_bug.pcf:

```
# For HX8k-EVN board

# on-board oscillator
set_io clk J3

# X20/Y0/io1:
set_io a N10
# X20/Y0/io0:
set_io b P10
```

clke_bug.sh:

```
#!/bin/bash

yosys -p "read_verilog clke_bug.v; synth_ice40 -json clke_bug.json"
nextpnr-ice40 --hx8k --package ct256 --pcf clke_bug.pcf --json clke_bug.json --asc clke_bug.asc
icepack clke_bug.asc clke_bug.bin
iceprog clke_bug.bin
```

I built icestorm and nextpnr-ice40 from latest github master before raising this.

**Expected output**

Two signals toggling at 90 degrees, or a build error

**Actual output**

Two signals toggling in phase

Found with help of @whitequark in [this thread](https://twitter.com/wren6991/status/1118246992189296645). Previously I've run up against the issue where I've driven different clock enables into different SB_IOs in the same IO tile, and nextpnr gives an error in this case.

However, in the above example, I'm driving a clock enable into just one SB_IO. By chance this is constrained to be in the same tile as a different SB_IO to which I have not connected a clock enable, so should be always enabled. What actually seems to happen is the clock enable is driven into both SB_IOs.

I understand from [this documentation](http://www.clifford.at/icestorm/bitdocs-8k/tile_18_0.html) that what I have asked nextpnr to do is not physically possible. Given this, though, it probably ought to give an error, as driving a clock enable into both SB_IOs completely changes the behaviour of the design.

smunaut:
Fixes #268

### DIFF
--- a/ice40/arch.cc
+++ b/ice40/arch.cc
@@ -1221,8 +1221,8 @@ void Arch::assignCellInfo(CellInfo *cell)
     } else if (cell->type == id_SB_IO) {
         cell->ioInfo.lvds = str_or_default(cell->params, id_IO_STANDARD, "SB_LVCMOS") == "SB_LVDS_INPUT";
         cell->ioInfo.global = bool_or_default(cell->attrs, this->id("GLOBAL"));
-        cell->ioInfo.pintype = int_or_default(cell->attrs, this->id("PIN_TYPE"));
-        cell->ioInfo.negtrig = bool_or_default(cell->attrs, this->id("NEG_TRIGGER"));
+        cell->ioInfo.pintype = int_or_default(cell->params, this->id("PIN_TYPE"));
+        cell->ioInfo.negtrig = bool_or_default(cell->params, this->id("NEG_TRIGGER"));
 
     } else if (cell->type == id_SB_GB) {
         cell->gbInfo.forPadIn = bool_or_default(cell->attrs, this->id("FOR_PAD_IN"));
--- a/ice40/arch_place.cc
+++ b/ice40/arch_place.cc
@@ -91,6 +91,18 @@ bool Arch::isBelLocationValid(BelId bel) const
     }
 }
 
+static inline bool _io_pintype_need_clk_in(unsigned pin_type) { return (pin_type & 0x01) == 0x00; }
+
+static inline bool _io_pintype_need_clk_out(unsigned pin_type)
+{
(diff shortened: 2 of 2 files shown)


Wren6991:
I built this PR, and it fixes my original problem, thanks!

However note that this now gives a build error, but is arguably valid:

```
// Without enable (a)
SB_IO #(
	.PIN_TYPE (6'b01_01_00), // normal registered output, OE always on
	.PULLUP (1'b 0)
) io_a (
	.OUTPUT_CLK   (clk),
	.D_OUT_0      (d),
	.PACKAGE_PIN  (a),
);

// With enable (b)
SB_IO #(
	.PIN_TYPE (6'b01_01_00),
	.PULLUP (1'b 0)
) io_b (
	.OUTPUT_CLK   (clk),
	.D_OUT_0      (d),
	.CLOCK_ENABLE (1'b1),
	.PACKAGE_PIN  (b)
);
```

This isn't hairsplitting; my code actually does this at the moment, as a fix for the original issue.

It also seems that connecting one CLOCK_ENABLE to a wrapper-module port which is then left disconnected at a higher level of hierarchy is treated differently from an SB_IO instantiation which does not connect CLOCK_ENABLE, and this too now gives a build error, where it previously did not.

Wren6991:
Ah, should have refreshed the page before commenting, sorry about that.

PR #272, opened 2019-04-27
Resolve difference in interpretation of TRELLIS_ROOT

jboone:
nextpnr appears to be out of date vs. prjtrellis's use of TRELLIS_ROOT. prjtrellis places libtrellis and pytrellis in a lib directory, and all other data in share/trellis. This patch adapts nextpnr accordingly.

I feel TRELLIS_ROOT might be better named TRELLIS_PREFIX, but that's up to @daveshah1, I imagine. And that would no longer match the ICEBOX_ROOT nomenclature.

Thanks to all the Yosys and SymbiFlow contributors for these wonderful tools!

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -5,7 +5,7 @@ task:
     memory: 20
     dockerfile: .cirrus/Dockerfile.ubuntu16.04
 
-  build_script: mkdir build && cd build && cmake .. -DARCH=all -DTRELLIS_ROOT=/usr/local/src/prjtrellis -DBUILD_TESTS=on && make -j $(nproc)
+  build_script: mkdir build && cd build && cmake .. -DARCH=all -DTRELLIS_ROOT=/usr/local -DBUILD_TESTS=on && make -j $(nproc)
   submodule_script: git submodule sync --recursive && git submodule update --init --recursive
   test_generic_script: cd build && ./nextpnr-generic-test
   test_ice40_script: cd build && ./nextpnr-ice40-test
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -5,7 +5,7 @@ task:
     memory: 20
     dockerfile: .cirrus/Dockerfile.ubuntu16.04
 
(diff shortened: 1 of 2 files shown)


eddiehung:
+1. I ran into this recently too, and Trellis' [README.md](https://github.com/SymbiFlow/prjtrellis/blame/5eb0ad870f30422b95d090ac9a476343809c62b9/README.md#L27) suggests you to install into `/usr` rather than `/usr/local`.
Looks like you broke CI though.

daveshah1:
I would prefer the parameter to be renamed TRELLIS_PREFIX of TRELLIS_INSTALL_PREFIX rather than TRELLIS_ROOT. This is because its behaviour is different to ICEBOX_ROOT. I also don't want any issues with CMake cached parameter values. 

The intention of TRELLIS_ROOT was/is to point to a in-tree built trellis git clone, not installed trellis, BTW. I would be open to suggestions to keeping an option that maintains this workflow option, e.g. by preserving TRELLIS_ROOT as an option. This will also keep existing scripts, etc, from breaking.

jboone:
@daveshah1 Thanks for chiming in. I don't have a strong opinion, because as you've no doubt seen from my last two commits (which failed CI), I am terrible at build systems. When there's a resolution about how the variables should work, I'll do my best to make them work and update this PR.

daveshah1:
Sorry for the delay in getting back to you, have been very busy with university exams.

Having thought about this some more, I think the best way forward is to keep `TRELLIS_ROOT` as is - intended to point to a Trellis clone - although perhaps deprecate it from documentation etc, so nothing breaks overnight. Then add `TRELLIS_INSTALL_PREFIX` as the preferred option going forward (with the behaviour that you have changed `TRELLIS_ROOT` to) for the install prefix when trellis is `make install`'d to a path.

If you get stuck with any CMake issues, happy to help further.

daveshah1:
Obsoleted by #392

PR #281, opened 2019-05-26
When choosing between labels for one net, always prefer ports

### LINKED ISSUE #280: Warning: net 'x' does not exist in design, ignoring clock constraint
The net (`clk12_0__i` in this case; [repro.zip](https://github.com/YosysHQ/nextpnr/files/3220488/repro.zip)) is definitely in the design, I can see it in the JSON file. In fact it is a toplevel port.

I think there are 2 nextpnr issues here:
  * nextpnr ignores impossible to apply clock constraint and proceeds (this should fail by default, like not passing timing was changed to fail recently);
  * nextpnr somehow loses the net.

whitequark:
1. Ports are already used in PCF and LPF files, so it is natural
that the developer would use them for constraints as well.
(Actually I think currently the LPF FREQUENCY PORT constraint
doesn't work correctly.)

2. Unpredictable (the set of netlabels nextpnr looks at depends on
the optimizations during synthesis) net naming makes it impossible
for code generators like nMigen to reliably apply clock constraints.

Fixes #280.

### DIFF
--- a/json/jsonparse.cc
+++ b/json/jsonparse.cc
@@ -692,11 +692,22 @@ void json_import(Context *ctx, string modname, JsonNode *node)
 
     log_info("Importing module %s\n", modname.c_str());
 
+    JsonNode *ports_parent = nullptr;
+    if (node->data_dict.count("ports") > 0)
+        ports_parent = node->data_dict.at("ports");
+
     // Multiple labels might refer to the same net. For now we resolve conflicts thus:
+    //  - (toplevel) ports are always preferred
     //  - names with fewer $ are always prefered
     //  - between equal $ counts, fewer .s are prefered
     //  - ties are resolved alphabetically
-    auto prefer_netlabel = [](const std::string &a, const std::string &b) {
+    auto prefer_netlabel = [ports_parent](const std::string &a, const std::string &b) {
+        if (ports_parent != nullptr) {
+            if (ports_parent->data_dict.count(a))
+                return true;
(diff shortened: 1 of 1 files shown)


daveshah1:
Thanks, agree with this change. Will merge once CI goes green.

whitequark:
This is a test bug. Test relies on unspecified behavior.
```patch
diff --git a/ice40/regressions/issue0084/mlaccel.sh b/ice40/regressions/issue0084/mlaccel.sh
index a0074be..76dfd09 100644
--- a/ice40/regressions/issue0084/mlaccel.sh
+++ b/ice40/regressions/issue0084/mlaccel.sh
@@ -1,3 +1,3 @@
 :> mlaccel.log
 ${NPNR} --json mlaccel.json --asc mlaccel.asc --pcf mlaccel.pcf `cat mlaccel.npnr` > mlaccel.log 2>&1 || true
-grep "^ERROR: PACKAGE_PIN of SB_IO 'qpi_in_buf\\[1\\]' connected to more than a single top level IO.$" mlaccel.log
+grep "^ERROR: PACKAGE_PIN of SB_IO 'qpi_in_buf\\[[01]\\]' connected to more than a single top level IO.$" mlaccel.log

```

daveshah1:
Thanks, applied patch to nextpnr-tests upstream 5182fd4. Can you bump the tests submodule in your PR to include this?

whitequark:
@daveshah1 Done

PR #284, opened 2019-06-01
Initial support for writing to json files from nextpnr.

mmicko:
Initial support for writing to json files from nextpnr.
Differentiate numbers and strings inside attributes and parameters (needed for using json again in yosys).
Store routing data as property (still not used in input)
Made nextpnr loading of json work fine (need to run pack/place to be visible but all data is preloaded)
Fixed issue in GUI so cells/nets trees are visible directly on load.

Please do not merge. Made for review and keeping track of progress.


common/command.cc
```diff
@@ -120,6 +121,7 @@ po::options_description CommandHandler::getGeneralOptions()
 
 #endif
     general.add_options()("json", po::value<std::string>(), "JSON design file to ingest");
+    general.add_options()("write", po::value<std::string>(), "JSON design file to write");
```
daveshah1: It would be a nice extra feature to be able to write JSON files after packing or placement as well as after routing - although perhaps this is best done as an option to disable routing altogether (for example) then just place, write JSON and exit.

common/command.cc
```diff
@@ -303,6 +305,12 @@ int CommandHandler::executeMain(std::unique_ptr<Context> ctx)
         customBitstream(ctx.get());
     }
 
+    if (vm.count("write")) {        
+        std::string filename = vm["write"].as<std::string>();
+        std::ofstream f(filename);
+        if (!write_json_file(f, filename, ctx.get()))
+            log_error("Loading design failed.\n");
```
daveshah1: Message should be "Saving design failed"

ice40/pack.cc
```diff
@@ -373,11 +375,10 @@ static void pack_constants(Context *ctx)
         ctx->cells[gnd_cell->name] = std::move(gnd_cell);
         ctx->nets[gnd_net->name] = std::move(gnd_net);
     }
-    // Vcc cell always inserted for now, as it may be needed during carry legalisation (TODO: trim later if actually
-    // never used?)
-    ctx->cells[vcc_cell->name] = std::move(vcc_cell);
-    ctx->nets[vcc_net->name] = std::move(vcc_net);
-
+    if (vcc_used) {
```
daveshah1: This change could cause carry packing to break, as it assumes the presence of the global Vcc net. Probably would be better to not add the cell and net if it already exists (which I presume is the real purpose of this change).

daveshah1:
Looks good!

One small concern is that certain arch-specific passes like global buffer insertion might not necessarily behave well processing their own packed output. Although most of these issues can be fixed, unnecessarily running packing/placing even if it does nothing will take some runtime on bigger designs. So I think it would be a good feature to store current design state (e.g. packed/placed/routed) so we don't run these passes again on a design that is already part run.

common/command.cc
mmicko: Added no-pack, no-place and no-route as options

daveshah1: LGTM

mmicko:
Read/write of constraints is not clean at all, this will be redone, needed to confirm that that data is actually was one needed.

common/placer1.cc
```diff
@@ -1021,7 +1022,7 @@ class SAPlacer
                 }
             }
 
-            if (ctx->timing_driven && int(pn->users.size()) < cfg.timingFanoutThresh) {
+            if (ctx->setting<bool>("timing_driven") && int(pn->users.size()) < cfg.timingFanoutThresh) {
```
daveshah1: This, and a few other similar places, will be close to the performance critical path in the placer - I'd really like to see thing faster for checking timing_driven here

gui/create_img.sh
```diff
@@ -2,5 +2,6 @@ convert -font helvetica -fill red -pointsize 8 -gravity center -draw "text 2,8 '
 convert -font helvetica -fill red -pointsize 8 -gravity center -draw "text 2,8 'PCF'" resources/open.png ice40/resources/open_pcf.png 
 convert -font helvetica -fill red -pointsize 8 -gravity center -draw "text 2,8 'BASE'" resources/open.png ecp5/resources/open_base.png 
 convert -font helvetica -fill red -pointsize 8 -gravity center -draw "text 2,8 'LPF'" resources/open.png ecp5/resources/open_lpf.png 
-convert -font helvetica -fill red -pointsize 8 -gravity center -draw "text 2,8 'ASC'" resources/save.png ice40/resources/save_asc.png 
-convert -font helvetica -fill red -pointsize 7 -gravity center -draw "text 2,8 'CONFIG'" resources/save.png ecp5/resources/save_config.png 
\ No newline at end of file
+convert -font helvetica -fill red -pointsize 8 -gravity center -draw "text 2,8 'ASC'" resources/save.png ice40/resources/save_asc.png
+convert -font helvetica -fill red -pointsize 7 -gravity center -draw "text 2,8 'CONFIG'" resources/save.png ecp5/resources/save_config.png
+convert -font helvetica -fill red -pointsize 8 -gravity center -draw "text 2,8 'JSON'" resources/save.png resources/save_json.png
```
daveshah1: Can you add a newline at the end here?

mmicko: fixed

common/placer1.cc
mmicko: fixed here, and added similar to placer_heap for consistency

PR #288, opened 2019-06-10
ice40: add RGB_DRV/LED_DRV_CUR support for u4k

corecode:


ice40/pack.cc
```diff
@@ -1125,14 +1140,20 @@ static void pack_special(Context *ctx)
...
                     continue;
 
                 if (net->users.size() > 0)
-                    log_error("SB_RGBA_DRV port connected to more than just package pin !\n");
+                    log_error("SB_RGB_DRV/SB_RGBA_DRV port connected to more than just package pin !\n");
 
                 ctx->nets.erase(net->name);
             }
+
+            if (is_sb_rgb_drv(ctx, ci) && !ci->ledInfo.ledCurConnected)
```
daveshah1: As far as I can tell, there is no guarantee that the SB_LED_DRV_CUR will be visited before the SB_RGB_DRV, so this could fail with some input netlists (probably depending on small changes to the Verilog affecting the order of cells)

corecode: the cells are iterated in sorted (string index) order - i think.

daveshah1: It's a bit more icky than that, they are sorted in IdString order, which will reflect the order in which IdStrings are created, ultimately reflecting the order in which cells were processed while reading the JSON. Yosys doesn't really guarantee the order of cells written in the JSON.

corecode: Oh, I thought the initial IdStrings are created from constids.  If not, how do you suggest we handle this?

daveshah1: That covers IdStrings used for build time constants like cell types, but not runtime things like cell names (which we sort by here). I would add another loop to deal with the CUR first.

PR #289, opened 2019-06-17
Major rewrite & speedup of timing analysis

daveshah1:
This forms the final part of my university final year project. It aims to avoid rebuilding all the timing data structures ever time STA is invoked, allowing faster updates during placement and routing and multithreaded timing analysis (calculating arrival/required times in their own thread and clock domains individually). The aim is to replace the existing timing analysis as much as possible, it's not quite there yet.

Remaining TODOs:
 - [ ] fix conflicts
 - [ ] add critical path printing (including more advanced timing reports like _N_ most critical start/endpoints)
 - [ ] add cross clock max delay constraints
 - [ ] replace all users of the old timing analysis (subject to discussion)

Also included in this PR are some improvements to the Python constraints API, although they don't have any function yet.


gatecat:
Closing now that a rewrite-of-the-rewrite is now in progress...

PR #293, opened 2019-06-21
Add support for upto and offset in JSON files

mmicko:
Depends on : https://github.com/YosysHQ/yosys/pull/1124

### DIFF
--- a/json/jsonparse.cc
+++ b/json/jsonparse.cc
@@ -337,7 +337,7 @@ static int const_net_idx = 0;
 template <typename F>
 void json_import_ports(Context *ctx, const string &modname, const std::vector<IdString> &netnames,
                        const string &obj_name, const string &port_name, JsonNode *dir_node, JsonNode *wire_group_node,
-                       F visitor)
+                       bool upto, int start_offset, F visitor)
 {
     // Examine a port of a cell or the design. For every bit of the port,
     // the connected net will be processed and `visitor` will be called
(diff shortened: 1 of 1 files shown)


json/jsonparse.cc
```diff
@@ -406,8 +406,11 @@ void json_import_ports(Context *ctx, const string &modname, const std::vector<Id
             wire_node = wire_group_node->data_array[index];
             //
             // Pick a name for this port
+            int ndx = index + start_offset;
+            if (!upto)
```
daveshah1: This looks like the wrong way round to me - when upto isn't set, we surely want to keep the old behaviour?

mmicko: Agree, but in that case yosys should always set "upto" to 1 or 0, right now we can not distinguish between 0 and old behavior. cc: @cliffordwolf

daveshah1: We should certainly not be silently breaking JSON files from older Yosys

json/jsonparse.cc
```diff
@@ -732,6 +747,18 @@ void json_import(Context *ctx, string modname, JsonNode *node)
 
             here = cell_parent->data_dict.at(cell_parent->data_dict_keys[nnid]);
             std::string basename = cell_parent->data_dict_keys[nnid];
+            bool upto = false;
+            int start_offset = 0;
+            if (here->data_dict.count("upto") != 0) {
+				JsonNode *val = here->data_dict.at("upto");
+				if (val->type == 'N')
+					upto = val->data_number != 0;
+			}
```
cliffordwolf: This indenting doesn't look right..  ;)

mmicko: Fixed now. Thanks for pointing out to this.

PR #306, opened 2019-08-01
Major improvements to constants/Property

daveshah1:
This is designed to mirror the JSON changes in YosysHQ/yosys#1241, and at the same time improve the handling and round-trip-ability of constants inside nextpnr.

Currently done:
 - Common core and iCE40 arch
 - ECP5
 - Generic

Still TODO:
 - serious regression testing


cliffordwolf:
I don't know if I really like having an `RTLIL::Const` in nextpnr. Why not just extend/change `Property` as needed?

daveshah1:
Ultimately, I think the interfaces we want are are the same - and I wanted to rename it so every old use is caught. It would be easy enough to pick a different name though.

PR #313, opened 2019-08-14
cmakelists: quote the last argument to REGEX REPLACE

xobs:
With newer versions of cmake, an unquoted 6th argument gives the
following error:

    STRING sub-command REGEX, mode REPLACE needs at least 6 arguments
    total to command.

Quoting this fixes the issue.

Signed-off-by: Sean Cross <sean@xobs.io>


xobs:
See https://github.com/facebook/rocksdb/pull/1231 for similar issues that were fixed in the same way

mmicko:
LGTM, tried on my build environment and works fine. 
@daveshah1 ok on your end ?

daveshah1:
LGTM, thanks

PR #314, opened 2019-08-15
cmake: fix static build on Windows

xobs:
On Windows, we can make a static build by setting the triple to
x64-windows-static, but due to some issues with cmake we also need to
modify the linker flags and Python definitions.

This patch also expands the check for `USE_STATIC` to not simply add the `-static` flag on Windows, which doesn't accomplish the desired goal there.

There are references to this requirement in the documentation and [elsewhere](https://stackoverflow.com/a/35440164).

With this patch I am now able to build static Windows binaries.


CMakeLists.txt
```diff
@@ -15,7 +15,14 @@ option(SERIALIZE_CHIPDB "Never build chipdb in parallel to reduce peak memory us
 set(link_param "")
 if (STATIC_BUILD)
     set(Boost_USE_STATIC_LIBS   ON)
-    if(NOT ${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+    if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+    elseif(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
```
mmicko: Please check here if  it is MSVC otherwise it will break mingw build

xobs:
I've added a guard, as requested.

Note that the static build doesn't actually work, and I'm not sure why.  When using a `--pre-pack` script, `nextpnr-ice40.exe` exits with code 0xC0000409.  It's fine when not using python.  This may be due to `nextpnr-ice40.exe` doing some re-initialization, or due to a bug in boost::python.  However that is a separate issue.

PR #316, opened 2019-08-21
cmake: add python38 to list of versions searched for boost components

gsomlo:
Also, factor out actual search code (identical across versions).

Signed-off-by: Gabriel Somlo <gsomlo@gmail.com>


gsomlo:
This enables building for Fedora 31/32 which ships with Python 3.8 (see https://bugzilla.redhat.com/show_bug.cgi?id=1743893).
This is the same fix as previously applied to prjtrellis here: https://github.com/SymbiFlow/prjtrellis/pull/89

daveshah1:
Thanks!

PR #317, opened 2019-08-21
Restrict clock promotion to global on ECP5

DurandA:
This PR allows optional disabling of the promotion to global of a clock using the `(* ECP5_IS_GLOBAL=0 *)` attribute.

This allows ring oscillators to be used as clock sources. When creating a ring oscillator with Yosys+NextPNR or Diamond and adding it to the sensitivity list, the ring oscillator stops oscillating. This was described in DurandA/verilog-buildingblocks#1.

To the best of my knowledge, this is not possible in Diamond as I couldn't find such a feature in the chapter _Specifying Constraints and Attributes_ from _Lattice Synthesis Engine for Diamond User Guide_. In this regards (and considering that the use-cases are limited), I don't necessarily expect this PR to be accepted.

### DIFF
--- a/ecp5/globals.cc
+++ b/ecp5/globals.cc
@@ -435,8 +435,8 @@ class Ecp5GlobalRouter
         log_info("Promoting globals...\n");
         auto clocks = get_clocks();
         for (auto clock : clocks) {
-            bool is_global = bool_or_default(clock->attrs, ctx->id("ECP5_IS_GLOBAL"), true);
-            if (!is_global)
+            bool is_noglobal = bool_or_default(clock->attrs, ctx->id("noglobal"), false);
+            if (is_noglobal)
                 continue;
             log_info("    promoting clock net %s to global network\n", clock->name.c_str(ctx));
             if (is_ooc) // Don't actually do anything in OOC mode, global routing will be done in the full design
--- a/ecp5/globals.cc
+++ b/ecp5/globals.cc
@@ -435,6 +435,9 @@ class Ecp5GlobalRouter
         log_info("Promoting globals...\n");
         auto clocks = get_clocks();
         for (auto clock : clocks) {
+            bool is_global = bool_or_default(clock->attrs, ctx->id("ECP5_IS_GLOBAL"), true);
+            if (!is_global)
+                continue;
             log_info("    promoting clock net %s to global network\n", clock->name.c_str(ctx));
             if (is_ooc) // Don't actually do anything in OOC mode, global routing will be done in the full design
                 clock->is_global = true;


eddiehung:
A general comment I have would be (since you've not found a vendor equivalent) would be  to consider a generic attribute that can be shared across all architectures? Did you also take a look to see if there is a `syn_*` constraint out there too?

daveshah1:
The `ECP5_IS_GLOBAL` attribute is intended for internal use, weird things could happen if it is also used with different meaning by end users. Please use a different name for this attribute (how about `(* noglobal *)`?

DurandA:
> A general comment I have would be (since you've not found a vendor equivalent) would be to consider a generic attribute that can be shared across all architectures?

Please excuse my ignorance, are global nets common to all FPGA architectures? 

> Did you also take a look to see if there is a `syn_*` constraint out there too?

If you are referring to pages 31-74 of [_Lattice Synthesis Engine for Diamond User Guide_](http://www.latticesemi.com/~/media/LatticeSemi/Documents/UserManuals/JL/LSEUGDiamond.pdf?document_id=51556), that is what I looked for. I quickly looked at [_Vivado Design Suite User Guide: Using Constraints_](https://www.xilinx.com/support/documentation/sw_manuals/xilinx2018_1/ug903-vivado-using-constraints.pdf) and I think that this feature is equivalent to `CLOCK_DEDICATED_ROUTE` but I don't know anything about the Xilinx vendor tools.

> The `ECP5_IS_GLOBAL` attribute is intended for internal use, weird things could happen if it is also used with different meaning by end users. Please use a different name for this attribute (how about `(* noglobal *)`?

From what I understood from the code, this attribute was used in the exact same context when a _NetInfo_ is promoted to global, hence my decision to reuse it. I'm perfectly fine with `(* noglobal *)`; @eddiehung are you as well?

dpiegdon:
`(* noglobal *)` sound good to me, and i'd be very happy to have an attribute like that!

eddiehung:
:+1: for `noglobal` as a generic attribute.
Architecture-specific backends can then support whatever attributes the vendors also use, but this attribute could be `portable` one for all arches.

> Please excuse my ignorance, are global nets common to all FPGA architectures?
No excuses necessary! But yes, global nets are very common across all FPGA vendors, primarily for low-skew clock routing -- ideally, a clock edge appears at the lower left corner of the chip at the same time as in the upper right corner and everywhere in between.

DurandA:
I just pushed DurandA/nextpnr@a26c9bb6d9aaa78fd3deae94e3db840286dacf8a to use `(* noglobal *)` attribute. Many thanks!

PR #330, opened 2019-09-15
bba: Default to native endian in bbasm

zeldin:
`bbasm` required endianness to be specified explicitly (except that explicit LE was indicated by omitting the `-b` option...).  The cmake files failed to do so, which resulted in broken builds on big endian hosts.

Rather than mucking around with the cmake files, I made native endian the default and added a new option flag for the case where LE is explicitly needed.

Tested on ppc64 (big endian) and aarch64 (little endian).


zeldin:
Actually, that doesn't work too well with cross compilation.  Better make the endianness option mandatory and fix the cmake files...

zeldin:
Going forward, the `chipdbs` directories should be created in the build tree rather than the source tree when making an out-of-tree build.  That will prevent mixup of files generated for different arches.

zeldin:
There we go.  `.bba` files in the source tree, since they can be shared between builds for different platforms, and `.cc` and `.bin` files in the build tree since they can not.

Ideally maybe this should be moved in under `generated`, but I skipped that for now to avoid changing the layout of in-tree builds.

PR #340, opened 2019-10-11
ecp5: IOLOGIC improvements

daveshah1:
This implements most of the missing IOLOGIC features, in particular:

 - 7:1 gearboxes (ODDR71B and IDDR71B)
 - IO (and tristate) flip flops, using the `syn_useioff` attribute (same as Diamond) to enable packing a TRELLIS_FF into the IOLOGIC. Tristate flip flops can be combined with output DDR
 - ECLKBRIDGECS for edge clocks that span the chip

See SymbiFlow/prjtrellis#107


whitequark:
This is extremely nice!

whitequark:
How is the combination of a IODDRX1F for i/o and OFS1P3DX for OE handled?

daveshah1:
It will be packed unless there is a mismatch between clock or reset signal; then it would be an error.

whitequark:
Excellent. I have an option of packing an SDR or DDR register into OE site, right?

daveshah1:
Only an SDR one, I'm not sure if DDR tristate (outside of memory DQS mode) is even supported by hardware

whitequark:
Ah OK. It's interesting that some vendors support it and some do not. Looks like Xilinx and Altera do.

PR #341, opened 2019-10-13
ice40: Add set_frequency pcf command; and document pcf format

daveshah1:
Right now the lack of any other timing constraint format limits builds not linked with Python. This should make useful static/distributable builds much smaller and simpler than at present.

Strictly speaking, the vendor-compatible way would be to use sdc files. But this is a bigger project as it applies to multiple architectures and would need more research into a reasonable subset to support (and any IP issues that might affect it). It might even be best to approach that from the Yosys side; as eventually synthesis-time constraints will be needed too (this could then be passed to nextpnr in an attribute).

### DIFF
--- a/docs/ice40.md
+++ b/docs/ice40.md
@@ -1,0 +1,17 @@
+# iCE40 Architecture Documentation
+
+## PCF format reference
+
+PCF files contain physical constraints and are specified using the `--pcf` argument. Each (non blank) line contains a command; lines beginning with `#` are comments.
+
+Two commands are supported: `set_io` and `set_frequency`.
+
+    set_io [-nowarn] [-pullup yes|no] [-pullup_resistor 3P3K|6P8K|10K|100K] port pin
+
+Constrains named port `port` to package pin `pin`. `-nowarn` disables the warning if `port` does not exist. `-pullup yes` can be used to enable the built in pullup for all iCE40 devices. `-pullup_resistor` sets the pullup strength, and is available on iCE40 UltraPlus only.
+
+    set_frequency net frequency
+
+Adds a clock constraint to a named net (any alias for the net can be used). `frequency` is in MHz.
+
+_Note that this is a non-standard extension, not supported by the vendor toolchain. It allows specifying clock constraints without needing the Python API._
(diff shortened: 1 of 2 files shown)


whitequark:
> It might even be best to approach that from the Yosys side; as eventually synthesis-time constraints will be needed too (this could then be passed to nextpnr in an attribute).

I believe vendor tools usually parse SDC files twice. In some cases it's even stranger, like Diamond uses both SDC and LPF, but IIRC it generates LPF from SDC if you specify it; they have an extremely confusing diagram explaining this.

whitequark:
cc @xobs

mithro:
FYI vtr uses https://github.com/kmurray/libsdcparse and supports the timing constraints described here @ https://docs.verilogtorouting.org/en/latest/vpr/timing_constraints/ and https://docs.verilogtorouting.org/en/latest/vpr/sdc_commands/

whitequark:
I think using anything short of real Tcl would be short-selling nextpnr compared to vendor tools, especially given how easy Tcl is to embed (unlike Python).

daveshah1:
Eventually, we need support for xdc (Xilinx's constraint format) that is another Tcl DSL similar to sdc. So there could probably be some reuse there too.

whitequark:
XDC is interestingly a quite restricted subset in that it admits no control flow, though you can still simulate some of it using its commands. nMigen uses constructs like the following in the Vivado Tcl script, which would be nice to be able to do in nextpnr without dragging Python around.

```
            foreach cell [get_cells -quiet -hier -filter {nmigen.vivado.false_path == "TRUE"}] {
                set_false_path -to $cell
            }
            foreach cell [get_cells -quiet -hier -filter {nmigen.vivado.max_delay != ""}] {
                set clock [get_clocks -of_objects \
                    [all_fanin -flat -startpoints_only [get_pin $cell/D]]]
                if {[llength $clock] != 0} {
                    set_max_delay -datapath_only -from $clock \
                        -to [get_cells $cell] [get_property nmigen.vivado.max_delay $cell]
                }
            }
```

xobs:
Is there a similar feature for `nextpnr-ecp5` that we could use to avoid pulling in Python?

daveshah1:
> Is there a similar feature for nextpnr-ecp5 that we could use to avoid pulling in Python?

Yes, you can use the same syntax in the lpf as the vendor tools - 

    FREQUENCY NET "eth_clocks0_rx" 125.0 MHz;

or

    FREQUENCY PORT "eth_clocks0_rx" 125.0 MHz;

in fact I think LiteX does this already

PR #345, opened 2019-10-23
Improve handling of top level IO and add SDF support

daveshah1:
SDF timing simulation requires some changes to the Yosys cell library that are still being worked on. Furthermore a closed source simulator (e.g. xsim) seems to be required; iverilog's SDF support seems to be too limited, in particular not supporting interconnect delays.

### DIFF
--- a/common/command.cc
+++ b/common/command.cc
@@ -150,6 +150,7 @@ po::options_description CommandHandler::getGeneralOptions()
     general.add_options()("timing-allow-fail", "allow timing to fail in design");
     general.add_options()("no-tmdriv", "disable timing-driven placement");
     general.add_options()("sdf", po::value<std::string>(), "SDF delay back-annotation file to write");
+    general.add_options()("sdf-cvc", "enable tweaks for SDF file compatibility with the CVC simulator");
 
     return general;
 }
@@ -343,7 +344,7 @@ int CommandHandler::executeMain(std::unique_ptr<Context> ctx)
         std::ofstream f(filename);
         if (!f)
             log_error("Failed to open SDF file '%s' for writing.\n", filename.c_str());
-        ctx->writeSDF(f);
+        ctx->writeSDF(f, vm.count("sdf-cvc"));
     }
 
 #ifndef NO_PYTHON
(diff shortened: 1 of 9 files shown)


mithro:
FYI - There is a libsdfparse library at https://github.com/kmurray/libsdfparse if you end up needing one.

common/command.cc
```diff
@@ -149,6 +149,9 @@ po::options_description CommandHandler::getGeneralOptions()
     general.add_options()("freq", po::value<double>(), "set target frequency for design in MHz");
     general.add_options()("timing-allow-fail", "allow timing to fail in design");
     general.add_options()("no-tmdriv", "disable timing-driven placement");
+    general.add_options()("sdf", po::value<std::string>(), "SDF delay back-annotation file to write");
+    general.add_options()("sdf-cvc", "enable tweaks for SDF file compatibility with the CVC simulator");
```
mithro: Just FYI - From what I can see CVC is not **actually** open source. It is more of a much more limited "shared source" style (I was super excited for a minute :-().

daveshah1: It is open source - just a bit hard to download - as far as I can see. But it also has several bugs that stop it from being useful here (very fussy about names in the SDF, incorrect handling of disconnected ports, and possibly others) - and the open source  version is abandoned for all intents and purposes. But there are some GH mirrors like https://github.com/cambridgehackers/open-src-cvc

mithro: From http://www.tachyon-da.com/wp-content/uploads/2016/04/OSS-CVC-MODIFIED-ARTISTIC-LIC.txt

> In particular OSS CVC is intended to use the dual licensing model so any
> commercial redistribution of a modified version, a forked project, a
> commercial product version of OSS CVC, or a version linked with
> commercial closed source software will require other distribution
> arrangements and license with the Copyright Holder.  All version of
> OSS CVC may not be sold by providing access on a cloud and charging 
> a cloud usage fee.

Pretty sure that set of restrictions fails the [Open Source Definition](https://opensource.org/osd-annotated).

PR #346, opened 2019-10-24
ice40: cmake: fix build with pregenerated bba path

xobs:
When building using non-pregenerated bba files, the rule to create bbasm
files gets called twice: once unconditionally, and once as part of the
conditional that determines we're not using a pregenerated bba path.

If we _are_ using a pregenerated bba path, then this rule gets called
anyway, resulting in a build error.

Remove the duplicate, unconditional creation of the bba file generation,
to fix the build when using pregenerated files, and to speed up the
build when not using pregenerated files.


daveshah1:
Oops, thanks for fixing!

PR #350, opened 2019-11-08
Dedicated output for LUT in GENERIC_SLICE

pepijndevos:
Currently the GENERIC_SLICE has one Q output that is either used by the DFF output or by the LUT output. However, it appears that most FPGA architectures have a seperate output for the LUT that bypasses the DFF.

This PR adds an F output to GENERIC_SLICE that is the direct output of the LUT, and removes the FF_USED parameter.

### DIFF
--- a/generic/cells.cc
+++ b/generic/cells.cc
@@ -44,13 +44,13 @@ std::unique_ptr<CellInfo> create_generic_cell(Context *ctx, IdString type, std::
     if (type == ctx->id("GENERIC_SLICE")) {
         new_cell->params[ctx->id("K")] = std::to_string(ctx->args.K);
         new_cell->params[ctx->id("INIT")] = 0;
-        new_cell->params[ctx->id("FF_USED")] = 0;
 
         for (int i = 0; i < ctx->args.K; i++)
             add_port(ctx, new_cell.get(), "I[" + std::to_string(i) + "]", PORT_IN);
 
         add_port(ctx, new_cell.get(), "CLK", PORT_IN);
 
+        add_port(ctx, new_cell.get(), "F", PORT_OUT);
         add_port(ctx, new_cell.get(), "Q", PORT_OUT);
     } else if (type == ctx->id("GENERIC_IOB")) {
         new_cell->params[ctx->id("INPUT_USED")] = 0;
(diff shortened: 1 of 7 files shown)


generic/cells.cc
```diff
@@ -80,18 +80,16 @@ void lut_to_lc(const Context *ctx, CellInfo *lut, CellInfo *lc, bool no_dff)
     }
 
     if (no_dff) {
-        replace_port(lut, ctx->id("Q"), lc, ctx->id("Q"));
-        lc->params[ctx->id("FF_USED")] = 0;
```
daveshah1: Might as well keep the FF_USED parameter, for devices where it matters

pepijndevos: Hmmm, true I guess.

generic/synth/prims.v
```diff
@@ -21,21 +21,19 @@ endmodule
...
-	wire lut_q;
-	LUT #(.K(K), .INIT(INIT)) lut_i(.I(I), .Q(lut_q));
+	LUT #(.K(K), .INIT(INIT)) lut_i(.I(I), .Q(f_wire));
 
-	generate if (FF_USED)
-		DFF dff_i(.CLK(CLK), .D(lut_q), .Q(Q));
-	else
-		assign Q = lut_q; 
-	endgenerate
+  DFF dff_i(.CLK(CLK), .D(f_wire), .Q(Q));
```
daveshah1: Something looks up with the indentation here

generic/examples/simple.sh
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
 set -ex
 yosys -p "tcl ../synth/synth_generic.tcl 4 blinky.json" blinky.v
-${NEXTPNR:-../../nextpnr-generic} --pre-pack simple.py --pre-place simple_timing.py --json blinky.json --post-route bitstream.py
+${NEXTPNR:-../../nextpnr-generic} --pre-pack simple.py --pre-place simple_timing.py --json blinky.json --post-route bitstream.py --write pnrblinky.json
+yosys -p "read_verilog -lib ../synth/prims.v; read_json pnrblinky.json; dump -o blinky.il; show -format png -prefix blinky"
```
daveshah1: Can you make sure these new output products are gitignore'd

generic/cells.cc
```diff
@@ -80,18 +80,16 @@ void lut_to_lc(const Context *ctx, CellInfo *lut, CellInfo *lc, bool no_dff)
...
 }
 
 void dff_to_lc(const Context *ctx, CellInfo *dff, CellInfo *lc, bool pass_thru_lut)
 {
-    lc->params[ctx->id("FF_USED")] = 1;
     replace_port(dff, ctx->id("CLK"), lc, ctx->id("CLK"));
 
     if (pass_thru_lut) {
-        lc->params[ctx->id("INIT")] = 2;
+        lc->params[ctx->id("INIT")] = 0xAAAA;
```
daveshah1: This does make sense for architectures where the LUT inputs aren't default-0, but this needs to be generic across all LUT K (including the case of K>6 where an int64 can't even be used)

generic/synth/prims.v
pepijndevos: Oh no, prepare for a bazillion commits where I think I fixed it and it turns out I made it worse.

generic/cells.cc
pepijndevos: Ouch, yes. Uhm... so should I revert it? Make it 64 bits? Make it some vector type?

daveshah1: Have a look at how Property works - it already is a vector type, the int thing is just a convenient constructor

PR #353, opened 2019-11-18
New hierarchy-capable generic frontend framework and json11 based JSON frontend

daveshah1:
This replaces the current hand-written; JSON specific frontend with two new components:

1. A hierarchy aware generic frontend framework; that calls a "frontend API" to access any format that provides data in a structure roughly isomorphic to the current JSON format but without excessive overhead - this should make a protobuf/flatbuffer frontend much much easier. Being hierarchy aware it can also handle unflattened hierarchical netlists for floorplanning etc. At the moment it just flattens them; I plan to work on some structures to store hierarchy for use during packing, placement and for constraints/PR

2. A JSON implementation of the generic frontend; but using json11 rather than a handwritten parser. Even if this might be slightly less performant (I haven't benchmarked it yet); a binary frontend in the near future using flatbuffers or whatever will realistically blow either JSON parser out of the water and will be the recommended route for big netlists/checkpoints.


whitequark:
:+1:

PR #356, opened 2019-11-20
ecp5: Improve flipflop packing density

daveshah1:


tomverbeure:
Is this commit good enough for testing?

If so, I’ll have some homework for tonight. :-)

daveshah1:
Not quite, but hopefully by tonight

daveshah1:
It's probably at a state now where it's worth a quick test, at least. I can't 100% guarantee that a very dense design will actually route successfully, but that's another problem in itself.

PR #361, opened 2019-11-30
Add bba #embed support

cliffordwolf:
Disabled by default. Use `cmake -DUSE_C_EMBED=1` to enable. `#embed_str` is part of [P1040](https://thephd.github.io/vendor/future_cxx/papers/d1040.html).

### DIFF
--- a/bba/main.cc
+++ b/bba/main.cc
@@ -445,9 +445,9 @@ int main(int argc, char **argv)
         for (auto &s : preText)
             fprintf(fileOut, "%s\n", s.c_str());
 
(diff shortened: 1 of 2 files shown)


cliffordwolf:
I'd like to merge this in the hope that it may help the effort to get something like `#embed_str` in future C/C++ standards. Long-term this may allow us to use the same method for embedding the binary blob on all platforms. For now it's just disabled by default, so this should not have any effect on builds without `-DUSE_C_EMBED=1`.

PR #367, opened 2019-12-06
Add missing header to main.cc

### LINKED ISSUE #366: basename not declared in bba/main.c
When I try to build for ice40 I get the following error
```
bba/main.cc:450:49: error: use of undeclared identifier
      'basename'
        fprintf(fileOut, "#embed_str \"%s\"\n", basename(bin_basename_buf));

```
I "fixed" the issue by including libgen.h in bba/main.c

twam:
This fixes building with clang on OS X.

Fixes #366.

### DIFF
--- a/bba/main.cc
+++ b/bba/main.cc
@@ -28,6 +28,7 @@
 #include <string.h>
 #include <string>
 #include <vector>
+#include <libgen.h>
 
 enum TokenType : int8_t
 {


daveshah1:
Thanks for the PR, but I think #368 is a better fix for the same problem (will solve Windows too)

PR #380, opened 2020-01-13
Remove barrier to out-of-tree builds

ldoolitt:
With this change, nextpnr can build successfully (and run correctly) using the git tree as a read-only reference.
No change in behavior observed for in-tree builds.


ldoolitt:
P.S. Thanks to the nextpnr developers for making reproducible bitfiles!  When I say that nextnpnr runs succesfully, it's unambiguous, because the bitfiles that result from my demo projects come out bit-for-bit identical. Of course (given the file that this patch touches), I'm talking about nextpnr-ecp5.

PR #382, opened 2020-01-15
ecp5: Add support for top pseudo diff outputs

daveshah1:
Set IO_TYPE to LVCMOS33D (for example) on the "A" side (nothing needed on the "B" side) to use


ktemkin:
I'll test this in a short bit (as requested), but I wanted to ask:

Is there a reason not to apply this same logic (and equivalent fuzzing) for the bottom tiles? TN1262 (sysIO) suggests that this same structure should be present on the bottom tiles, as well.

ktemkin:
Tested on hardware; works great. :+1:

daveshah1:
Thanks for testing. I need to have a general look at bottom IO as there as some inconsistencies between Trellis and diamond, I think due to issues with the clamp being enabled by default on config pins.

PR #389, opened 2020-02-02
Add spent time info to report

mmicko:


common/router1.cc
```diff
@@ -769,14 +769,19 @@ bool router1(Context *ctx, const Router1Cfg &cfg)
         int last_arcs_with_ripup = 0;
         int last_arcs_without_ripup = 0;
 
-        log_info("           |   (re-)routed arcs  |   delta    | remaining\n");
-        log_info("   IterCnt |  w/ripup   wo/ripup |  w/r  wo/r |      arcs\n");
+        log_info("           |   (re-)routed arcs  |   delta    | remaining|       time spent      |\n");
+        log_info("   IterCnt |  w/ripup   wo/ripup |  w/r  wo/r |      arcs| batch(sec)| total(sec)|\n");
```
daveshah1: nitpick: there is no `|` between  `w/ripup   wo/ripup`; but there is between `batch(sec)| total(sec)`, would be nice to be consistent

mmicko: sure, updated it now

PR #392, opened 2020-02-03
Fix paths to prjtrellies project

se-bi:
Rebased, updated and squashed commits by @jboone on jboone/nextpnr@246fff9
(also mentioned in #272)

Necessary, if prjtrellis or rather libtrellis is install to a custom install path (e.g. `/opt/prjtrellis/`),
then `cmake -DARCH=ecp5 -DTRELLIS_ROOT=/opt/prjtrellis` did not found pytrellis library etc.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -5,7 +5,7 @@ task:
     memory: 20
     dockerfile: .cirrus/Dockerfile.ubuntu16.04
 
-  build_script: mkdir build && cd build && cmake .. -DARCH=all -DTRELLIS_ROOT=/usr/local/src/prjtrellis -DBUILD_TESTS=on && make -j $(nproc)
+  build_script: mkdir build && cd build && cmake .. -DARCH=all -DBUILD_TESTS=on && make -j $(nproc)
   submodule_script: git submodule sync --recursive && git submodule update --init --recursive
   test_generic_script: cd build && ./nextpnr-generic-test
   test_ice40_script: cd build && ./nextpnr-ice40-test
--- a/.cirrus/Dockerfile.ubuntu16.04
+++ b/.cirrus/Dockerfile.ubuntu16.04
@@ -47,7 +47,7 @@ RUN set -e -x ;\
     cd /usr/local/src ;\
     git clone --recursive https://github.com/SymbiFlow/prjtrellis.git ;\
     cd prjtrellis ;\
-    git reset --hard 46e95314be7f8850db80ad1ef6b3359b091fad93 ;\
+    git reset --hard 668ce3492cbe1566c61760f06bdf676f6fb265c3 ;\
     cd libtrellis ;\
     cmake -DCMAKE_INSTALL_PREFIX=/usr . ;\
     make -j $(nproc) ;\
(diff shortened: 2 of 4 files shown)


daveshah1:
What about the comments I left on #272?

> I would prefer the parameter to be renamed TRELLIS_PREFIX of TRELLIS_INSTALL_PREFIX rather than TRELLIS_ROOT. This is because its behaviour is different to ICEBOX_ROOT. I also don't want any issues with CMake cached parameter values.

> The intention of TRELLIS_ROOT was/is to point to a in-tree built trellis git clone, not installed trellis, BTW. I would be open to suggestions to keeping an option that maintains this workflow option, e.g. by preserving TRELLIS_ROOT as an option. This will also keep existing scripts, etc, from breaking.

> Sorry for the delay in getting back to you, have been very busy with university exams.

> Having thought about this some more, I think the best way forward is to keep TRELLIS_ROOT as is - intended to point to a Trellis clone - although perhaps deprecate it from documentation etc, so nothing breaks overnight. Then add TRELLIS_INSTALL_PREFIX as the preferred option going forward (with the behaviour that you have changed TRELLIS_ROOT to) for the install prefix when trellis is make install'd to a path.

> If you get stuck with any CMake issues, happy to help further.

I am relatively happy to change the behaviour if there is a good consensus though, and you are willing to have a quick check for places that this might break?

se-bi:
Okay, little rework on that.

I renamed TRELLIS_ROOT to TRELLIS_INSTALL_PREFIX. I think this convenient, because it follows the naming when configuring that build with -DCMAKE_INSTALL_PREFIX.
I did not keep any fall back of TRELLIS_ROOT, nowhere else used in this project.
The CI does not use any more the INSTALL_PREFIX for prjtrellis.

Other CI/Docker projects: Only checked ghdl/docker@7b7b1f9, at least they do not use TRELLIS_ROOT, but also not default destination...

Sorry, did not thoroughly check the comments of #272,
first intention was to build against upstream of prjtrellis and represent that by the CI.

daveshah1:
LGTM, thanks, will merge once CI goes green.

se-bi:
Thanks too,
might close obsolete #272,

also follow-up in prjtrellis is coming :wink:

cr1901:
I wish I saw this issue earlier. By removing `TRELLIS_ROOT`, does this mean that people who want to hack on nextpnr against a more recent database must now run `make install` for each change they want to test in the DB, or can `TRELLIS_INSTALL_PREFIX` be use in the same manner as `TRELLIS_ROOT` to avoid numerous `make install`s?

se-bi:
Did not thought about pointing to the build instead install path for developing,
the procedure for *ice40* also works to point `ICESTORM_ROOT` to the install path.

From the perspective of "just" using nextpnr for projects,
therefore building the toolchain for both architectures,
one is instructed to build and install *icestorm*,
thus the Readme is not more comprehensive for *ecp5*,
i assumed to go the same way, build and install *prjtrellis*.

i would suggest to add an option for developing purposes,
add a note under "Notes for developer"
and keep the normal way for beginners and user straight forward working.

PR #406, opened 2020-03-04
Update README with TRELLIS_INSTALL_PREFIX example

gojimmypi:
see https://github.com/YosysHQ/nextpnr/pull/404 and https://github.com/SymbiFlow/prjtrellis/pull/123

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ such as pack, place, route, and write output files.)
 
 For ECP5 support, you must download [Project Trellis](https://github.com/SymbiFlow/prjtrellis),
 then follow its instructions to download the latest database and build _libtrellis_. 
-(for example: `-DTRELLIS_INSTALL_PREFIX=/usr` will install to `/usr/share/trellis`) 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -97,7 +97,8 @@ such as pack, place, route, and write output files.)
 ### nextpnr-ecp5
 
 For ECP5 support, you must download [Project Trellis](https://github.com/SymbiFlow/prjtrellis),
-then follow its instructions to download the latest database and build _libtrellis_.
+then follow its instructions to download the latest database and build _libtrellis_. 
+(for example: `-DTRELLIS_INSTALL_PREFIX=/usr` will install to `/usr/share/trellis`)
```
daveshah1: `TRELLIS_INSTALL_PREFIX` doesn't install to anywhere, it is the prefix where trellis is installed to. And trellis isn't just installed to `/usr/share` as it puts the compiled library in `/usr/lib` too

gojimmypi: ah. so the text in my example should be "will **look** here" ?

daveshah1: Please add 'and `/usr/lib/trellis`' at the end, then it is good

gojimmypi: done

PR #410, opened 2020-03-13
ecp5: Proper support for '12k' device

### LINKED ISSUE #409: nextpnr-ecp5+ecppack generate invalid SVF for LFE5U-12F
`nextpnr-ecp5` doesn't know the difference between a 25F and a 12F. For most purposes, this is fine, however, the two chips do have different JTAG ID codes. Although this is mostly handled by whatever programming tool one uses, the IDCODE appears to exist inside the SVF generated by `ecppack` as well, so to get a working bitstream for a 12F one must use the `--idcode` flag of `ecppack` and specify the IDCODE directly. This is not ideal - it would be nice if nextpnr did the right thing here.

daveshah1:
Also need SymbiFlow/prjtrellis#127 for bitstream generation.

Fixes #409 

@awygle please can you test this and report back, as I don't actually have any '12k' hardware?

### DIFF
--- a/ecp5/arch.cc
+++ b/ecp5/arch.cc
@@ -106,7 +106,8 @@ Arch::Arch(ArchArgs args) : args(args)
         log_error("Unsupported ECP5 chip type.\n");
     }
 #else
-    if (args.type == ArchArgs::LFE5U_25F || args.type == ArchArgs::LFE5UM_25F || args.type == ArchArgs::LFE5UM5G_25F) {
+    if (args.type == ArchArgs::LFE5U_12F || args.type == ArchArgs::LFE5U_25F || args.type == ArchArgs::LFE5UM_25F ||
+        args.type == ArchArgs::LFE5UM5G_25F) {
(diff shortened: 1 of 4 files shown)


mmicko:
Tested this on ULX3S with 12K and all works fine

daveshah1:
Thanks!

SiliconWizard:
Still seems to be an issue. I haven't checked the output .bit file to know if it is correct or not, but using --12k (on the latest nextpnr-ecp5) acts like with --25k; in particular it considers the --12k to have the same number of slices as the --25k.

So this isn't fixed as far as I can tell.

gatecat:
Think of it as a free bonus :)

SiliconWizard:
Does that mean that the 12F is actually the same device as the 25F, except for the IDCODE, which would seem to be confirmed by this file from Prjtrellis?

https://github.com/SymbiFlow/prjtrellis/blob/master/devices.json

If so, that is good to know. And shame for Lattice. ;) The 12F in the same package and speed grade is about 30% cheaper than the 25F... :D

gatecat:
Correct!

PR #415, opened 2020-03-29
ice40: Fix output register timing analysis for registered output enable

smunaut:
Wrong bits were being tested. [5:4] is what's controlling the output
enable path.


daveshah1:
Thanks

PR #435, opened 2020-04-29
Alter MULT18X18D timing db based on register config

rschlaikjer:
If the REG_INPUTA_CLK and REG_INPUTB_CLK values are set, then we should
use the faster setup/hold timings for the 18x8 multiplier.
Similarly, check the value of REG_OUTPUT_CLK for whether or not to use
faster timings for the output.

This is based on how I currently understand the registers to work - if
anyone knows the actual rules for when each timing applies please do
chime in to correct this implementation if necessary.

Along the same lines, this PR does not address the case when the
pipeline registers are enabled, since it is not clear to me how exactly
that affects the timing.

### DIFF
--- a/ecp5/arch.cc
+++ b/ecp5/arch.cc
@@ -940,12 +940,27 @@ TimingPortClass Arch::getPortTimingClass(const CellInfo *cell, IdString port, in
             return TMG_CLOCK_INPUT;
         std::string pname = port.str(this);
         if (pname.size() > 1) {
-            if (pname.front() == 'A' && std::isdigit(pname.at(1)))
-                return cell->multInfo.is_in_a_registered ? TMG_REGISTER_INPUT : TMG_COMB_INPUT;
-            if (pname.front() == 'B' && std::isdigit(pname.at(1)))
-                return cell->multInfo.is_in_b_registered ? TMG_REGISTER_INPUT : TMG_COMB_INPUT;
-            if (pname.front() == 'P' && std::isdigit(pname.at(1)))
-                return cell->multInfo.is_output_registered ? TMG_REGISTER_OUTPUT : TMG_COMB_OUTPUT;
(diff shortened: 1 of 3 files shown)


ecp5/arch.cc
```diff
@@ -855,7 +855,7 @@ bool Arch::getCellDelay(const CellInfo *cell, IdString fromPort, IdString toPort
         std::string fn = fromPort.str(this), tn = toPort.str(this);
         if (fn.size() > 1 && (fn.front() == 'A' || fn.front() == 'B') && std::isdigit(fn.at(1))) {
             if (tn.size() > 1 && tn.front() == 'P' && std::isdigit(tn.at(1)))
-                return getDelayFromTimingDatabase(id_MULT18X18D_REGS_NONE, id(std::string("") + fn.front()), id_P,
+                return getDelayFromTimingDatabase(cell->multInfo.timing_id, id(std::string("") + fn.front()), id_P,
```
daveshah1, line 858: This won't function correctly, clock-output and setup/hold times are accessed via getPortClockingInfo, not getCellDelay. For the latter, you will also need to query the database using getSetupHoldFromTimingDatabase. You will also need to update getPortTimingClass to report the DSP inputs/outputs as registered when the registers are enabled.

rschlaikjer: Is it safe to still leave this as the same timing DB used elsewhere? Or is there a reason that it must be the REGS_NONE version?

daveshah1: If a port is registered, then in fact getCellDelay should be returning false as it is only for combinational paths.

rschlaikjer: Updated to return false early if the mult block is clocked

daveshah1:
Thanks, merged with some small fixes in 84327b634c09b67ef965e3056dd29a7ee4169039

PR #437, opened 2020-05-11
ecp5: Allow setting drive strength for LVCMOS33D IOs

miek:


ecp5/bitstream.cc
```diff
@@ -915,6 +915,16 @@ void write_bitstream(Context *ctx, std::string base_config_file, std::string tex
...
+                } else if (iotype == "LVCMOS33D") {
+                    std::string other;
+                    if (pio == "PIOA")
+                        other = "PIOB";
+                    else if (pio == "PIOC")
+                        other = "PIOD";
+                    else
+                        log_error("cannot set DRIVE on differential IO at location %s\n", pio.c_str());
+                    cc.tiles[pio_tile].add_enum(pio + ".DRIVE", str_or_default(ci->attrs, ctx->id("DRIVE"), "12"));
+                    cc.tiles[pio_tile].add_enum(other + ".DRIVE", str_or_default(ci->attrs, ctx->id("DRIVE"), "12"));
```
daveshah1, line 927: For top/bottom IO tiles this will be in the "complement tile" which you will need to get using `get_comp_pio_tile(ctx, bel);`

miek: Ah, thanks. I see now how it should work for the top IO tiles, but I'm not sure about the bottom tiles. It looks like `get_comp_pio_tile` is for top tiles only. Is LVCMOS33D actually supported on bottom tiles? I can't see options for it in the tilegrid.

daveshah1: Oh no, I don't think it is. Adding support for top (and left/right as at the moment) tiles only is fine then.

ecp5/bitstream.cc
```diff
@@ -915,6 +915,24 @@ void write_bitstream(Context *ctx, std::string base_config_file, std::string tex
                 static bool drive_3v3_warning_done = false;
                 if (iotype == "LVCMOS33") {
                     cc.tiles[pio_tile].add_enum(pio + ".DRIVE", str_or_default(ci->attrs, ctx->id("DRIVE"), "8"));
+                } else if (iotype == "LVCMOS33D") {
+                    if (bel.location.y == 0) {
+                        // Pseudo differential top IO
+                        NPNR_ASSERT(dir == "OUTPUT");
+                        NPNR_ASSERT(pio == "PIOA");
+                        std::string cpio_tile = get_comp_pio_tile(ctx, bel);
+                        cc.tiles[cpio_tile].add_enum("PIOB.DRIVE", str_or_default(ci->attrs, ctx->id("DRIVE"), "12"));
```
daveshah1, line 924: You need to set DRIVE here in both the PIOA and PIOB tiles

miek: Doh, of course. Thanks

PR #442, opened 2020-05-14
Fix spelling of 'unsupported'

nategraff-sifive:
Hello nextpnr developers!

I've been working lately to package nextpnr for Debian, and one of the things that popped up during that process is that "unsupported" happens to be misspelled in a few places.

Thanks, and happy hacking!
Nate Graff


daveshah1:
Good catch! Thanks

Issue #215, opened 2019-01-18
Boost problem? CheckSymbolExists.c:(.text+0x16): undefined reference to `pthread_create'


gojimmypi:
I've been trying to get nextpnr to compile and have encountered a variety of issues as noted in https://github.com/YosysHQ/nextpnr/issues/214 and https://github.com/YosysHQ/nextpnr/issues/129#issuecomment-455707572. I believe my latest error of `No version of Boost::Python 3.x could be found` is misleading and inaccurate. Is this really a `boost` problem?

Given the whole issue with cmake cache that I found in issue #214 and all, I removed all instances of python3.x and all respective libraries, then reinstalled fresh like this:

```
sudo apt-get install python3.5
sudo apt-get install qt5-default python3.5-dev
sudo apt-get install libboost-all-dev
```
Next I cloned nextpnr into a completely new directory in order to run `cmake -DARCH=ice40 .`  

The fresh output is:

```
0 $ cmake -DARCH=ice40 .
-- The C compiler identification is GNU 4.8.4
-- The CXX compiler identification is Clang 3.4.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found PythonInterp: /usr/bin/python3.5 (found suitable version "3.5.2", minimum required is "3.5")
-- Found PythonLibs: /usr/lib/x86_64-linux-gnu/libpython3.5m.so (found suitable version "3.5.2", minimum required is "3.5")
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Looking for pthread_create
-- Looking for pthread_create - not found
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - found
-- Found Threads: TRUE
-- Boost version: 1.54.0
-- Found the following Boost libraries:
--   filesystem
--   thread
--   program_options
--   system
--   chrono
--   date_time
--   atomic
[...]

PR #443, opened 2020-05-18
Update README with workaround for "cannot find -lpthreads"

gojimmypi:
I encountered this error again on WSL Ubuntu 20.04 and despite seemingly having all dependencies installed, `find /lib -name "*pthread*" ` does not return any results. (This is not a problem on my 
Ubuntu WSL 18.04).

However, the [workaround for #215](https://github.com/YosysHQ/nextpnr/issues/215#issuecomment-456053526) seemed to work. 

I've updated the README with info on the workaround.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,23 @@ sudo make install
 
  - Examples of the ECP5 flow for a range of boards can be found in the [Project Trellis Examples](https://github.com/SymbiFlow/prjtrellis/tree/master/examples).
 
+If you encounter an error `/usr/bin/ld: cannot find -lpthreads` and you are sure all libaries are installed; see also `CMakeError.log`:
+```
+/usr/bin/ld: CMakeFiles/cmTC_6be3a.dir/src.c.o: in function `main':
+src.c:(.text+0x46): undefined reference to `pthread_create'
+/usr/bin/ld: src.c:(.text+0x52): undefined reference to `pthread_detach'
+/usr/bin/ld: src.c:(.text+0x63): undefined reference to `pthread_join'
+collect2: error: ld returned 1 exit status
+make[1]: *** [CMakeFiles/cmTC_6be3a.dir/build.make:87: cmTC_6be3a] Error 1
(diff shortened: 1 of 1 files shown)


whitequark:
Is there an open bug for this problem on WSL 20.04 so that it can be tracked and hopefully fixed? Stuffing a workaround into README feels like sweeping problems under the carpet.

daveshah1:
What is the actual problem anyway? I think the pthread error in that other issue might have been a red herring and the actual issue was boost related.

gojimmypi:
@whitequark I didn't open a new issue... I thought perhaps it is just WSL related and a workaround exists. Would you like me to open a new issue?

@daveshah1 I am seeing a new `cannot find -lpthreads` error at compile time in a fresh Ubuntu 20.04 WSL that I do not see in 18.04, despite having  (what I believe to be)  all the dependencies installed.

I'm working on a ["bare bones" install](https://github.com/gojimmypi/ulx3s-toolchain/blob/master/install_barebones.sh) for the FPGA build/synthesis from within Visual Studio. The specific install script I have for nextpnr is [here](https://github.com/gojimmypi/ulx3s-toolchain/blob/master/install_nextpnr.sh).

whitequark:
> I thought perhaps it is just WSL related and a workaround exists. Would you like me to open a new issue?

A workaround helps people who were blocked by the issue enough to go over README again. A fix in the build system helps everyone. Shouldn't we prefer the latter?

gojimmypi:
Yes, I certainly agree it is good to have a robust build system to help everyone. Still, some part of me still would like to have the knowledge available for the workaround. 

I've created https://github.com/YosysHQ/nextpnr/issues/444 with the details.

whitequark:
AFAICT the error in #444 has nothing to do with nextpnr, at least not directly; your system is running out of memory and the compiler is handling that badly.

gojimmypi:
@whitequark you are quite correct. [Low memory](https://github.com/YosysHQ/nextpnr/issues/444#issuecomment-630886150) was the root cause.

> Stuffing a workaround into README feels like sweeping problems under the carpet

I completely agree. I'll come up with something to instead warn about low memory.

PR #452, opened 2020-06-02
ice40: Add support for the 2nd bit of SHIFTREG_DIV_MODE

smunaut:
This requires the matching chipdb update from icestorm project !

### DIFF
--- a/ice40/bitstream.cc
+++ b/ice40/bitstream.cc
@@ -718,7 +718,7 @@ void write_asc(const Context *ctx, std::ostream &out)
                                                                          {"PLLOUT_SELECT_A", 2},
                                                                          {"PLLOUT_SELECT_B", 2},
                                                                          {"PLLTYPE", 3},
-                                                                         {"SHIFTREG_DIV_MODE", 1},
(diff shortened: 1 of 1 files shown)


ice40/bitstream.cc
```diff
@@ -190,7 +200,7 @@ void configure_extra_cell(chipconfig_t &config, const Context *ctx, CellInfo *ce
         }
 
         value.resize(p.second);
-        if (p.second == 1) {
+        if ((p.second == 1) || !has_ec_cbit(bc, p.first + "_0")) {
```
daveshah1, line 203: I think this also needs a check that bit 1 isn't set, or it would be a silent miscompilation of divide-by-5 using an older database.

smunaut: Well building with an older nextpnr already leads to a miscompilation right now.

smunaut: (although miscompilation for something that's unsupported and doesn't work under radiant or icecube might be a bit of a stretch)

daveshah1: Ah yeah, in that case it doesn't really matter, I'm happy to merge once the CI is green.

PR #456, opened 2020-06-23
ecp5: Split the SLICE bel into smaller components

daveshah1:
This PR takes a similar approach to the architecture of nextpnr-xilinx, relying on fast legality checks and the performance of remove the need for a significant pre-placement packing step.

It splits the current SLICE bel into:
 - 2x TRELLIS_COMB containing LUT4, 1-bit carry, MUX2
 - 2x TRELLIS_FF
 - slice 2 also adds a TRELLIS_RAMW for the LUTRAM write port

This brings some significant benefits:
 - Cells are better preserved from synthesis, good for future incremental P&R work
 - Retiming will be much easier with LUTs and FFs separated
 - necessary for the incremental pack-and-place in my current Ripple placer work
 - possible general QoR improvements by giving the placer more choice rather than forcing things together

At the moment this is a WIP. It needs the [split_slice](https://github.com/SymbiFlow/prjtrellis/tree/split_slice) branch of prjtrellis.


daveshah1:
A small progress update on this, all functionality is now working with split SLICEs. On small designs there seems to be some QoR increase due to denser packing, but this decreases QoR and increases on bigger designs due to higher congestion (despite substantially reduced total HPWL). All this is further evidence that a routeability-driven placer than can move cells out of high-congestion areas is needed, fortunately that is my other workstream.

gatecat:
Closing, this may be revisited some time in the distant future when prjtrellis and prjoxide are harmonised.

PR #457, opened 2020-06-23
CMake: promote bba to a true subproject

whitequark:
The direct motivation for this change is better cross-builds: since the only thing I actually *want* from the native build is bba (and everything else just gets in the way), bba can and should be its own subproject, which can be independently configured and built with CMake. Non-cross builds are not affected.

I also took the liberty to clean up the bba CMake files, since they were using some really ancient style.


whitequark:
cc @daveshah1 @mmicko

whitequark:
Actually, I think I didn't get the scoping quite right, I need to amend this.

mmicko:
LGTM as well

whitequark:
PTAL, I changed the approach slightly. It is now cleaner and also actually works for cross-builds.

PR #459, opened 2020-06-24
CMake: rewrite chipdb handling from ground up

whitequark:
This addresses a lot of different issues:
  * The build system logic that handles the `.bba` files is massively simplified.
  * `.bba` files can now be built separately from the rest of nextpnr, simplifying pre-generation and cross-compilation.
  * `.bba` files can now be included into the binary using any desired approach (selected by the internal variable `BBASM_MODE`), independent from whether they are built separately or from the selected compiler. The default behavior is unchanged, with the exception that non-MSVC Win32 binaries also use the more efficient approach of using resource files.
  * The generation of `.bba` files is now serialized across a multi-architecture build, avoiding OOMs caused by building iCE40 and ECP5 chipdbs in parallel.
  * Both iCE40 and ECP5 can now be subsetted to use only some of the devices. This isn't yet handled by the nextpnr code (other than the existing HX1K case), which is left to a future PR.
  * Since both ECP5 and iCE40 require IceStorm/Trellis to be installed, their configuration options are unified; Trellis uses `-DTRELLIS_INSTALL_PREFIX=`, and IceStorm now uses `-DICESTORM_INSTALL_PREFIX=`. The old `-DICEBOX_ROOT` option is retained for compatibility.
  * Better documentation for cross-compilation and chipdb pre-generation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ of the selected architecture:
 - For building on macOS, brew utility is needed.
   - Install all needed packages `brew install cmake python boost boost-python3 qt5 eigen`
   - Do not forget to add qt5 in path as well `echo 'export PATH="/usr/local/opt/qt/bin:$PATH"' >> ~/.bash_profile`
-  
+
     NOTE: this change is effective in next terminal session, so please re-open terminal window before next step
 
 Getting started
@@ -54,13 +54,10 @@ Getting started
 
 ### nextpnr-ice40
 
-To build the iCE40 version of nextpnr, install [icestorm](http://www.clifford.at/icestorm/) with chipdbs installed in `/usr/local/share/icebox`,
-or another location, which should be passed as `-DICEBOX_ROOT=/path/to/share/icebox` (ensure to point it to `share/icebox` and not where the
-icebox binaries are installed) to CMake.
-Then build and install `nextpnr-ice40` using the following commands:
+For iCE40 support, install [Project IceStorm](http://www.clifford.at/icestorm/) to `/usr/local` or another location, which should be passed as `-DICESTORM_INSTALL_PREFIX=/usr` to CMake. Then build and install `nextpnr-ice40` using the following commands:
 
 ```
-cmake -DARCH=ice40 .
+cmake . -DARCH=ice40
 make -j$(nproc)
 sudo make install
 ```
@@ -68,7 +65,7 @@ sudo make install
 On Windows, you may specify paths explicitly:
 
 ```
-cmake -DARCH=ice40 -DICEBOX_ROOT=C:/ProgramData/icestorm/share/icebox -DCMAKE_TOOLCHAIN_FILE=C:/vcpkg/scripts/buildsystems/vcpkg.cmake -DVCPKG_TARGET_TRIPLET=x64-windows -G "Visual Studio 15 2017 Win64" -DPYTHON_EXECUTABLE=C:/Python364/python.exe -DPYTHON_LIBRARY=C:/vcpkg/packages/python3_x64-windows/lib/python36.lib -DPYTHON_INCLUDE_DIR=C:/vcpkg/packages/python3_x64-windows/include/python3.6
+cmake . -DARCH=ice40 -DICEBOX_INSTALL_PREFIX=C:/ProgramData/icestorm -DCMAKE_TOOLCHAIN_FILE=C:/vcpkg/scripts/buildsystems/vcpkg.cmake -DVCPKG_TARGET_TRIPLET=x64-windows -G "Visual Studio 15 2017 Win64" -DPYTHON_EXECUTABLE=C:/Python364/python.exe -DPYTHON_LIBRARY=C:/vcpkg/packages/python3_x64-windows/lib/python36.lib -DPYTHON_INCLUDE_DIR=C:/vcpkg/packages/python3_x64-windows/include/python3.6 .
 cmake --build . --config Release
 ```
 
@@ -96,13 +93,10 @@ such as pack, place, route, and write output files.)
 
 ### nextpnr-ecp5
 
-For ECP5 support, you must download [Project Trellis](https://github.com/SymbiFlow/prjtrellis),
-then follow its instructions to download the latest database and build _libtrellis_. 
-(for example: `-DTRELLIS_INSTALL_PREFIX=/usr` tells nextpnr to look in `/usr/share/trellis` 
-and `/usr/lib/trellis`) 
+For ECP5 support, install [Project Trellis](https://github.com/SymbiFlow/prjtrellis) to `/usr/local` or another location, which should be passed as `-DTRELLIS_INSTALL_PREFIX=/usr/local` to CMake. Then build and install `nextpnr-ecp5` using the following commands:
 
 ```
-cmake -DARCH=ecp5 -DTRELLIS_INSTALL_PREFIX=/path/to/prjtrellis .
+cmake . -DARCH=ecp5 -DTRELLIS_INSTALL_PREFIX=/usr
 make -j$(nproc)
 sudo make install
 ```
@@ -115,45 +109,71 @@ sudo make install
 The generic target allows running placement and routing for arbitrary custom architectures.
 
 ```
-cmake -DARCH=generic .
+cmake . -DARCH=generic
 make -j$(nproc)
 sudo make install
 ```
 
 An example of how to use the generic flow is in [generic/examples](generic/examples). See also the [Generic Architecture docs](docs/generic.md).
 
+### Multiple architectures
+
+To build nextpnr for multiple architectures at once, a semicolon-separated list can be used with `-DARCH`.
+
+```
+cmake . -DARCH="ice40;ecp5"
+make -j$(nproc)
+sudo make install
+```
+
+To build every available architecture, use `-DARCH=all`.
+
+Pre-generating chip databases
+-----------------------------
+
+It is possible to pre-generate chip databases (`.bba` files). This can come in handy when building on time-constrained cloud instances, or in situations where Python is unable to use modules. To do this, build the architecture as a standalone project, which will produce the chip database alone. For example, for iCE40:
+
+```
+cd ice40
+cmake .
+make
+```
+
+This will create a `chipdb` directory with `.bba` files. Provide the path to this directory when building nextpnr by using `-D<arch>_CHIPDB=/path/to/chipdb`.
+
+Cross-compilation
+-----------------
+
+Apart from chip databases, nextpnr requires the `bba` tool to be compiled for the build system. This tool can be compiled as a separate project:
+
+```
+cd bba
+cmake .
+make
+```
+
+This will create a `bba-export.cmake` file. Provide the path to this file when cross-building nextpnr by using `-DBBA_IMPORT=/path/to/bba-export.cmake`.
+
 Additional notes for building nextpnr
 -------------------------------------
 
-Use cmake `-D` options to specify which version of nextpnr you want to build.
-
-Use `-DARCH=...` to set the architecture. It is a semicolon separated list.
-Use `cmake . -DARCH=all` to build all supported architectures.
-
-The following runs a debug build of the iCE40 architecture without GUI,
- without Python support, without the HeAP analytic placer and only HX1K support:
+The following runs a debug build of the iCE40 architecture without GUI, without Python support, without the HeAP analytic placer and only HX1K support:
 
 ```
-cmake -DARCH=ice40 -DCMAKE_BUILD_TYPE=Debug -DBUILD_PYTHON=OFF -DBUILD_GUI=OFF -DBUILD_HEAP=OFF -DICE40_HX1K_ONLY=1 .
+cmake . -DARCH=ice40 -DCMAKE_BUILD_TYPE=Debug -DBUILD_PYTHON=OFF -DBUILD_GUI=OFF -DBUILD_HEAP=OFF -DICE40_HX1K_ONLY=1
 make -j$(nproc)
 ```
 
-To make static build relase for iCE40 architecture use the following:
+To make static build release for iCE40 architecture use the following:
 
 ```
-cmake -DARCH=ice40 -DBUILD_PYTHON=OFF -DBUILD_GUI=OFF -DSTATIC_BUILD=ON .
+cmake . -DARCH=ice40 -DBUILD_PYTHON=OFF -DBUILD_GUI=OFF -DSTATIC_BUILD=ON
 make -j$(nproc)
 ```
 
-The HeAP placer's solver can optionally use OpenMP for a speedup on very large designs. Enable this by passing
-`-DUSE_OPENMP=yes` to cmake (compiler support may vary).
+The HeAP placer's solver can optionally use OpenMP for a speedup on very large designs. Enable this by passing `-DUSE_OPENMP=yes` to cmake (compiler support may vary).
 
-You can change the location where nextpnr will be installed (this will usually default to `/usr/local`) by using
-`-DCMAKE_INSTALL_PREFIX=/install/prefix`.
-
-It is possible to pre-generate `.bba` files.  This can come in handy when building on time-constrained cloud
-instances, or in situations where python is unable to use modules.  To do this, specify the path to pre-
-generated `.bba` files by passing `-DPREGENERATED_BBA_PATH=` to cmake.
+You can change the location where nextpnr will be installed (this will usually default to `/usr/local`) by using `-DCMAKE_INSTALL_PREFIX=/install/prefix`.
 
 Notes for developers
 --------------------
@@ -182,7 +202,7 @@ Testing
   - `-DSANITIZE_THREAD=ON`
   - `-DSANITIZE_UNDEFINED=ON`
 - Running valgrind example `valgrind --leak-check=yes --tool=memcheck ./nextpnr-ice40 --json ice40/blinky.json`
-- Running tests with code coverage use `-DBUILD_TESTS=ON -DCOVERAGE` and after `make` run `make ice40-coverage` 
+- Running tests with code coverage use `-DBUILD_TESTS=ON -DCOVERAGE` and after `make` run `make ice40-coverage`
 - After that open `ice40-coverage/index.html` in your browser to view the coverage report
 - Note that `lcov` is needed in order to generate reports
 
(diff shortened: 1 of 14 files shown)


daveshah1:
Thank you very much for doing this! I will give it a proper review tomorrow.

whitequark:
I've tested this in the following configurations:
  * x86_64 Linux with system IceStorm/Trellis
  * x86_64 Linux with IceStorm with explicit prefix
  * x86_64 Linux with prebuilt IceStorm bbas and system Trellis
  * x86_64 Linux with only ice40 1k device (GUI does not build but I think I didn't introduce that)
  * x64 MinGW on x86_64 Linux (Boost craps out so I had to do some manual workarounds, however my changes seem to be OK)
  * VS2015 with IceStorm with explicit prefix
  * WASM with IceStorm with explicit prefix

cr1901:
@whitequark 

Can confirm a working `nextpnr-ice40` on my end with your changes that routes a blinky design successfully. I used the following command line:

```
cmake -G Ninja -DARCH=ice40 -DCMAKE_INSTALL_PREFIX=/mingw64 -DICESTORM_INSTALL_PREFIX=/mingw64 ..
```

For `nextpnr-ecp5`, I was able to build successfully _without_ needing to set `TRELLIS_LIBDIR` with the following patch:

```diff
diff --git a/ecp5/CMakeLists.txt b/ecp5/CMakeLists.txt
index 696f379..8939cd8 100644
--- a/ecp5/CMakeLists.txt
+++ b/ecp5/CMakeLists.txt
@@ -26,10 +26,17 @@ else()
     message(STATUS "Trellis install prefix: ${TRELLIS_INSTALL_PREFIX}")
 
     if(NOT DEFINED TRELLIS_LIBDIR)
-        find_library(PYTRELLIS pytrellis${CMAKE_SHARED_LIBRARY_SUFFIX}
+        if(WIN32)
+            set(PREVIOUS_CMAKE_FIND_LIBRARY_SUFFIXES ${CMAKE_FIND_LIBRARY_SUFFIXES})
+            set(CMAKE_FIND_LIBRARY_SUFFIXES "${CMAKE_FIND_LIBRARY_SUFFIXES}.pyd")
+        endif()
+        find_library(PYTRELLIS pytrellis
             PATHS ${TRELLIS_INSTALL_PREFIX}/lib/${TRELLIS_PROGRAM_PREFIX}trellis
             PATH_SUFFIXES ${TRELLIS_PROGRAM_PREFIX}trellis
             DOC "Location of the pytrellis library")
+        if(WIN32)
+            set(CMAKE_FIND_LIBRARY_SUFFIXES ${PREVIOUS_CMAKE_FIND_LIBRARY_SUFFIXES})
+        endif()
         if(NOT PYTRELLIS)
             message(FATAL_ERROR "Failed to locate the pytrellis library")
         endif()
```

`.pyd` _is_ in fact the [correct](https://docs.python.org/3/faq/windows.html#is-a-pyd-file-the-same-as-a-dll) suffix on Windows to use for Python DLLs, so I manually force CMake to consider them while looking for `pytrellis`. The default suffix list is something like `.dll.dll.a.a.lib` (don't ask me why some are included twice).

With that patch and the below command line, I have a working `nextpnr-ecp5` GUI; I don't have any designs to test on atm, so the GUI will have to suffice as proof:

```
cmake --debug -G Ninja -DARCH=ecp5 -DCMAKE_INSTALL_PREFIX=/mingw64 -DTRELLIS_INSTALL_PREFIX=/mingw64 ..
```

image

whitequark:
> ```
> cmake -G Ninja -DARCH=ice40 -DCMAKE_INSTALL_PREFIX=/mingw64 -DICESTORM_INSTALL_PREFIX=/mingw64 ..
> ```

If you set `CMAKE_INSTALL_PREFIX`, that should also set the `ICESTORM_INSTALL_PREFIX`, so in theory it shouldn't be necessary to enter it explicitly.

whitequark:
> `.pyd` _is_ in fact the [correct](https://docs.python.org/3/faq/windows.html#is-a-pyd-file-the-same-as-a-dll) suffix on Windows to use for Python DLLs, so I manually force CMake to consider them while looking for `pytrellis`.

What if you do `find_library(PYTRELLIS pytrellis.pyd ...)` instead of that hack?

gsomlo:
On Fedora, the icestorm `chipdb` and `timings` files go in `/usr/share/icestorm/`, so up until now I've been building with:

```
%cmake . -DARCH=all \
        -DICEBOX_ROOT=%{_datadir}/icestorm \
        -DTRELLIS_ROOT=%{_datadir}/trellis
%make_build
```

With this patch set, I've had to additionally specify `ICEBOX_DATADIR`:

```
%cmake . -DARCH=all \
        -DICEBOX_ROOT=%{_datadir}/icestorm \
        -DICEBOX_DATADIR=%{_datadir}/icestorm \
        -DTRELLIS_ROOT=%{_datadir}/trellis
%make_build
```

For now, that seems to make it build without complaints. I'm going to try leaving out `ICEBOX_ROOT` entirely tomorrow, and report back if that gets me into any kind of trouble.

cr1901:
> > `.pyd` _is_ in fact the [correct](https://docs.python.org/3/faq/windows.html#is-a-pyd-file-the-same-as-a-dll) suffix on Windows to use for Python DLLs, so I manually force CMake to consider them while looking for `pytrellis`.
> 
> What if you do `find_library(PYTRELLIS pytrellis.pyd ...)` instead of that hack?

Forcing the suffix does not override the behavior of `CMAKE_FIND_LIBRARY_SUFFIXES`.

whitequark:
Okay, so that patch actually breaks *nix build. Should we even use find_library here at all?

cr1901:
If you're not manually specifying the paths, confirming that the Python module actually exists seems reasonable.

I think I misunderstood the [docs](https://cmake.org/cmake/help/latest/command/find_library.html):

>Each library name given to the NAMES option is first considered as a library file name and then considered with platform-specific prefixes (e.g. lib) and suffixes (e.g. .so). Therefore one may specify library file names such as libfoo.a directly. This can be used to locate static libraries on UNIX-like systems.

I.e. the *nix code is looking for `libpytrellis.so` and then `pytrellis` since I didn't specify a suffix. So combine my above patch with your `pytrellis${CMAKE_SHARED_LIBRARY_SUFFIX}`, replacing `${CMAKE_SHARED_LIBRARY_SUFFIX}` with a local variable, like `PYTRELLIS_LIBRARY_PREFIX`?

whitequark:
Nah, that's awful. Does this work for you?
```
diff --git a/ecp5/CMakeLists.txt b/ecp5/CMakeLists.txt
index 696f3791..d6363ecd 100644
--- a/ecp5/CMakeLists.txt
+++ b/ecp5/CMakeLists.txt
@@ -26,8 +26,14 @@ else()
     message(STATUS "Trellis install prefix: ${TRELLIS_INSTALL_PREFIX}")
 
     if(NOT DEFINED TRELLIS_LIBDIR)
-        find_library(PYTRELLIS pytrellis${CMAKE_SHARED_LIBRARY_SUFFIX}
-            PATHS ${TRELLIS_INSTALL_PREFIX}/lib/${TRELLIS_PROGRAM_PREFIX}trellis
+        if(WIN32)
+            set(pytrellis_lib pytrellis.pyd)
+        else()
+            set(pytrellis_lib pytrellis${CMAKE_SHARED_LIBRARY_SUFFIX})
+        endif()
+        find_path(PYTRELLIS ${pytrellis_lib}
+            PATHS ${CMAKE_SYSTEM_LIBRARY_PATH} ${CMAKE_LIBRARY_PATH}
+            HINTS ${TRELLIS_INSTALL_PREFIX}/lib/${TRELLIS_PROGRAM_PREFIX}trellis
             PATH_SUFFIXES ${TRELLIS_PROGRAM_PREFIX}trellis
             DOC "Location of the pytrellis library")
         if(NOT PYTRELLIS)
```

cr1901:
That works on my end.

whitequark:
Works for me too, and is actually the appropriate way to fix this problem, I believe. Thanks for testing!

whitequark:
> ^This is wrong; the equivalent patch to yours that I made had a bug in it. With your patch, minus my bug, configuration works successfully.

Right, the problem is that it's basically undoing most of the logic that comprises the purpose of `find_library`, so we should just not use `find_library` in that case.

whitequark:
Per IRC discussion, can you check if the following works, applied on top of the updated branch?
```diff
diff --git a/ecp5/CMakeLists.txt b/ecp5/CMakeLists.txt
index e59d33ea..379ac12c 100644
--- a/ecp5/CMakeLists.txt
+++ b/ecp5/CMakeLists.txt
@@ -31,7 +31,7 @@ else()
         else()
             set(pytrellis_lib pytrellis${CMAKE_SHARED_LIBRARY_SUFFIX})
         endif()
-        find_path(PYTRELLIS ${pytrellis_lib}
+        find_library(PYTRELLIS ${pytrellis_lib}
             HINTS ${TRELLIS_INSTALL_PREFIX}/lib/${TRELLIS_PROGRAM_PREFIX}trellis
             PATHS ${CMAKE_SYSTEM_LIBRARY_PATH} ${CMAKE_LIBRARY_PATH}
             PATH_SUFFIXES ${TRELLIS_PROGRAM_PREFIX}trellis
```

cr1901:
> Per IRC discussion, can you check if the following works, applied on top of the updated branch?
> 
> ```diff
> diff --git a/ecp5/CMakeLists.txt b/ecp5/CMakeLists.txt
> index e59d33ea..379ac12c 100644
> --- a/ecp5/CMakeLists.txt
> +++ b/ecp5/CMakeLists.txt
> @@ -31,7 +31,7 @@ else()
>          else()
>              set(pytrellis_lib pytrellis${CMAKE_SHARED_LIBRARY_SUFFIX})
>          endif()
> -        find_path(PYTRELLIS ${pytrellis_lib}
> +        find_library(PYTRELLIS ${pytrellis_lib}
>              HINTS ${TRELLIS_INSTALL_PREFIX}/lib/${TRELLIS_PROGRAM_PREFIX}trellis
>              PATHS ${CMAKE_SYSTEM_LIBRARY_PATH} ${CMAKE_LIBRARY_PATH}
>              PATH_SUFFIXES ${TRELLIS_PROGRAM_PREFIX}trellis
> ```

No, the quoted patch applied on top of your [previous](https://github.com/YosysHQ/nextpnr/pull/459#issuecomment-649202034) patch does not work:

```
-- Found Boost: C:/msys64/mingw64/include (found version "1.72.0") found components:  program_options filesystem system
-- Check if the system is big endian
-- Searching 16 bit integer
-- Looking for C++ include sys/types.h
-- Looking for C++ include sys/types.h - found
-- Looking for C++ include stdint.h
-- Looking for C++ include stdint.h - found
-- Looking for C++ include stddef.h
-- Looking for C++ include stddef.h - found
-- Check size of unsigned short
-- Check size of unsigned short - done
-- Using unsigned short
-- Check if the system is big endian - little endian
-- Configuring architecture: ecp5
-- Enabled ECP5 devices: 25k;45k;85k
-- Trellis install prefix: C:/msys64/mingw64
CMake Error at ecp5/CMakeLists.txt:40 (message):
  Failed to locate the pytrellis library


-- Configuring incomplete, errors occurred!
See also "C:/msys64/home/William/Projects/FPGA/nextpnr/build-ecp5/CMakeFiles/CMakeOutput.log".

William@DESKTOP-H0PMN4M MINGW64 ~/Projects/FPGA/nextpnr/build-ecp5
$
```

gsomlo:
On Fedora, `pytrellis.so` is provided by the `trellis-devel` subpackage, and is located in `/usr/lib64trellis/pytrellis.so`. As such, with the current version of the patch set, I can build using:

```
%cmake . -DARCH=all \
        -DICEBOX_ROOT=%{_datadir}/icestorm \
        -DICEBOX_DATADIR=%{_datadir}/icestorm \
        -DTRELLIS_ROOT=%{_datadir}/trellis \
        -DTRELLIS_LIBDIR=%{_libdir}/trellis
%make_build
```

(note the additional explicit setting of `TRELLIS_LIBDIR`).

whitequark:
FYI there is no such thing as `TRELLIS_ROOT` at all, and there is no such thing as `ICEBOX_ROOT` anymore.

whitequark:
> As such, with the current version of the patch set,

Is that with commit 0645da40f214afba1c8a23866b97723b7e1831d1? It was supposed to take care of that issue, maybe it didn't...

README.md
```diff
@@ -46,29 +46,26 @@ of the selected architecture:
...
+cmake . -DARCH=ice40
 make -j$(nproc)
 sudo make install
 ```
 
 On Windows, you may specify paths explicitly:
 
 ```
-cmake -DARCH=ice40 -DICEBOX_ROOT=C:/ProgramData/icestorm/share/icebox -DCMAKE_TOOLCHAIN_FILE=C:/vcpkg/scripts/buildsystems/vcpkg.cmake -DVCPKG_TARGET_TRIPLET=x64-windows -G "Visual Studio 15 2017 Win64" -DPYTHON_EXECUTABLE=C:/Python364/python.exe -DPYTHON_LIBRARY=C:/vcpkg/packages/python3_x64-windows/lib/python36.lib -DPYTHON_INCLUDE_DIR=C:/vcpkg/packages/python3_x64-windows/include/python3.6
+cmake . -DARCH=ice40 -DICEBOX_ROOT=C:/ProgramData/icestorm/share/icebox -DCMAKE_TOOLCHAIN_FILE=C:/vcpkg/scripts/buildsystems/vcpkg.cmake -DVCPKG_TARGET_TRIPLET=x64-windows -G "Visual Studio 15 2017 Win64" -DPYTHON_EXECUTABLE=C:/Python364/python.exe -DPYTHON_LIBRARY=C:/vcpkg/packages/python3_x64-windows/lib/python36.lib -DPYTHON_INCLUDE_DIR=C:/vcpkg/packages/python3_x64-windows/include/python3.6 .
```
daveshah1, line 68: Is ICEBOX_ROOT deprecated for Windows too, and should this be updated?

README.md
```diff
@@ -96,13 +93,10 @@ such as pack, place, route, and write output files.)
 
 ### nextpnr-ecp5
 
-For ECP5 support, you must download [Project Trellis](https://github.com/SymbiFlow/prjtrellis),
-then follow its instructions to download the latest database and build _libtrellis_. 
-(for example: `-DTRELLIS_INSTALL_PREFIX=/usr` tells nextpnr to look in `/usr/share/trellis` 
-and `/usr/lib/trellis`) 
+For ECP5 support, install [Project Trellis](https://github.com/SymbiFlow/prjtrellis) to `/usr/local` or another location, which should be passed as `-DTRELLIS_INSTALL_PREFIX=/usr` to CMake. Then build and install `nextpnr-ecp5` using the following commands:
```
daveshah1, line 96: Mixing `/usr/local` and `/usr` seems a bit confusing here

gsomlo:
> > As such, with the current version of the patch set,
> 
> Is that with commit [0645da4](https://github.com/YosysHQ/nextpnr/commit/0645da40f214afba1c8a23866b97723b7e1831d1)? It was supposed to take care of that issue, maybe it didn't...

I just tried again (making sure I have the latest of everything), and I can apparently get away with just:

```
%cmake . -DARCH=all \
        -DICEBOX_DATADIR=%{_datadir}/icestorm \
        -DTRELLIS_LIBDIR=%{_libdir}/trellis
%make_build
```

ice40/CMakeLists.txt
```diff
@@ -0,0 +1,88 @@
...
+    find_package(PythonInterp 3.5 REQUIRED)
+
+    # shared among all families
+    set(SERIALIZE_CHIPDBS TRUE CACHE BOOL
+        "Serialize device data preprocessing to minimize memory use")
+
+    set(icestorm_default_install_prefix ${CMAKE_INSTALL_PREFIX})
+    # for compatibility with old build scripts
+    if(DEFINED ICEBOX_ROOT)
+        message(WARNING "-DICEBOX_ROOT= is deprecated, use -DICESTORM_INSTALL_PREFIX= instead")
```
daveshah1, line 21: Can you make this use `${dir}` at the end to print something like `use -DICESTORM_INSTALL_PREFIX=/usr/local`

ice40/CMakeLists.txt
```diff
@@ -0,0 +1,88 @@
...
+
+        set(device_bba chipdb/chipdb-${device}.bba)
+        add_custom_command(
+            OUTPUT ${device_bba}
+            COMMAND ${PYTHON_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/chipdb.py
+                -p ${CMAKE_CURRENT_SOURCE_DIR}/constids.inc
+                -g ${CMAKE_CURRENT_SOURCE_DIR}/gfx.h
+                ${timing_opts}
+                ${ICEBOX_DATADIR}/chipdb-${device}.txt
+                > ${device_bba}
```
daveshah1, line 65: it's a bit of a shame to lose the `mv` step that the old rules had, as if generation fails mid-way (e.g. OOM killed) then something will now fail more subtly on a half-written bba file rather than just re-running the generator script.

README.md
whitequark: I missed this, fixed

README.md
whitequark: Agree, fixed

ice40/CMakeLists.txt
whitequark: Good idea, done

ice40/CMakeLists.txt
whitequark: Ah I missed this, added back but without a dependency on Unix `mv`

whitequark:
All review comments addressed, I think

README.md
```diff
@@ -115,45 +109,69 @@ sudo make install
...
+To build nextpnr for multiple architectures at once, a semicolon-separated list can be used with `-DARCH`.
+
+```
+cmake . -DARCH="ice40;ecp5"
+make -j$(nproc)
+sudo make install
+```
 
-Use `-DARCH=...` to set the architecture. It is a semicolon separated list.
-Use `cmake . -DARCH=all` to build all supported architectures.
```
Xiretza, line 131: I think `-DARCH=all` is worth mentioning.

whitequark: Added in the section on multiple architectures

whitequark:
Thank you!

PR #460, opened 2020-06-25
Simplify and improve chipdb embedding/loading

whitequark:
This addresses multiple different issues:
  * It is now possible to build nextpnr with an arbitrary selection of chipdbs, instead of choosing between all of them, and using an undocumented option picking a single mid-size device.
  * It is now possible, if the external chipdb is used, to have nextpnr dynamically determine the availability of chipdbs at startup. This is critical for the ability of [YoWASP](https://yowasp.org/) to ship a binary plus a separate collection of chipdbs, since PyPI upload size limits prevent distributing a full build of nextpnr.
  * It is now possible to use external chipdb on Win32.
  * The code is a lot simpler and more robust.


whitequark:
I've tested this on:
  * Linux with GUI (internal chipdb)
  * Linux with GUI (external chipdb)
  * Windows MSVC with no GUI
  * WASM

common/embed.h
```diff
@@ -0,0 +1,43 @@
...
+    }
+};
+
+#endif
+
+const void *get_chipdb(const std::string &filename);
+
+NEXTPNR_NAMESPACE_END
+
+#endif // ROUTER1_H
```
daveshah1, line 43: nitpick:  `// ROUTER1_H` isn't right here

common/embed.h
```diff
@@ -0,0 +1,43 @@
...
+#include "nextpnr.h"
+NEXTPNR_NAMESPACE_BEGIN
+
+#if !defined(EXTERNAL_CHIPDB_ROOT) && !defined(WIN32)
+
+struct EmbeddedFile
+{
+    static std::map<std::string, const void *> all;
+
+    EmbeddedFile(const std::string &filename, const void *content) {
```
daveshah1, line 32: Because C++ is weird, it is guaranteed that `all` will be initialised before the EmbeddedFile constructor runs, right?`

common/embed.cc
```diff
@@ -0,0 +1,45 @@
...
+    }
+    if (files.count(filename))
+        return files.at(filename).data();
+    return nullptr;
+}
+
+#elif defined(WIN32)
+
+const void *get_chipdb(const std::string &filename) {
+    HRSRC rc = ::FindResource(NULL, filename.c_str(), RT_RCDATA);
```
daveshah1, line 28: I think the old code was using NULL, but can you change that to nullptr for consistency with other code in ths file?

common/embed.h
whitequark: Fixed

common/embed.cc
whitequark: Fixed

common/embed.h
whitequark: Agh, you're right: this isn't guaranteed AFAIK, it depends on linking order. Let's make it a linked list? Then it doesn't matter which order this is initialized in.

whitequark: Done with a linked list

whitequark:
Thanks!

PR #468, opened 2020-07-08
Support 4K parts directly

mmicko:
Support Lattice HX4K  directly, old way  (over HX8K) still supported without change.

### DIFF
--- a/gui/ice40/mainwindow.cc
+++ b/gui/ice40/mainwindow.cc
@@ -89,6 +89,8 @@ void MainWindow::new_proj()
         arch.insert("Lattice UP5K", ArchArgs::UP5K);
     if (Arch::isAvailable(ArchArgs::LP8K))
         arch.insert("Lattice LP8K", ArchArgs::LP8K);
+    if (Arch::isAvailable(ArchArgs::HX4K))
+        arch.insert("Lattice HX4K", ArchArgs::HX4K);
     if (Arch::isAvailable(ArchArgs::HX8K))
         arch.insert("Lattice HX8K", ArchArgs::HX8K);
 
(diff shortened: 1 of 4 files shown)


smunaut:
I'm not sure why bothering with that in the first place given that unlike the ECP5 there is no IDCODE or anything to fudge.
But if you go for adding "marketing variants", then I'd add all of them not just the HX4K.

whitequark:
> I'm not sure why bothering with that in the first place given that unlike the ECP5 there is no IDCODE or anything to fudge.

Because this is obnoxious and I shouldn't have to do it: https://github.com/nmigen/nmigen/blob/23da2fdda6963843c0b17c5857a889a16db7120f/nmigen/vendor/lattice_ice40.py#L78-L98

mmicko:
@smunaut about other variants, I am just not sure if we have confirmation that those are tested on real hardware, if they are I would like to add them as well.

daveshah1:
Can you at least add the lp4k for consistency too? I am relatively sure this one has been tested in the past.

mmicko:
for iCE40UP3K, iCE5LP2K and iCE5LP1K, would you all agree that it should contain at least additional text in options like

  --up3k                      set device type to iCE40UP3K (initial support only)
  --up5k                      set device type to iCE40UP5K

Please just suggest some better text :)

daveshah1:
I personally don't think we need any warning text. I've compared bitstreams in the past and they are identical. For these parts icecube doesn't even lock out part of the device, it just jmplements a total resource count limit.

mmicko:
Think all should be covered now, I have also use chance to add proper names in GUI since those did not make much sense

daveshah1:
Thanks!

Issue #478, opened 2020-07-25
Print better error message when instantiated i2c IP block lacks BUS_ADDR74 defparam


johnMamish:
I synthesized the below verilog file (some newlines removed for compactness) with yosys using
`yosys -p "read_verilog i2c_minimal.v; synth_ice40 -top i2c_minimal -json i2c_minimal.json"`

``` Verilog
module i2c_minimal(inout sda, inout scl, input digclk, input sb_wr_i, input sb_stb_i,
	           input [7:0] sb_adr_i, input [7:0] sb_dat_i, output [7:0] sb_dat_o,
	           output sb_ack_o);
    wire sda_in, sda_oe, sda_o, scl_in, scl_oe, scl_o;

    assign sda = sda_oe ? sda_o : 1'bz;
    assign sda_in = sda;
    assign scl = scl_oe ? scl_o : 1'bz;
    assign scl_in = scl;

    SB_I2C i2c1(.SBCLKI(digclk), .SBRWI(sb_wr_i), .SBSTBI(sb_stb_i),
	        .SBADRI7(sb_adr_i[7]), .SBADRI6(sb_adr_i[6]), .SBADRI5(sb_adr_i[5]), .SBADRI4(sb_adr_i[4]),
	        .SBADRI3(sb_adr_i[3]), .SBADRI2(sb_adr_i[2]), .SBADRI1(sb_adr_i[1]), .SBADRI0(sb_adr_i[0]),
	        .SBDATI7(sb_dat_i[7]), .SBDATI6(sb_dat_i[6]), .SBDATI5(sb_dat_i[5]), .SBDATI4(sb_dat_i[4]),
                .SBDATI3(sb_dat_i[3]), .SBDATI2(sb_dat_i[2]), .SBDATI1(sb_dat_i[1]), .SBDATI0(sb_dat_i[0]),
                .SCLI(scl_in), .SDAI(sda_in),
                .SBDATO7(sb_dat_o[7]), .SBDATO6(sb_dat_o[6]), .SBDATO5(sb_dat_o[5]), .SBDATO4(sb_dat_o[4]),
                .SBDATO3(sb_dat_o[3]), .SBDATO2(sb_dat_o[2]), .SBDATO1(sb_dat_o[1]), .SBDATO0(sb_dat_o[0]),
                .SBACKO(sback_o),
                .I2CIRQ(), .I2CWKUP(),
[...]

PR #479, opened 2020-07-25
improve error message when BUS_ADDR74 is unspecified for hard i2c/spi…

johnMamish:
Fix issue I talked about in [issue 478](https://github.com/YosysHQ/nextpnr/issues/478)

### DIFF
--- a/ice40/pack.cc
+++ b/ice40/pack.cc
@@ -1256,10 +1256,27 @@ static void pack_special(Context *ctx)
                     {std::make_tuple(id_SB_SPI, "0b0010"), Loc(25, 0, 1)},
                     {std::make_tuple(id_SB_I2C, "0b0011"), Loc(25, 31, 0)},
             };
-            if (map_ba74.find(std::make_tuple(ci->type, ci->params[ctx->id("BUS_ADDR74")].as_string())) ==
-                map_ba74.end())
-                log_error("Invalid value for BUS_ADDR74 for cell '%s' of type '%s'\n", ci->name.c_str(ctx),
-                          ci->type.c_str(ctx));
+            if (!ci->params[ctx->id("BUS_ADDR74")].is_string ||
+                map_ba74.find(std::make_tuple(ci->type, ci->params[ctx->id("BUS_ADDR74")].as_string())) ==
+                        map_ba74.end()) {
+                std::vector<std::string> valid_addr74_values;
+                for (const auto &kvp : map_ba74) {
+                    if (std::get<0>(kvp.first) == ci->type) {
+                        valid_addr74_values.push_back(std::get<1>(kvp.first));
+                    }
+                }
+
+                std::stringstream ss;
+                ss << "Invalid value for parameter BUS_ADDR74 for cell " << ci->name.c_str(ctx) << " of type "
+                   << ci->type.c_str(ctx) << ". Valid values include: ";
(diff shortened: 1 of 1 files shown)


ice40/pack.cc
```diff
@@ -1256,18 +1256,26 @@ static void pack_special(Context *ctx)
                     {std::make_tuple(id_SB_SPI, "0b0010"), Loc(25, 0, 1)},
                     {std::make_tuple(id_SB_I2C, "0b0011"), Loc(25, 31, 0)},
             };
-            if (map_ba74.find(std::make_tuple(ci->type, ci->params[ctx->id("BUS_ADDR74")].as_string())) ==
-                map_ba74.end())
-                log_error("Invalid value for BUS_ADDR74 for cell '%s' of type '%s'\n", ci->name.c_str(ctx),
-                          ci->type.c_str(ctx));
-            Loc bel_loc = map_ba74.at(std::make_tuple(ci->type, ci->params[ctx->id("BUS_ADDR74")].as_string()));
```
smunaut, line 1263: ?!? But you're removing part of the packing logic ?!?

smunaut:
So as noted above, I don't get why your changes remove the actual packing logic.

Also you need to run your change through clang-format to get formatting consistent with project guidelines.

ice40/pack.cc
johnMamish: hahaha holy crap. I definitely meant to delete ONE line, not TWELVE.

smunaut:
Also, it'd be nice if you could just squash and just force push 1 clean commit.
It's generally better for git bisect purposes if all the intermediates state are clean / buildable and valid.  Here it's a single logical change that doesn't really need to be split in several commits.

johnMamish:
@smunaut Anything else I should fix to merge this?

I know this is a small change, but it would've saved me a few hours when trying to figure out why nextpnr wouldn't place a hard i2c IP block for me. Because cells_sim.v in yosys has a default parameter value for ADDR_74, I assumed that I wouldn't need to specify one myself and it would default to using the "upper left" i2c block.

smunaut:
I mean functionally it looks goot to me.
I find it a bit weird you construct `valid_addr74_values` and then just scan through it to print the message, I don't see why you can't just combine the two iterations and never go through that intermediate step.

But in any case, I can't merge code, I'm just an interested observer because I did some work in that part of the packing code.

johnMamish:
>I find it a bit weird you construct `valid_addr74_values` and then just scan through it to print the message
Yeah. It is a bit weird. It's to take care of the whole aesthetic issue of having a trailing comma at the end of the list of valid values. I don't write C++ much, so I don't know if stringstream easily lets me clip off the last comma. This seemed like a readable way to do it :shrug:.

smunaut:
Huh ?

What I mean is, wouldn't this do the same thing ?

```C++
                std::stringstream ss;
                ss << "Invalid value for parameter BUS_ADDR74 for cell " << ci->name.c_str(ctx) << " of type "
                   << ci->type.c_str(ctx) << ". Valid values include: ";
                std::string delim = "";
                for (const auto &kvp : map_ba74) {
                    if (std::get<0>(kvp.first) == ci->type) {
                        ss << delim << std::get<1>(kvp.first);
                        delim = ", ";
                    }
                }
```

Issue #484, opened 2020-08-06
Nextpnr generic no longer throws python exceptions on assertion errors


pepijndevos:
See https://github.com/YosysHQ/apicula/pull/13

Apicula used to rely on catching an assertion error when a duplicated wire was added. But recent versions of Nextpnr appear to just crash rather than throwing a Python exception. This has been worked around, but is rather unfortunate.

daveshah1:
@mmicko can you look at this?

daveshah1:
Closing since #485 has been merged.

PR #485, opened 2020-08-07
Made proper exception translation

mmicko:
Possible fix for #484


pepijndevos:
I can test this weekend probably. Not sure I can meaningfully review this other than seeing if it works.

mmicko:
No hurry, but please do try with your test case.

PR #490, opened 2020-08-20
nextpnr: Outline API for partition pins

daveshah1:
This doesn't yet aim to actually implement full partial reconfig/partition support; instead it is the groundwork for such features.

Still TODO: add a Python function for creating these pins, and find out what breaks when we try and use them...

### DIFF
--- a/common/nextpnr.cc
+++ b/common/nextpnr.cc
@@ -271,6 +271,8 @@ const char *BaseCtx::nameOfGroup(GroupId group) const
 
 WireId Context::getNetinfoSourceWire(const NetInfo *net_info) const
 {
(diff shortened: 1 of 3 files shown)


mithro:
@daveshah1 - Any chance you talked to @andrewb1999 at all? He has been working on a GSoC project for things like this in SymbiFlow-Arch-Defs and VPR.

daveshah1:
Not yet, the internal API is fairly constrained by nextpnr internals but coordinating an external interface for specifying these (and other important things like timing across partition boundaries) would be useful.

gatecat:
Closing, if revisited this will probably follow a different approach

PR #496, opened 2020-09-04
Preserve cmd parameters  when loading json from GUI

mmicko:
When nextpnr is started with `nextpnr-ice40 --hx8k --package ct256 --gui ` after loading JSON from GUI it would switch to default architecture and package ignoring command line parameters. 
This patch makes sure that currently set command line parameters are effective until "New" option is set and target FPGA is changed.


daveshah1 (approved):
LGTM, thanks for noticing

PR #507, opened 2020-10-03
allow adding extra cxxflags from command line

edbordin:
This would be make it easy for me to try adding a few different build flags on MinGW (open-tool-forge/fpga-toolchain#50)


whitequark:
You don't need that; you can already do it just fine with `cmake -DCMAKE_CXX_FLAGS="..."` and so on.

PR #514, opened 2020-11-13
RelPtr: remove copy constructor and copy assignment

daveshah1:
These operations are meaningless for a data structure that references another structure relative to its location.

cc @pepijndevos


pepijndevos:
Thanks, I'll also apply that change to my WIP Gowin branch.

PR #515, opened 2020-11-14
Python code cleanup

mmicko:
Use pybind11 capabilites to get more clean and stable code. It is now possible to use python console in GUI after context change (new device selected)

Note to future self:
PYBIND11_EMBEDDED_MODULE -> also include PyImport_AppendInittab
py::initialize_interpreter -> also include adding current path to sys.path


daveshah1 (approved):
LGTM, thanks

mmicko:
Thanks, will just wait for tests to finish.

PR #516, opened 2020-11-17
missing '/local' at ecp5 installing

meijianyue:


mmicko:
Thanks. Old instructions did work for Linux users, but not for macos (due to install restrictions) enforcing /usr/local is then fine.

PR #518, opened 2020-11-17
Update CMakeLists.txt

BracketMaster:
It seems Prjtrellis has CMAKE define pytrellis as a shared module, not library.
On MacOS, this makes a difference as NextPNR expected ``.dylib``
instead of ``.so``. Shared module also work on Linux.


mmicko:
Confirm working. Thanks

PR #529, opened 2020-12-02
fix clang complaint on turbofish

aep:


daveshah1:
Thanks, I'll merge once the CI goes green as a double check

whitequark:
This shouldn't be necessary on C++11 and later. What's the error?

aep:
```
[ 50%] Building CXX object CMakeFiles/bbasm.dir/main.cc.o
/Users/aep/kram/nextpnr/bba/main.cc:89:69: error: a space is required between consecutive right angle brackets (use '> >')
    options.add_options()("files", po::value<std::vector<std::string>>(), "file parameters");
                                                                    ^~

```

daveshah1:
Oh, maybe the C++ standard isn't being set correctly in the bbasm cmake

aep:
oh yeah, im building each thing separate because i dont understand how to make cmake find dependencies, so that could be causing the issue

whitequark:
Add this to CMakeLists:
```
set(CMAKE_CXX_STANDARD 11)
set(CMAKE_CXX_STANDARD_REQUIRED YES)
```

PR #530, opened 2020-12-03
ice40: Derive PLL timing constraints

daveshah1:
This also refactors PLL packing into its own function, so that oscillators can be packed and auto-constrained first, then an oscillator driving a PLL will derive constraints correctly.


whitequark:
Very nice!

PR #536, opened 2020-12-08
cmake: Disable building GUI by default

daveshah1:
Following large numbers of [GUI-related issues](https://github.com/YosysHQ/nextpnr/issues?q=is%3Aissue+qt5), it seems like disabling the GUI by default is going to be the path of least resistance, given that the command line flow is what most people are primarily using.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -5,7 +5,7 @@ task:
     memory: 20
     dockerfile: .cirrus/Dockerfile.ubuntu16.04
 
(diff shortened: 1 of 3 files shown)


mmicko (approved):
LGTM

whitequark (commented):
There's more references to Qt packages in sample command lines in README—I would personally split that section into two.

daveshah1:
> There's more references to Qt packages in sample command lines in README—I would personally split that section into two.

Updated, how do you think it looks now?

whitequark (approved):
Very nice!

mithro:
BTW - We ship separate `nextpnr` and a `nextpnr-gui` GUI packages for this exactly this reason.

PR #541, opened 2020-12-11
Fix opensuse build error

### LINKED ISSUE #467: Build error on opensuse leap
I'm using opensuse leap and tried to build nextpnr from sources (commit 137241cfef385a68d2ce732a5b3283458790543d). Eigen 3.3.4 is installed. The build fails with the following output:
```bash
[ 68%] Building CXX object CMakeFiles/nextpnr-ice40.dir/common/placer_heap.cc.o
/home/martin/dev/nextpnr/common/placer_heap.cc:37:10: fatal error: Eigen/Core: Datei oder Verzeichnis nicht gefunden
 #include <Eigen/Core>
          ^~~~~~~~~~~~
compilation terminated.
make[2]: *** [CMakeFiles/nextpnr-ice40.dir/build.make:197: CMakeFiles/nextpnr-ice40.dir/common/placer_heap.cc.o] Fehler 1
make[1]: *** [CMakeFiles/Makefile2:226: CMakeFiles/nextpnr-ice40.dir/all] Fehler 2
make: *** [Makefile:147: all] Fehler 2
```

My workaround was to create a symbolic link `sudo ln -s /usr/include/eigen3/Eigen /usr/include/Eigen` as described [here](https://github.com/opencv/opencv/issues/14868). Another workaround would be to use `# include <eigen3/Eigen/Core>` instead of `# include <Eigen/Core>`.

I guess it works on other linux distributions, since I didn't find any related bug report. Is there a way to fix this for all distributions?

marph91:
Fixes #467.

The `EIGEN3_*` variables could be removed, too. But for the fix, I wanted to keep the changes minimal.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -261,6 +261,10 @@ foreach (family ${ARCH})
         set(family_targets ${family_targets} ${PROGRAM_PREFIX}nextpnr-${family}-test)
     endif()
 
+    if(BUILD_HEAP)
+        target_link_libraries(${PROGRAM_PREFIX}nextpnr-${family} PRIVATE Eigen3::Eigen)
+    endif()
+
     # Include the family-specific CMakeFile
     include(${family}/family.cmake)
     foreach (target ${family_targets})


marph91:
I can't test it with Ubuntu right now, but it seems like Ubuntu 16.04 uses libeigen3-dev at the version [3.3~beta1-2](https://gitlab.com/libeigen/eigen/-/blob/3.3-beta1/cmake/Eigen3Config.cmake.in). This version doesn't have the eigen target yet.

If I find the time, I will try to setup Ubuntu for local builds and try to fix the error.

PR #542, opened 2020-12-15
Gowin target

pepijndevos:
As the title suggests, it's a Gowin target for Nextpnr. It's a bit rough around the edges, but I just successfully ran a full flow and tested it on a GW1NR-9.

It's mostly a direct port from the old nexpnr-generic target, with added support for all LUTs, DFF reset/enable, correct validation checks, and pin constraints.

This means there is no deduplication and loading isn't instantaneous. When I add support for larger GW2A devices, I intend to revisit this target to add those things as needed. For the GW1N devices, it's not a real problem.

The main problem for now is that routing the attosoc demo takes forever. It has 53% utilization, but at about 8k unrouted nets it basically makes no progress at all.

For now the main things to do are

- [x] remove all debugging leftovers
- [x] figure out if attosoc routing congestion can be improved
- [x] address any feedback on this PR

### How to test

I have built this branch with the following CMake variables, where the notable thing is that the python interpreter is the same one as is used for Project Apicula. If you're using system Python, I guess these options are not needed.

```json
{
    "ARCH": "gowin",
    "BUILD_GUI": false,
    "PYTHON_EXECUTABLE": "/home/pepijn/code/apicula/env/bin/python",
    "GOWIN_BBA_EXECUTABLE": "/home/pepijn/code/apicula/env/bin/gowin_bba",
}
```

I'm pushing a new Apycula to Pypi as I write this, so you should be able to `pip install apycula` soon.

From there you should be able to use the latest yosys to `synth_gowin` an example program.
A simple way is to just `make blinky.json` in https://github.com/YosysHQ/apicula/tree/master/examples

Then you should be able to run nextpnr like so
```
nxtpnr-gowin --json apicula/examples/attosoc.json \
                --write out.json \
                --device GW1NR-9 \
                --family GW1N-9 \
                --package QN881 \
                --speed C6/I5 \
                --cst apicula/examples/gw1n9.cst
```

And then finally pack the bitstream with 

```
gowin_pack -d GW1N-9 -o blinky.json out.json
```


gowin/arch.cc
```diff
@@ -0,0 +1,1248 @@
+/*
+ *  nextpnr -- Next Generation Place and Route
+ *
+ *  Copyright (C) 2018  Clifford Wolf <clifford@symbioticeda.com>
```
daveshah1, line 4: Please update and add your own name

gowin/arch.cc
```diff
@@ -0,0 +1,1248 @@
...
+        row -= segment;
+        break;
+    case 'E':
+        col -= segment;
+        break;
+    case 'W':
+        col += segment;
+        break;
+    default:
+        snprintf(buf, 32, "R%dC%d_%s", row + 1, col + 1, wirename.c_str());
```
daveshah1, line 279: I'm not all that happy with snprintfs and maps of IdStrings, I'd really rather see a database with flat indices (like ice40), but I guess it's OK if  you are planning to move away from it

gowin/arch.cc
```diff
@@ -0,0 +1,1248 @@
...
+        getCtx()->settings[getCtx()->id("place")] = 1;
+        archInfoToAttributes();
+        return retVal;
+    } else if (placer == "sa") {
+        bool retVal = placer1(getCtx(), Placer1Cfg(getCtx()));
+        getCtx()->settings[getCtx()->id("place")] = 1;
+        archInfoToAttributes();
+        return retVal;
+    } else {
+        log_error("Generic architecture does not support placer '%s'\n", placer.c_str());
```
daveshah1, line 1054: Gowin, not Generic

gowin/arch.cc
```diff
@@ -0,0 +1,1248 @@
...
+    const NetInfo *clk[4] = {nullptr, nullptr, nullptr, nullptr};
+    const NetInfo *ce[4] = {nullptr, nullptr, nullptr, nullptr};
+    const NetInfo *lsr[4] = {nullptr, nullptr, nullptr, nullptr};
+    int group = -1;
+    for (int i = 0; i < count; i++) {
+        const CellInfo *ci = cells[i];
+        if (ci->is_slice && ci->slice_clk != nullptr) {
+            Loc loc = getBelLocation(ci->bel);
+            int cls = loc.z / 2;
+            bool ff_used = ci->params.at(id_FF_USED).as_bool();
```
daveshah1, line 1222: You could get a small speedup from adding ff_used to ArchCellInfo and assigning it in assignArchInfo

gowin/arch_pybindings.cc
```diff
@@ -0,0 +1,235 @@
...
+    readonly_wrapper<Context, decltype(&Context::cells), &Context::cells, wrap_context<CellMap &>>::def_wrap(ctx_cls,
+                                                                                                             "cells");
+    readonly_wrapper<Context, decltype(&Context::nets), &Context::nets, wrap_context<NetMap &>>::def_wrap(ctx_cls,
+                                                                                                          "nets");
+
+    fn_wrapper_2a_v<Context, decltype(&Context::addClock), &Context::addClock, conv_from_str<IdString>,
+                    pass_through<float>>::def_wrap(ctx_cls, "addClock");
+
+    // Generic arch construction API
+    fn_wrapper_4a_v<Context, decltype(&Context::addWire), &Context::addWire, conv_from_str<IdString>,
```
daveshah1, line 155: All these add* pybindings aren't realistically needed, and can be removed (doing so will probably appreciably reduce build time and build memory usage)

gowin/constids.inc
```diff
@@ -0,0 +1,429 @@
...
+X(SPINE_TAP_PCLK)
+X(TAP_BRANCH_PCLK)
+X(BRANCH_PCLK)
+X(clksetpos)
+X(clkholdpos)
+X(clk_qpos)
+X(a_f)
+X(b_f)
+X(c_f)
+X(d_f)
```
daveshah1, line 429: Please add a newline at the end of file

gowin/main.cc
```diff
@@ -0,0 +1,82 @@
...
+  protected:
+    po::options_description getArchOptions() override;
+};
+
+GowinCommandHandler::GowinCommandHandler(int argc, char **argv) : CommandHandler(argc, argv) {}
+
+po::options_description GowinCommandHandler::getArchOptions()
+{
+    po::options_description specific("Architecture specific options");
+    specific.add_options()("device", po::value<std::string>(), "device name");
```
daveshah1, line 48: I would like all new nextpnr arches to follow nextpnr-nexus and take a single device string like `GW1N-LV1QN48C6/I5` and extract family/package/speed from it. This will make integration with things like nMigen and LiteX easier.

gowin/pack.cc
```diff
@@ -0,0 +1,290 @@
...
+static void pack_lut_lutffs(Context *ctx)
+{
+    log_info("Packing LUT-FFs..\n");
+
+    std::unordered_set<IdString> packed_cells;
+    std::vector<std::unique_ptr<CellInfo>> new_cells;
+    for (auto cell : sorted(ctx->cells)) {
+        CellInfo *ci = cell.second;
+        if (ctx->verbose)
+            log_info("cell '%s' is of type '%s'\n", ci->name.c_str(ctx), ci->type.c_str(ctx));
```
daveshah1, line 41: tbf this probably needs refactoring in `generic` and other arches too, but it's neater to use `ctx->nameOf(ci)` here

CMakeLists.txt
```diff
@@ -107,9 +107,9 @@ if (MSVC)
 else()
     set(CMAKE_CXX_FLAGS_DEBUG "-Wall -fPIC -ggdb -pipe")
     if (USE_OPENMP)
-        set(CMAKE_CXX_FLAGS_RELEASE "-Wall -fPIC -O3 -g -pipe -fopenmp")
+        set(CMAKE_CXX_FLAGS_RELEASE "-Wall -fPIC -O0 -g -pipe -fopenmp")
```
daveshah1, line 110: Revert this back to O3

gowin/arch.cc
```diff
@@ -0,0 +1,1248 @@
...
+}
+
+const std::vector<BelId> &Arch::getBelsByTile(int x, int y) const { return bels_by_tile.at(x).at(y); }
+
+bool Arch::getBelGlobalBuf(BelId bel) const { return bels.at(bel).gb; }
+
+uint32_t Arch::getBelChecksum(BelId bel) const
+{
+    // FIXME
+    return 0;
```
daveshah1, line 784: returning the index here is fine

gowin/arch.h
```diff
@@ -0,0 +1,435 @@
...
+    std::vector<GroupId> getGroups() const;
+    const std::vector<BelId> &getGroupBels(GroupId group) const;
+    const std::vector<WireId> &getGroupWires(GroupId group) const;
+    const std::vector<PipId> &getGroupPips(GroupId group) const;
+    const std::vector<GroupId> &getGroupGroups(GroupId group) const;
+
+    delay_t estimateDelay(WireId src, WireId dst) const;
+    delay_t predictDelay(const NetInfo *net_info, const PortRef &sink) const;
+    delay_t getDelayEpsilon() const { return 0.001; }
+    delay_t getRipupDelayPenalty() const { return 0.015; }
```
daveshah1, line 387: I think you will probably get better results with a higher ripup penalty. For perspective ECP5 uses 400ps. epsilon might be worth increasing too.

gowin/arch.cc
```diff
@@ -0,0 +1,1248 @@
...
+    }
+}
+
+void Arch::read_cst(std::istream &in) {
+    std::regex iobre = std::regex("IO_LOC +\"([^\"]+)\" +([^ ;]+);");
+    std::smatch match;
+    std::string line;
+    while (!in.eof()) {
+        std::getline(in, line);
+        if(!std::regex_match(line, match, iobre)) continue;
```
daveshah1, line 487: Can you add a warning if the regex fails but the line isn't all blank? An important constraint could be being missed

gowin/arch.cc
```diff
@@ -0,0 +1,1248 @@
...
+        const PairPOD *belname = pairLookup(package->pins.get(), package->num_pins, pinname.index);
+        if ( belname == nullptr)
+            log_error("Pin %s not found", pinname.c_str(this));
+        //BelId bel = getBelByName(belname->src_id);
+        // for (auto cell : sorted(cells)) {
+        //     std::cout << cell.first.str(this) << std::endl;
+        // }
+        auto it = cells.find(net);
+        if (it == cells.end())
+            log_error("Cell %s not found", net.c_str(this));
```
daveshah1, line 500: Unless the Gowin tools behave like this, this should be a `continue` rather than an error - it's common for board constraint files to declare all the pins but only a subset are used in the design

gowin/arch.cc
```diff
@@ -0,0 +1,1248 @@
...
+}
+
+void Arch::read_cst(std::istream &in) {
+    std::regex iobre = std::regex("IO_LOC +\"([^\"]+)\" +([^ ;]+);");
+    std::smatch match;
+    std::string line;
+    while (!in.eof()) {
+        std::getline(in, line);
+        if(!std::regex_match(line, match, iobre)) continue;
+        std::cout << match[1] << " " << match[2] << std::endl;
```
daveshah1, line 488: Remove this debugging, or replace it with a friendlier `log_info`

daveshah1:
Thank you very much for submitting this. Please understand if some of my review comments appear harsh, as I am using a standard that is higher than the current quality of the nextpnr codebase - some of the things I have commented on I know just came from generic as-is - because improving things is always good :).

gowin/arch.cc
```diff
@@ -0,0 +1,1248 @@
...
+    pip_ids.push_back(name);
+
+    if (int(tilePipDimZ.size()) <= loc.x)
+        tilePipDimZ.resize(loc.x + 1);
+
+    if (int(tilePipDimZ[loc.x].size()) <= loc.y)
+        tilePipDimZ[loc.x].resize(loc.y + 1);
+
+    gridDimX = std::max(gridDimX, loc.x + 1);
+    gridDimY = std::max(gridDimY, loc.x + 1);
```
daveshah1, line 92: loc.z

pepijndevos:
Makes complete sense. I'll probably get back to your feedback next week, because of a deadline coming Friday. I'm really excited to finally have a proper* nextpnr target, as a Christmas present to the community ;-)

* I agree the runtime loading of maps of IdString is not ideal...

gowin/arch.cc
```diff
@@ -0,0 +1,1248 @@
...
+    cellTiming[cell].clockingInfo[port].push_back(ci);
+    cellTiming[cell].portClasses[port] = TMG_REGISTER_OUTPUT;
+}
+
+// ---------------------------------------------------------------
+
+// TODO represent wires more intelligently.
+IdString Arch::wireToGlobal(int &row, int &col, const DatabasePOD *db, IdString &wire)
+{
+    std::string wirename = wire.str(this);
```
daveshah1, line 253: `const std::string &` avoids a copy

gowin/arch.cc
```diff
@@ -0,0 +1,1248 @@
...
+        break;
+    default:
+        if (wire.str(this).rfind("SPINE", 0) == 0) {
+            glbsrc = ID_CENT_SPINE_PCLK;
+        } else if (wire.str(this).rfind("UNK", 0) == 0) {
+            glbsrc = ID_PIO_CENT_PCLK;
+        }
+        break;
+    }
+    if (len != id("")) {
```
daveshah1, line 467: I suspect `!= IdString()` will be faster

gowin/arch.cc
```diff
@@ -0,0 +1,1248 @@
...
+                z++;
+            case ID_LUT1:
+                z++;
+            case ID_LUT0:
+                // common LUT+DFF code
+                snprintf(buf, 32, "R%dC%d_SLICE%d", row + 1, col + 1, z);
+                belname = id(buf);
+                addBel(belname, id_SLICE, Loc(col, row, z), false);
+                snprintf(buf, 32, "R%dC%d_F%d", row + 1, col + 1, z);
+                wirename = id(buf);
```
daveshah1, line 623: is there any need for the wirename temporary variable - wouldn't doing `id(buf)` as an argument to `addBelOutput` etc be neater?

daveshah1: I also think you could have a list of pin names and do this in a loop

gowin/arch.cc
```diff
@@ -0,0 +1,1248 @@
...
+            for (int i = 0; i < 4; i++) {
+                DelayInfo delay = delayLookup(speed->lut.timings.get(), speed->lut.num_timings, port_delay[i]);
+                // DelayInfo delay = getDelayFromNS(0.1);
+                addCellTimingDelay(cname, ports[i], id_F, delay);
+            }
+
+        } else {
+            ci->is_slice = false;
+        }
+        ci->user_group = int_or_default(ci->attrs, id("PACK_GROUP"), -1);
```
daveshah1, line 1207: `PACK_GROUP` looks like dead code from `generic` to me

daveshah1:
Great! A few other things I would suggest:

 - [ ] add a stub of the GUI files so that CI passes and you can do a `BUILD_GUI=yes` build combining gowin and other arches even if you don't care about the GUI for gowin ATM
 - [ ] add a brief note about Gowin to the README, even if just mentioning that it is highly experimental

gowin/arch.cc
pepijndevos: Should I do anything with @clairexen's entry? Seems to be using her deadname across the whole project except in _one_ location, so not sure what the accepted approach is for a new copy of old code.

daveshah1: Yes, update it in new code (afaik Yosys is mostly doing the same)

gowin/arch.cc
```diff
@@ -484,8 +485,14 @@ void Arch::read_cst(std::istream &in) {
...
     while (!in.eof()) {
         std::getline(in, line);
-        if(!std::regex_match(line, match, iobre)) continue;
-        std::cout << match[1] << " " << match[2] << std::endl;
+        if(!std::regex_match(line, match, iobre)) {
+            // empty line or comment
+            if(line.empty() || line.rfind("//", 0) == 0)
+                continue;
+            else
+                log_warning("Invalid constraint: %s", line.c_str());
```
daveshah1, line 493: Missing newline in message

gowin/arch.cc
```diff
@@ -496,8 +503,10 @@ void Arch::read_cst(std::istream &in) {
         //     std::cout << cell.first.str(this) << std::endl;
         // }
         auto it = cells.find(net);
-        if (it == cells.end())
-            log_error("Cell %s not found", net.c_str(this));
+        if (it == cells.end()) {
+            log_info("Cell %s not found", net.c_str(this));
```
daveshah1, line 507: missing newline here too

gowin/cells.cc
```diff
@@ -83,14 +84,14 @@ void lut_to_lc(const Context *ctx, CellInfo *lut, CellInfo *lc, bool no_dff)
     }
 
     if (no_dff) {
-        lc->params[id_FF_USED] = 0;
+        lc->ff_used = false;
```
daveshah1, line 87: Sorry, I may have explained this badly before. The correct way to do it is to set a param here and then set the ArchCellInfo field to mirror that param in `assignArchInfo` - ArchCellInfo should never be the only storage for important things, to avoid creating any more problems for loading/storing designs or post/pack simulations than needed (even if both these things are admittedly experimental right now)

gowin/arch.cc
pepijndevos: They are a mix of inputs and outputs, so I think it would be less clear in the end.

gowin/cells.cc
pepijndevos: FF_USED is not actually used on the Apicula side of things, so it's no important info. Actually the only place it's used is in validation that no DFF are used in the top two slices that don't have any.

daveshah1: And that would break if you saved the netlist between pack and place

pepijndevos:
Okay, I think I addressed all of your feedback now.

Routing attosoc is still.... not ideal. Increasing the ripup penalty did improve it slightly, now it slows down at 5k instead of 8k...

Also the build is still failing. Currently because Apicula requires Python 3.6 or higher, but of course the CI does not install Apicula at all.

pepijndevos:
Some things that need to be done still:

- [ ] add Apicula to CI, will require Python 3.6
- [ ] hunt down bug that breaks attosoc example. Issue with global clock routing I think, bug could be on the Apicula side.

optionally:
- [ ] better timing estimation
- [ ] dedicated global router
- [ ] more comprehensive constraints and configuration (clock domains?)

Anything else?

daveshah1:
That list looks reasonable to me. I think making CI pass is the number one priority to get this merged which I'd rather do sooner than later, as GitHub doesn't charge for follow-up PRs.

smunaut:
Are you planning on squashing the commits for the merge ?
Or at least adding a `gowin: ` prefix to their title ?

pepijndevos:
Squashing them seems like a good plan, because I've also commited some WIP things that will be a pain if someone ever tries to bisect across these commits.

daveshah1:
Now that CI is working, I think all that's left is squashing the commits if you want to do that; or I can merge-and squash.

An informal agreement to tidy up the arch database at some point would be nice too :)

pepijndevos:
Yea, feel free to merge-and-squash. Saves me some googling.

There will definitely be follow-up PRs to fix issues, clean up stuff, and add features.
I think once I add GW2A support going back and adding deduplication will be unavoidable.

daveshah1:
Done :champagne: :confetti_ball:

whitequark:
Nice!

PR #543, opened 2020-12-17
Use std::string::find(char c) when searching for a single character.

tcal-x:
Signed-off-by: Tim Callahan <tcal@google.com>


daveshah1:
Thanks!

PR #548, opened 2020-12-30
C++17 compatibility: Don't use std::random_shuffle

per-gron:
std::random_shuffle deprecated in C++14 and was removed in C++17.


daveshah1:
Thanks!

PR #557, opened 2021-01-28
RFC: Initial refactoring of placer API.

litghost:
This PR is in service of seperating the requirement that cell type == BEL type.  The concrete example of this is Xilinx fabrics, where there are multiple FF types, multiple LUT types.  To avoid contortions to force incoming cell type to merge with the BEL type, some new arch APIs are being proposed.

The first is the simplest, which is `isValidBelForCellType`.  For arch's where the cell and BEL type are equivalent, this is simply:
```
    bool isValidBelForCellType(IdString cell_type, BelId bel) const {
        return cell_type == getBelType(bel);
    }
```

The other set of APIs is to enable the analytic placer to operate.  The analytic placer wants to partition the design to seperate LUTs, FFs, BRAMs, etc.  This is where the "Partition Methods" come in.  These provide a way for an analytic placer to partition BELs into buckets for AP purposes.

### DIFF
--- a/common/placer_heap.cc
+++ b/common/placer_heap.cc
@@ -523,20 +523,29 @@ class HeAPPlacer
                 bool placed = false;
                 int attempt_count = 0;
                 while (!placed) {
-                    if (!available_bels.count(ci->type) || available_bels.at(ci->type).empty()) {
-                        log_error("Unable to place cell '%s', no BELs remaining to implement cell type '%s'\n",
-                                ci->name.c_str(ctx),
-                                ci->type.c_str(ctx));
-                    }
                     ++attempt_count;
                     if (attempt_count > 25000) {
                         log_error("Unable to find a placement location for cell '%s'\n", ci->name.c_str(ctx));
+                    }
+
+                    // Make sure this cell type is in the available BEL map at
+                    // all.
+                    if (!available_bels.count(ci->type)) {
+                        log_error("Unable to place cell '%s', no BELs remaining to implement cell type '%s'\n",
+                                ci->name.c_str(ctx),
+                                ci->type.c_str(ctx));
                     }
 
                     // Find an unused BEL from bels_for_cell_type.
                     auto &bels_for_cell_type = available_bels.at(ci->type);
                     BelId bel;
                     while(true) {
+                        if (bels_for_cell_type.empty()) {
+                            log_error("Unable to place cell '%s', no BELs remaining to implement cell type '%s'\n",
+                                    ci->name.c_str(ctx),
+                                    ci->type.c_str(ctx));
+                        }
+
                         BelId candidate_bel = bels_for_cell_type.back();
                         bels_for_cell_type.pop_back();
                         if(bels_used.count(candidate_bel)) {
--- a/common/placer_heap.cc
+++ b/common/placer_heap.cc
@@ -177,21 +177,21 @@ class HeAPPlacer
 
         std::vector<std::tuple<CellInfo *, BelId, PlaceStrength>> solution;
 
-        std::vector<std::unordered_set<PartitionId>> heap_runs;
-        std::unordered_set<PartitionId> all_partitions;
-        std::unordered_map<PartitionId, int> partition_count;
+        std::vector<std::unordered_set<BelBucketId>> heap_runs;
+        std::unordered_set<BelBucketId> all_buckets;
+        std::unordered_map<BelBucketId, int> bucket_count;
 
         for (auto cell : place_cells) {
-            PartitionId partition = ctx->getPartitionForCellType(cell->type);
-            if (!all_partitions.count(partition)) {
-                heap_runs.push_back(std::unordered_set<PartitionId>{partition});
-                all_partitions.insert(partition);
+            BelBucketId bucket = ctx->getBelBucketForCellType(cell->type);
+            if (!all_buckets.count(bucket)) {
+                heap_runs.push_back(std::unordered_set<BelBucketId>{bucket});
+                all_buckets.insert(bucket);
             }
-            partition_count[partition]++;
+            bucket_count[bucket]++;
         }
         // If more than 98% of cells are one cell type, always solve all at once
         // Otherwise, follow full HeAP strategy of rotate&all
-        for (auto &c : partition_count) {
+        for (auto &c : bucket_count) {
             if (c.second >= 0.98 * int(place_cells.size())) {
                 heap_runs.clear();
                 break;
@@ -205,7 +205,7 @@ class HeAPPlacer
             heap_runs.clear();
         }
 
-        heap_runs.push_back(all_partitions);
+        heap_runs.push_back(all_buckets);
         // The main HeAP placer loop
         log_info("Running main analytical placer.\n");
         while (stalled < 5 && (solved_hpwl <= legal_hpwl * 0.8)) {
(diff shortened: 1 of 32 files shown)


docs/archapi.md
```diff
@@ -463,21 +471,66 @@ Cell Delay Methods
...
 Return the _clocking info_ (including port name of clock, clock polarity and setup/hold/clock-to-out times) of a
 port. Where ports have more than one clock edge associated with them (such as DDR outputs), `index` can be used to obtain
 information for all edges. `index` must be in [0, clockInfoCount), behaviour is undefined otherwise.
 
+Partition Methods
+-----------------
+
+Partitions are subsets of BelIds and cell types used by analytic placer to
+seperate types of BELs during placement.  Typical partitions are:
+ - All LUT BELs
```
daveshah1, line 491: ime HeAP has worked better with LUTs and FFs in the same partition - this may be more of an implementation detail than a general truth though

generic/arch.h
```diff
@@ -271,6 +271,34 @@ struct Arch : BaseCtx
     bool place();
     bool route();
 
+    std::vector<IdString> getCellTypes() const {
+        return {};
```
daveshah1, line 275: This should be being set up dynamically as bels are added to `generic`

nexus/arch.cc
```diff
@@ -171,6 +171,19 @@ Arch::Arch(ArchArgs args) : args(args)
...
+    for(BelId bel : getBels()) {
+        bel_types.insert(getBelType(bel));
+    }
+
+    for(IdString bel_type : bel_types) {
+        cell_types.push_back(bel_type);
+
+        PartitionId partition;
+        partition.name = bel_type;
+        partitions.push_back(partition);
```
daveshah1, line 185: as Nexus uses a split-slice arrangement, to match the way that HeAP cellgroups (which are essentially the precursor to partitions) are being set up `OXIDE_FF` and `OXIDE_COMB` should really be combined into one `SLICE` partition

daveshah1:
Overall approach LGTM so far

Some general outstanding issues:
 - is `Partition` really the best word here? to me it is a bit ambiguous and could also apply to totally different concepts like SLRs
 - we need Python bindings for newly added APIs and types
 - the [placement](https://github.com/YosysHQ/nextpnr/blob/master/docs/coding.md#developing-cad-algorithms---placement) section of coding.md should be updated with a brief description of how to use the new APIs from a developing-a-placer point of view

docs/archapi.md
litghost: That may be true, but I think this is a problem with the documentation, rather than the implementation.  The CutSpreader needs occupancy / bel counts, and this is definitely based on effectively interchangeable locations rather than LUT+FF.

I'll try to reword this to clarify.

nexus/arch.cc
litghost: I believe I did that here: https://github.com/YosysHQ/nextpnr/pull/557/files#diff-f8b12dcf636347f6473ca026e04cce243869c14dd298d2be18d1801a48f3635dR651-R652

Based on my understanding of placer_heap, the AP should run as:
 - All partitions
 - OXIDE_FF / OXIDE_COMB partitions
 - Each other BEL type / partition by themselves

Did I missing something?

litghost: > OXIDE_FF and OXIDE_COMB should really be combined into one SLICE partition

I'd argue that because of how occupancy / bel counts are tracked (both before and after this PR), these need to remain as separate partitions because occupancy counts are only really relevant as FF vs COMB, not together.  For example putting all the FF's in one cut and all the COMB's in the other would result in a very poor solution.

litghost:
> * is `Partition` really the best word here? to me it is a bit ambiguous and could also apply to totally different concepts like SLRs

No disagreement here.  I'll noodle over some potential options.  Nothing else comes to mind immediatly.

daveshah1:
> No disagreement here. I'll noodle over some potential options. Nothing else comes to mind immediatly.

You talked about 'buckets' above. That might form part of a better name?

litghost:
> > No disagreement here. I'll noodle over some potential options. Nothing else comes to mind immediatly.
> 
> You talked about 'buckets' above. That might form part of a better name?

What do you think about `BelBucket`?

daveshah1:
> What do you think about `BelBucket`?

Sounds good!

litghost:
> > What do you think about `BelBucket`?
> 
> Sounds good!

I'll rename it after I figure out why the ice40 smoke test exploded (e.g. https://cirrus-ci.com/task/5698940160966656?command=smoketest_ice40#L0 )

generic/arch.h
litghost: Fixed!

litghost:
- [x] Is Partition really the best word here? to me it is a bit ambiguous and could also apply to totally different concepts like SLRs
  - `Partition` -> `BelBucket`
- [ ] We need Python bindings for newly added APIs and types
- [x] The placement section of coding.md should be updated with a brief description of how to use the new APIs from a developing-a-placer point of view
- [x] Update the terminology section of faq.md

docs/archapi.md
litghost: I've tried to reword this.

litghost:
@daveshah1 I believe I've address all review comments.

docs/coding.md
```diff
@@ -39,26 +39,58 @@ Note that `IdString`s need a `Context` (or `BaseCtx`) pointer to convert them ba
...
+}
+```
+
+The alternative is to implement a fast `isValidBelForCellType` method that
+determines if this cell type can be bound to this BEL.
+
 ## Developing CAD algorithms - placement
 
-The job of the placer in nextpnr is to find a suitable bel for each cell in the design; while respecting legality and relative constraints.
+The job of the placer in nextpnr is to find a suitable BEL for each cell in the design; while respecting legality and relative constraints.
```
daveshah1, line 72: bit of a nitpick but the general convention in nextpnr has not been to have bel/pip in all-caps

common/placer_heap.cc
```diff
@@ -544,28 +523,49 @@ class HeAPPlacer
...
                         log_error("Unable to find a placement location for cell '%s'\n", ci->name.c_str(ctx));
-                    BelId bel = available_bels.at(ci->type).back();
-                    available_bels.at(ci->type).pop_back();
+                    }
+
+                    // Find an unused BEL from bels_for_cell_type.
+                    auto &bels_for_cell_type = available_bels.at(ci->type);
+                    BelId bel;
+                    while(true) {
+                        BelId candidate_bel = bels_for_cell_type.back();
```
daveshah1, line 540: is there some error handling needed for the case that `bels_for_cell_type` might become empty during the loop?

litghost: Ya, fixed.

docs/coding.md
litghost: Sure.

litghost:
@daveshah1 What else would you like to see with this PR prior to merging?

daveshah1:
I think I'm happy with it but I wanted to double check a few test designs (which will probably be tomorrow). 

One other thing is have you run a `make clangformat` on it?

litghost:
> One other thing is have you run a `make clangformat` on it?

Not yet.  Will do.

litghost:
> I think I'm happy with it but I wanted to double check a few test designs (which will probably be tomorrow).

Do you think it make sense to create a repo with the test designs?  Eventually we'll want to extend FPGA tool perf with those design and track nextpnr runtime and QoR over time.

daveshah1:
Yes definitely - historically we had https://github.com/YosysHQ/nextpnr-bench but it's in need of some TLC and some ECP5 test designs. I will try and get something up and running tomorrow.

daveshah1:
Still working on tidying up the benchmark set enough for it to be generally useful, but it's clear from the tests I've done so far that this hasn't introduced any regressions so good to go once the small conflicts are sorted.

litghost:
> Still working on tidying up the benchmark set enough for it to be generally useful, but it's clear from the tests I've done so far that this hasn't introduced any regressions so good to go once the small conflicts are sorted.

Rebase is in.

PR #561, opened 2021-01-29
Use IdStringList for bel/wire/pip/group names

### LINKED ISSUE #558: More efficient storage of wire/bel/pip names
Based on recent discussions with @litghost (in the context of `getPipName`/`getPipByName` which are currently rarely used but might see more use in the future) and @TechnoMancer (in the context of nextpnr-generic) it is very easy for pip names in particular to cause a memory explosion due to needing to store the full name string for every pip.

The easiest solution would be to use a concatenated list of `IdString`s for object names (starting with wires/bels/pips but potentially extending to other objects too). Typically this would be a tile/site name in the first part(s) followed by an object name within the tile.

Outstanding issues:
 - due to different arches having different conventions, the delimiter will need to be flexible
 - efficient ways of storing these string lists will be needed (probably statically allocated for ≤4 strings i.e. 128 bits and heap allocated above that?) 
 - fast ways of parsing and generating them will be needed
 - functions like `nameOfPip` that return a `const char *` and are heavily used with `%s` in logging will need some rework because we no longer have a stored string we can just return a pointer too. Instead we'll need a string buffer pool like Yosys uses: https://github.com/YosysHQ/yosys/blob/708eb327a1e88f726b8d426dba35abc29339a634/kernel/log.cc#L605-L613
 - there will probably be a breaking change to the `nextpnr-generic` API to not require a full name for a bel/wire/pip but instead automatically create one given a tile prefix and object suffix

daveshah1:
As bel/wire/pip/group names are generally comprised of a tile component (which itself contains many objects) and an object-in-tile component (which is the same across all tiles of the same type); storing the full name string for all objects, forever, is very inefficient.

This introduces a new type, `IdStringList`, comprising of a series of `IdString`s and an arch-specific delimiter, for these identifiers that avoids this waste.

Currently this PR contains a proof-of-concept RFC implementation for ECP5 only, excluding the GUI (which must be disabled for it even to build) but including the Python API.

Fixes #558 ; cc @litghost

### DIFF
--- a/docs/archapi.md
+++ b/docs/archapi.md
@@ -92,13 +92,16 @@ Get Z dimension for the specified tile for bels. All bels with at specified X an
 
 Get Z dimension for the specified tile for pips. All pips with at specified X and Y coordinates must have a Z coordinate in the range `0 .. getTileDimZ(X,Y)-1` (inclusive).
 
+### char getNameDelimiter() const
+
+Returns a delimiter that can be used to build up bel, wire and pip names out of hierarchical components (such as tiles and sites) to avoid the high memory usage of storing full names for every object.
+
 Cell Methods
 -----------
 
 ### const\_range\<IdString\> getCellTypes() const
 
 Get list of cell types that this architecture accepts.
-
 
 Bel Methods
 -----------
--- a/docs/archapi.md
+++ b/docs/archapi.md
@@ -106,11 +106,11 @@ Get list of cell types that this architecture accepts.
 Bel Methods
 -----------
 
-### BelId getBelByName(IdString name) const
+### BelId getBelByName(IdStringList name) const
 
-Lookup a bel by its name.
+Lookup a bel by its name, which is a list of IdStrings joined by `getNameDelimiter()`.
 
-### IdString getBelName(BelId bel) const
+### IdStringList getBelName(BelId bel) const
 
 Get the name for a bel. (Bel names must be unique.)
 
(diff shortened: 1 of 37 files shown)


docs/archapi.md
```diff
@@ -88,14 +88,18 @@ Get Z dimension for the specified tile for bels. All bels with at specified X an
 
 Get Z dimension for the specified tile for pips. All pips with at specified X and Y coordinates must have a Z coordinate in the range `0 .. getTileDimZ(X,Y)-1` (inclusive).
 
+### char getNameDelimiter() const
```
litghost, line 91: I have a suggestion here, but it does add complexity.  Have the `IdStringList` have a `DelimiterType` field.  A new archapi method could look something like:
```
const_range<IdString> getDelimiters(DelimiterType delimiters) const
```

This could enable different IdStringList's to have different presentations.  This would allow an arch to break down as needed, examples:
```
Pip: $tile/$wire0->>$wire1 ==> getDelimiters() => ["/", "->>"]
SitePip: $site.$site_type/$bel/$bel_pin ==> getDelimiters() => [".", "/", "/"]
```

This can get dangerous if misused, and I'm a little worried about complexity and / or IdStringList size bloat.  However I think it answers a TODO you had over here: https://github.com/YosysHQ/nextpnr/pull/561/files#diff-effe03ae96858aeff9455ed9406ec25b61ff5a52122abb6b3cb49b16f51b0dd0R314

daveshah1: I need to think about this; the bigger problem I think is the complexity that it would add to the parsing path.

litghost: Don't disagree about complexity.  But I do think the runtime cost can be managed.

daveshah1: So at the moment we can parse a string into an `IdStringList` with the only needed context being the current architecture. Doing it with variable delimiters would potentially require more context in the parsing, particularly if you want to support things like different delimiters inside different pip types.

litghost: A DFA (discrete finite automata) could be generated pretty easily to do the parsing without additional context.

daveshah1: I have thought about this a bit more, and while it's a nice idea given the feedback I have had about the Arch API being too complex already I think it is simply too much extra complexity for the benefit right now.

PR #562, opened 2021-01-29
Avoid linear scan in PIP check loop.

litghost:
The previous additions to archcheck increased the runtime of the nexus
archcheck quiet a bit.


common/archcheck.cc
```diff
@@ -136,11 +138,25 @@ void archcheck_conn(const Context *ctx)
         for (PipId pip : ctx->getPipsDownhill(wire)) {
             WireId wire2 = ctx->getPipSrcWire(pip);
             log_assert(wire == wire2);
+
+            auto iter = pips_downhill.find(pip);
+            if(iter != pips_downhill.end()) {
+                log_assert(iter->second == wire);
+            } else {
+                pips_downhill.insert(std::make_pair(pip, wire));
```
daveshah1, line 146: unless there's a subtle difference I'm missing, I think `emplace` is neater than `insert` and `make_pair`?

litghost: Fixed.

daveshah1 (approved):

PR #564, opened 2021-02-01
Add simple python file for doing Arch API sanity checks.

litghost:


python/check_arch_api.py
```diff
@@ -0,0 +1,54 @@
...
+    bel_pins_tested = 0
+    with open('test_data.yaml', 'r') as f:
+        test_data = yaml.safe_load(f.read())
+        if 'pip_test' in test_data:
+            for pip_test in test_data['pip_test']:
+                pip = None
+                for pip_name in ctx.getPipsDownhill(pip_test['src_wire']):
+                    if ctx.getPipDstWire(pip_name) == pip_test['dst_wire']:
+                        pip = pip_name
+                        assert ctx.getPipSrcWire(pip_name) == pip_test['src_wire']
```
daveshah1, line 40: it would be useful to print `ctx.getPipSrcWire(pip_name)` and `pip_test['src_wire']` in the assert, like in the bel pin test

litghost: Agreed, done.

PR #565, opened 2021-02-03
Use snake case consistently for non-Arch-API functions

daveshah1:
This improves the ability to distinguish between what is part of the official Arch API and what is arch-specific.

Longer term, there are potentially better ways to do this like https://github.com/SymbiFlow/nextpnr/pull/195#discussion_r566203422 (potentially with their own complexities) but this is a first step in that direction to improve the clarity.

An annoying exception here is the nextpnr-generic `add*` functions; which can't be renamed without a breaking change to downstream users of the Python API, which I try and keep stable without very good reasons for change.


daveshah1:
I am now working, on top of this PR, of making the Arch API use virtual functions and override in the [arch-override](https://github.com/YosysHQ/nextpnr/tree/dave/arch-override) branch

PR #567, opened 2021-02-03
Initial FPGA interchange arch

litghost:
This is the initial FPGA interchange arch implementation.  It is missing many important and required features, which has been listed in the README: https://github.com/YosysHQ/nextpnr/pull/567/files#diff-2bc12e83141b8893b3851e9bd7a07db94b0f7d807730b761cc91fb3c53078f76R36-R74

However the implementation can pass archcheck, and should contain a complete copy of the routing graph and all placable locations.


fpga_interchange/README.md
```diff
@@ -0,0 +1,162 @@
...
+ - [ ] Placement constraints are unimplemented, meaning invalid or unroutable
+       designs can be generated from the placer.
+ - [ ] Logical netlist macro expansion is not implemented, meaning that any
+       macro primitives are unplacable.  Common macro primitives examples are
+       differential IO buffers (IBUFDS) and some LUT RAM (e.g. RAM64X1D).
+ - [ ] Cell -> BEL pin mapping is not in place, meaning any primitives that
+       have different BEL pins with respect to their cell pins will not be
+       routable.
+ - [ ] Nextpnr only allows for cell -> BEL pin maps that are 1 to 1.  The
+       FPGA interchange accomidates cell -> BEL pin maps that include 1 to
```
daveshah1, line 48: nitpick - can you run a spelling check on this?

fpga_interchange/arch.h
```diff
@@ -0,0 +1,1308 @@
...
+        // FIXME: This probably needs to be fixed!
+        return false;
+    }
+
+    bool getBelHidden(BelId bel) const { return locInfo(bel).bel_data[bel.index].category != BEL_CATEGORY_LOGIC; }
+
+    IdString getBelType(BelId bel) const
+    {
+        NPNR_ASSERT(bel != BelId());
+        return IdString(locInfo(bel).bel_data[bel.index].type);
```
daveshah1, line 821: something I did in nexus, which improved the code quality a bit; was to have `bel_data`, etc, functions instead of using `locInfo` and indexing, which decouples the chip data accesses a bit. If iterators where given a `Arch` pointer instead of a direct database pointer; then I think they could use these same functions in a few places too.

the implementation in nexus actually has `bel_data` resolving to a static `chip_bel_data` because iterators there have database pointers but I don't think that was actually needed, retrospectively.

https://github.com/YosysHQ/nextpnr/blob/master/nexus/arch.h#L1432
https://github.com/YosysHQ/nextpnr/blob/master/nexus/arch.h#L314

fpga_interchange/arch.h
```diff
@@ -0,0 +1,1308 @@
...
+    }
+
+    std::vector<std::pair<IdString, std::string>> getBelAttrs(BelId bel) const;
+
+    int getBelPinIndex(BelId bel, IdString pin) const
```
daveshah1, line 826: can this follow the convention from #565 and use snake case for these non-Arch-API functions?

fpga_interchange/arch.h
```diff
@@ -0,0 +1,1308 @@
...
+    }
+
+    bool checkWireAvail(WireId wire) const
+    {
+        NPNR_ASSERT(wire != WireId());
+        auto w2n = wire_to_net.find(wire);
+        return w2n == wire_to_net.end() || w2n->second == nullptr;
+    }
+
+    NetInfo *getReservedWireNet(WireId wire) const
```
daveshah1, line 931: getReservedWireNet was a nextpnr-xilinx thing that isn't in upstream nextpnr and should be removed (router2 can actually infer most of the needed reservations in practice)

fpga_interchange/arch.cc
```diff
@@ -0,0 +1,577 @@
+/*
+ *  nextpnr -- Next Generation Place and Route
+ *
+ *  Copyright (C) 2018  Clifford Wolf <clifford@symbioticeda.com>
```
daveshah1, line 4: as you touch files, can you add your name to the copyright and also update claire's name?

fpga_interchange/arch.h
```diff
@@ -0,0 +1,1308 @@
...
+//     2. The second type is site pin edges, which connect site wires to tile
+//        wires (or vise-versa).
+
+NPNR_PACKED_STRUCT(struct BelInfoPOD {
+    int32_t name;       // bel name (in site) constid
+    int32_t type;       // Type name constid
+    int32_t bel_bucket; // BEL bucket constid.
+
+    int32_t num_bel_wires;
+    RelPtr<int32_t> ports; // port name constid
```
daveshah1, line 76: Would you like to switch these over to `RelSlice` as the other arches did in #553 ?

fpga_interchange/README.md
litghost: Done!

fpga_interchange/arch.cc
litghost: I believe I've fixed this, please take a look!

fpga_interchange/arch.h
litghost: Removed!  I thought I had cleaned up all of the nextpnr-xilinx specific methods, but missing this one.

fpga_interchange/arch.h
litghost: Fixed.

fpga_interchange/arch.h
litghost: Fixed.

fpga_interchange/arch.h
litghost: I'm working on this now, and it found some bugs!  Yay!

litghost: Done

daveshah1:
I missed the `constids.inc` change, and can't comment directly on it because of GitHub limitations.

But imo, including a 1MB, Xilinx-7-series-specific file, that creates 48k * 2 static variables probably isn't going to do wonders for build times nor the generic nature of the arch.

In general, going forwards, things only should go into `constids.inc` if you actually need to use them as identifiers in code rather than just in the database (see how nextpnr-xilinx and nextpnr-nexus both include extra IdStrings in the database only)

fpga_interchange/arch.h
```diff
@@ -0,0 +1,1275 @@
...
+    int cursor = -1;
+
+    void operator++()
+    {
+        cursor++;
+
+        while (twi != twi_end) {
+            WireId w = *twi;
+            auto &tile = tile_info(chip, w.tile);
+            if (cursor < tile.wire_data[w.index].bel_pins.size())
```
daveshah1, line 576: this is giving a signed/unsigned comparison warning (and in several other places too)

litghost:
> I missed the `constids.inc` change, and can't comment directly on it because of GitHub limitations.
> 
> But imo, including a 1MB, Xilinx-7-series-specific file, that creates 48k * 2 static variables probably isn't going to do wonders for build times nor the generic nature of the arch.
> 
> In general, going forwards, things only should go into `constids.inc` if you actually need to use them as identifiers in code rather than just in the database (see how nextpnr-xilinx and nextpnr-nexus both include extra IdStrings in the database only)

Committing that file was an error, I’ll fix it.

daveshah1:
> Committing that file was an error, I’ll fix it.

Thanks, can you add it and the nextpnr-fpga_interchange binary to the gitignore too?

Long term constids.inc probably isn't the best solution for this but we can cross that bridge later, I'm fine with it being used but not committed for the initial PR.

litghost:
> Thanks, can you add it and the nextpnr-fpga_interchange binary to the gitignore too?
> 
> Long term constids.inc probably isn't the best solution for this but we can cross that bridge later, I'm fine with it being used but not committed for the initial PR.

I'm working on simply removing it for now.  At this point I don't think I need it at all.  It is easy enough to add back.

litghost:
> Thanks, can you add it and the nextpnr-fpga_interchange binary to the gitignore too?

Fixed.  I'm working on the signed/unsigned errors next.

fpga_interchange/arch.h
litghost: Ya, fixed.

litghost:
Alright, I think this is back to a good state.

daveshah1 (approved):

daveshah1:
Rebasing the ArchAPI stuff on top of this; I expect you would prefer it to be based on the pure-virtual `ArchAPI` and kept separate from `BaseArch`?

litghost:
> Rebasing the ArchAPI stuff on top of this; I expect you would prefer it to be based on the pure-virtual `ArchAPI` and kept separate from `BaseArch`?

Yes please. After #568 I'll look at the state of things and see if further refactoring is warrented.

PR #568, opened 2021-02-04
Create a new BaseArch that formally specifies the Arch API and provides some base implementations 

daveshah1:
This creates a new `BaseArch` layer which `Arch` inherits from, and which implements the `Arch` functions virtually with a struct of `using`s as a template parameter for the functions where the return type is Arch-defined (as suggested in https://github.com/SymbiFlow/nextpnr/pull/195#discussion_r566203422 ).

Some further template magic is used to provide default implementations when the range types comply to certain traits or error out otherwise.

Things still to consider:
 - [ ] arches other than ECP5
 - [ ] more default implementations of functions
 - [ ] should BelId/WireId/PipId/etc also become template parameters to BaseArch, or will that make downstream code unwieldy

cc @ravenslofty @litghost

### DIFF
--- a/ecp5/arch.h
+++ b/ecp5/arch.h
@@ -437,6 +437,7 @@ NEXTPNR_NAMESPACE_BEGIN
 
 struct ArchRanges
 {
+    using ArchArgsType = ArchArgs;
     // Bels
     using AllBelsRange = BelRange;
     using TileBelsRange = BelRange;
@@ -490,8 +491,8 @@ struct Arch : BaseArch<ArchRanges>
     std::string getChipName() const override;
     std::string get_full_chip_name() const;
 
-    ArchArgs archArgs() const { return args; }
-    IdString archArgsToId(ArchArgs args) const;
+    ArchArgs archArgs() const override { return args; }
+    IdString archArgsToId(ArchArgs args) const override;
 
     // -------------------------------------------------
 
--- a/ecp5/arch.h
+++ b/ecp5/arch.h
@@ -457,6 +457,8 @@ struct ArchRanges
     using GroupWiresRange = std::vector<WireId>;
     using GroupPipsRange = std::vector<PipId>;
     using GroupGroupsRange = std::vector<GroupId>;
+    // Decals
+    using DecalGfxRange = std::vector<GraphicElement>;
     // Placement validity
     using CellTypeRange = const std::vector<IdString> &;
     using BelBucketRange = std::vector<BelBucketId>;
@@ -485,10 +487,9 @@ struct Arch : ArchBase<ArchRanges>
     static bool is_available(ArchArgs::ArchArgsTypes chip);
     static std::vector<std::string> get_supported_packages(ArchArgs::ArchArgsTypes chip);
 
-    std::string getChipName() const;
+    std::string getChipName() const override;
     std::string get_full_chip_name() const;
 
-    IdString archId() const { return id("ecp5"); }
     ArchArgs archArgs() const { return args; }
     IdString archArgsToId(ArchArgs args) const;
 
@@ -864,7 +865,7 @@ struct Arch : ArchBase<ArchRanges>
 
     // -------------------------------------------------
 
-    std::vector<GraphicElement> getDecalGraphics(DecalId decal) const;
+    std::vector<GraphicElement> getDecalGraphics(DecalId decal) const override;
 
     DecalXY getBelDecal(BelId bel) const override;
     DecalXY getWireDecal(WireId wire) const override;
--- a/ecp5/arch.h
+++ b/ecp5/arch.h
@@ -602,12 +602,6 @@ struct Arch : ArchBase<ArchRanges>
         return id;
     }
 
-    std::vector<std::pair<IdString, std::string>> getBelAttrs(BelId) const override
-    {
-        std::vector<std::pair<IdString, std::string>> ret;
-        return ret;
-    }
-
     WireId getBelPinWire(BelId bel, IdString pin) const override;
 
     BelPinRange getWireBelPins(WireId wire) const override
@@ -703,12 +697,6 @@ struct Arch : ArchBase<ArchRanges>
 
     PipId getPipByName(IdStringList name) const override;
     IdStringList getPipName(PipId pip) const override;
-
-    std::vector<std::pair<IdString, std::string>> getPipAttrs(PipId) const
-    {
-        std::vector<std::pair<IdString, std::string>> ret;
-        return ret;
-    }
 
     uint32_t getPipChecksum(PipId pip) const override { return pip.index; }
 
--- a/ecp5/arch.h
+++ b/ecp5/arch.h
@@ -472,8 +472,6 @@ struct Arch : ArchBase<ArchRanges>
     mutable std::unordered_map<IdStringList, PipId> pip_by_name;
 
     std::vector<CellInfo *> bel_to_cell;
-    std::unordered_map<WireId, NetInfo *> wire_to_net;
-    std::unordered_map<PipId, NetInfo *> pip_to_net;
     std::unordered_map<WireId, int> wire_fanout;
 
     // fast access to  X and Y IdStrings for building object names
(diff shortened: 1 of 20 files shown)


litghost:
I strongly recommend that you make BaseArch a pure virtual interface (e.g. no concrete implementations).  The reason for this is to have the `BaseArch` soley be an interface, rather than an interface and a partial implementation.  Where there are commonly re-used methods, move those into small classes with methods.  Subclasses of the BaseArch can then re-use those components by invoking them.

I can already see at least one case where this change is valuable as an example.  `bindBel` is a method that all arches have a relatively common method, which you have put as a member method of the BaseArch: https://github.com/YosysHQ/nextpnr/pull/568/files#diff-413fc08cf5a101d0a7a508ca4adb7df3d69bfd5d5095af89d595ede2da7c7432R1085-R1094

For arches that don't need to change it, they can use this method as is.  But for arches that need to modify the method (e.g. an incremental placement checker), they cannot re-use any of logic.  The concrete suggestion here is to move `base_bel2cell` to a new class, like `BasicCellInfo`: 

```
struct BasicCellInfo {
  void unbindBel(BelId bel);
  void bindBel(BelId bel, CellInfo *cell, PlaceStrength strength);
  CellInfo *getBoundBelCell(BelId bel) const;

  std::unordered_map<BelId, CellInfo *> bel2cell;
};
```

For arches that just want to use the `BasicCellInfo`, then can include this class, and just call it in `unbindBel`, `bindBel`, `getBoundBelCell`.  For arches that want `BasicCellInfo` plus a little more, then can still re-use the `BasicCellInfo` implementation, but add as needed.

Does that make sense?

daveshah1:
> Where there are commonly re-used methods, move those into small classes with methods. Subclasses of the BaseArch can then re-use those components by composing them into their classes and invoking them.

The downside of this is it increases the overall amount of per-arch code, which based on feedback from several people is something that I want to minimise as much as possible going forward

> For arches that don't need to change it, they can use this method as is. But for arches that need to modify the method (e.g. an incremental placement checker), they cannot re-use any of logic. The concrete suggestion here is to move base_bel2cell data (and its methods) to a new class, like BasicCellInfo:

I think I'm missing how this would differ from something like this?
https://github.com/YosysHQ/nextpnr/blob/f73061b60e6501dea35785e8b1267cbda6c8b799/ecp5/arch.h#L709-L713

Ravenslofty:
Requiring an arch to use a lot of other small classes to do tasks does not seem significantly better than just copy/pasting the same code from ECP5 or wherever.

I suggested something along the lines of a partial implementation to Dave, borrowing from Rust where a trait can require some methods, but implement others on top of that.

Methods should be provided but overridable, because this minimises the time and effort it takes to bring up an architecture, which is one of the biggest stumbling blocks that nextpnr has at the moment. The fastest method to write is the one you don't have to write.

litghost:
> I think I'm missing how this would differ from something like this?

The difference has to do with scale and scope.  The size of the code and the amount of state that is in something like `BasicCellInfo` is very small.  So `basic_cell_info->bindBel(bel)` can only possible touch a small amount of code and state.  Where-as `BaseArch::bindBel(bel)` could potentially affect anything in `BaseArch`.  `BaseArch` in its current form has 8 state variables and dozens of methods.  `BasicCellInfo` only has 1 state variable and 3 methods.

This is partially an argument for having smaller domains of state and code reach.  `Arch` is a bit of a god class, and creating smaller state/code domains is one way to shrink that scale.

> Methods should be provided but overridable, because this minimises the time and effort it takes to bring up an architecture, which is one of the biggest stumbling blocks that nextpnr has at the moment. The fastest method to write is the one you don't have to write.

I don't disagree with what you are saying, but I want to mention that you are focusing on making nextpnr faster to write new arches in.  I don't think that is necessarily the only optimization target we want to aim for.  Optimizing for maintainability and lack of surprises are also valid choices.  By making smaller components with smaller loci, I believe we can also be optimizing for those other properties.

I think there is something to be said for creating a smaller Arch API for architectures that need to implement less.  I have some ideas on that front, but they are out of scope.

litghost:
At a minimum, I believe that there should be a pure virtual arch API that exists soley to enforce the Arch API specification, and if it is desired to have a partially implemented base class for use in ecp5/nexus, that is a valid choice.  But it should be opt in by arch developers, rather than required.

daveshah1:
> I think there is something to be said for creating a smaller Arch API for architectures that need to implement less. I have some ideas on that front, but they are out of scope for this particular discussion

An important issue here is that arches often want to start small and incrementally use more functionality, so having two different APIs wouldn't work. This is why Lofty and I settled on the overriding approach.

> At a minimum, I believe that there should be a pure virtual arch API that exists soley to enforce the Arch API specification, and if it is desired to have a partially implemented base class for use in ecp5/nexus, that is a valid choice. But it should be opt in by arch developers, rather than required.

That seems reasonable, it would add another layer of complexity and potential confusion though.

daveshah1:
In terms of the maintenability side, I think minimising the amount of per-arch code as far as possible will have at least as many benefits as compartmentalisation would.

Ravenslofty:
Not to mention that because the compartmentalised classes increase the number of lines of code required to implement an architecture, trying to use that approach means there will be less code to maintain.

Because nobody will want to write an architecture.

litghost:
> > I think there is something to be said for creating a smaller Arch API for architectures that need to implement less. I have some ideas on that front, but they are out of scope for this particular discussion
> 
> An important issue here is that arches often want to start small and incrementally use more functionality, so having two different APIs wouldn't work. This is why Lofty and I settled on the overriding approach.

I think the start small and incrementally using more functionality is more addressed with using composable components.  You start with the "starter" kit of components, and then remove them as the Arch out grows them.  By having small composable components there are natural steps to growing by replacing those components.  It is also very clear what those components are doing, which aids the reader in what these parts of doing.

I guess my only other comment here is that nextpnr is not the first project to try to use inheritance in this way. Projects that have tried this over time find that the inheritance based models have generally proven to be problematic and they generally move to composition.  I cannot prove this as a natural law, but it a trend.

My request is to have the pure virtual base class, and have the partially implemented version as opt-in.

daveshah1:
> I think the start small and incrementally using more functionality is more addressed with using composable components. You start with the "starter" kit of components, and then remove them as the Arch out grows them. By having small composable components there are natural steps to growing by replacing those components. It is also very clear what those components are doing, which aids the reader in what these parts of doing.

I do agree with that in general, and would probably take a more composition-based approach to nextpnr if starting from scratch. But in this particular case, I don't think it really solves much beyond making all the arches bigger. At the end of the day the default implementation is not very large - only a couple of hundred lines most of which is stubs returning a default value - so splitting it up doesn't add much value imo.

Ravenslofty:
The Arch API is huge. I think everybody agrees with that. And of course, if there was a way to fragment the Arch API so that it was more composable, then I think everybody would agree with that too.

But if you're going to redesign the architecture API, then you need to redesign the existing architectures to fit that new API, along with the core algorithms, too.

At that point you have something which doesn't look very much like nextpnr, and if you're going to invest the effort of redesigning the API, you might as well rewrite nextpnr; I'm aware that Dave dislikes much of the design of the program.

So let's see this for what it is: a plaster over some very questionable historical design decisions, trying to prolong the life of the nextpnr codebase. Let's not spend days seeking perfection in something that does not have a clear future.

daveshah1:
> My request is to have the pure virtual base class, and have the partially implemented version as opt-in.

This has now been done

ecp5/arch.h
```diff
@@ -459,48 +487,47 @@ struct Arch : BaseCtx
     static bool is_available(ArchArgs::ArchArgsTypes chip);
     static std::vector<std::string> get_supported_packages(ArchArgs::ArchArgsTypes chip);
 
-    std::string getChipName() const;
+    std::string getChipName() const override;
     std::string get_full_chip_name() const;
 
-    IdString archId() const { return id("ecp5"); }
     ArchArgs archArgs() const { return args; }
```
litghost, line 493: Nit: If these aren't override methods, should be `snake_case`?

daveshah1: Actually looking through how these are used, they should become Arch API methods so I will add them as such

litghost: I don't think I need to say this, but please update `archapi.md` if these are not already there.

litghost:
Please make the `~BaseCtx` virtual (https://github.com/YosysHQ/nextpnr/pull/568/files#diff-413fc08cf5a101d0a7a508ca4adb7df3d69bfd5d5095af89d595ede2da7c7432R861), see https://www.geeksforgeeks.org/virtual-destructor/

Basically, if the base class destructor isn't virtual, surprising and dumb things can happen.  Best to just avoid that case, even if the current code doesn't trigger it.

litghost:
Other than `~BaseCtx` virtual, I think the rest LGTM.

gatecat:
(new username, in case you are surprised!)

That's done now, I'll merge once the CI is green. 

There are a few more things I'd like to do like the option for a shared `DelayInfo` implementation, but this PR is big enough already so those can be for another day.

litghost:
Agreed, smaller PRs are easier to review!

PR #570, opened 2021-02-04
Mark IdString and IdStringList single argument constructors explicit.

litghost:
Non-explicit single argument constructors will silently convert to that type.  This
is typically not the right thing to do.  For example, the nexus and
ice40 arch_pybindings.h files were incorrectly parsing bel name strings,
etc.

I discovered this issue when checking the python API.

### DIFF
--- a/common/archcheck.cc
+++ b/common/archcheck.cc
@@ -135,8 +135,10 @@ void archcheck_locs(const Context *ctx)
 //
 // This allows a fast way to check getPipsDownhill/getPipsUphill from getPips,
 // without balloning memory usage.
-struct LruWireCacheMap {
-    LruWireCacheMap(const Context *ctx, size_t cache_size) : ctx(ctx), cache_size(cache_size) {
+struct LruWireCacheMap
+{
+    LruWireCacheMap(const Context *ctx, size_t cache_size) : ctx(ctx), cache_size(cache_size)
+    {
         cache_hits = 0;
         cache_misses = 0;
         cache_evictions = 0;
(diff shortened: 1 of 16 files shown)


daveshah1:
Oops, good catch!

litghost:
I'm not totally certain about the right thing to do for the generic arch change.

daveshah1:
Pushed an alternate fix for the generic arch - that was my error in the original patch. Will merge once the CI with that change goes green.

pepijndevos:
Apparently @trabucayre found a regression in the Gowin backend and bisected it back to this PR.

> terminate called after throwing an instance of 'nextpnr_gowin::assertion_failure' what():  Assertion failure: pips.count(name) == 0 (/home/gwe/misc/icestick/nextpnr/gowin/arch.cc:74)
> if I display name I have R1C1_N240_CLK0 twice
> I like git bisect. Regression is in c99fbde0eb0b1b9b725ba2fead13d3210ce961a7

gatecat:
I am looking into it (I suspect it might be more down to the general IdString changes than this specific patch)

litghost:
> I am looking into it (I suspect it might be more down to the general IdString changes than this specific patch)

Thanks, I was going to take a look in a bit.

@pepijndevos / @trabucayre Do you have a particular invocation that fails?  In case @gatecat doesn't find it by inspection.

gatecat:
I just found it via archcheck (which I've also added to CI so this doesn't happen again!) and fixed in #588

PR #575, opened 2021-02-10
Support for cell pin to bel pin mappings

### LINKED ISSUE #560: Remove the direct mapping between cell pins and bel pins
Currently, there is an assumption of 1:1 naming and cardinality between cell pins and bel pins. In the early days of nextpnr the `CellInfo::pins` structure was designed to allow mapping rules between cell pin name and bel pin names; but this doesn't solve the problem of 1:N cardinality (needed for Xilinx BRAM write enables, for example) and has never been used 'in anger' even though a partial implementation does exist.

@litghost 's proposal boils down to adding an Arch API function that returns a list of bel pins that a cell pin maps to:

```
const_range<BelPinId> getBelPinsForCellPin(CellInfo *cell_info, IdString s);
```

at the same time, using `BelPinId` rather than `IdString` to index bel pins to allow faster lookups of bel pin wires. It is notable that currently the entirety of nextpnr assumes 1:1 cardinality between cell/bel pins (and moreover between net sinks and bel pins) so some significant refactoring of place, route and timing would be needed to implement this correctly. This would include changing the `getNetinfoSinkWire` API to return or otherwise allow iteration over a list of wires. It has also been suggested that this could be used to implement things like LUT permutation without requiring extra PIPs for this, by returning a list of wires but only requiring that a subset of those wires are actually routed to:

```
struct NetSinks {
  const_range<WireId> candidate_sink_wires;
  size_t required_sinks;
};
```

An alternative proposal would be to maintain a 1:1 cardinality between cell and bel pins; and have a mapping back from physical cell pins to logical cell pins (which could be a N:1 mapping). This would reduce the amount of complexity being pushed onto the router and keep the netlist structure closer to its current form; but is less clean in other ways and wouldn't extend as well to the LUT permutation proposal above.

gatecat:
This adds a new Arch API to request the range of bel pin(s) that correspond to a cell pin and provides a trivial implementation for all arches. It is primarily a RFC to work out the kinds of changes that the router etc would need to use such an API at this stage.

Outstanding issues: 
 - [ ] actually find a use case for multiple cell->bel pin mappings and test (possibly Nexus BRAM to start with)
 - [ ] should we create a new Id type for bel pins instead of using `IdString? Profiling suggests that the time to lookup bel pin to wire is <1% of runtime and there is plenty of other low hanging fruit - but this would improve type safety as it would be impossible to accidentally use a cell pin as a bel pin without going through the new API. Downside is increased complexity.
 - [ ] should we have a "N of M" API as proposed at https://github.com/SymbiFlow/nextpnr/pull/195#discussion_r566258705 for LUT permuation, etc, or is there an alternative way of dealing with this that doesn't add complexity to paths that don't need it?

Fixes #560

### DIFF
--- a/common/nextpnr.cc
+++ b/common/nextpnr.cc
@@ -354,19 +354,59 @@ WireId Context::getNetinfoSourceWire(const NetInfo *net_info) const
     if (src_bel == BelId())
         return WireId();
 
-    IdString driver_port = net_info->driver.port;
-    return getBelPinWire(src_bel, driver_port);
+    auto bel_pins = getBelPinsForCellPin(net_info->driver.cell, net_info->driver.port);
+    auto iter = bel_pins.begin();
+    if (iter == bel_pins.end())
+        return WireId();
+    WireId driver = getBelPinWire(src_bel, *iter);
+    ++iter;
+    NPNR_ASSERT(iter == bel_pins.end()); // assert there is only one driver bel pin;
+    return driver;
 }
 
-WireId Context::getNetinfoSinkWire(const NetInfo *net_info, const PortRef &user_info) const
+SSOArray<WireId, 2> Context::getNetinfoSinkWires(const NetInfo *net_info, const PortRef &user_info) const
 {
     auto dst_bel = user_info.cell->bel;
-
     if (dst_bel == BelId())
-        return WireId();
+        return SSOArray<WireId, 2>(0, WireId());
+    size_t bel_pin_count = 0;
+    // We use an SSOArray here because it avoids any heap allocation for the 99.9% case of 1 or 2 sink wires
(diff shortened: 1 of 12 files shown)


litghost:
With the removal of the `pins` field from `CellInfo*` in #574, #576 is now blocked on this PR (or equivalent).  What work remains in this PR?

gatecat:
Work that remains:
 - decide if we are OK with keeping IdString as the identifier for bel pins for now
 - make sure there hasn't been a significant runtime regression from the extra overhead 
 - add support for setting up pin mappings in nextpnr-generic (but this could be a separate PR)

litghost:
> decide if we are OK with keeping IdString as the identifier for bel pins for now

Let's leave the for future development.  If there is a large performance need to change it, we can cross that bridge then.  This is definitely a "make it right, then make it fast" case.

PR #576, opened 2021-02-11
Complete FPGA interchange Arch to the point where it can route a wire

litghost:
This PR is too big for review/merge, but it is a WIP branch with a first cut at something that can:
 - Read a FPGA interchange logical netlist
 - Constrain IO ports via XDC
 - Route from an IBUF to an OBUF
 - Write a FPGA interchange physical netlist

I have the nextpnr portion of the flow complete, but I need to finish debugging the writing of the physical netlist.  After I get something that works (e.g. can route a wire, etc), I'll need to refactor this PR into smaller portions.

### DIFF
--- a/common/exclusive_state_groups.h
+++ b/common/exclusive_state_groups.h
@@ -24,6 +24,8 @@
 #error Include after "nextpnr.h" only.
 #endif
 
+#include "bits.h"
+
 NEXTPNR_NAMESPACE_BEGIN
 
 // Implementation for exclusive state groups, used to implement generic
@@ -103,10 +105,10 @@ template <size_t StateCount, typename StateType = int8_t, typename CountType = u
 
             // Check whether the state group is now unselected or satified.
             auto value = selected_states.to_ulong();
-            auto number_selected = __builtin_popcount(value);
+            auto number_selected = nextpnr::Bits::popcount(value);
             if (number_selected == 1) {
                 // Group is no longer overconstrained.
-                state = __builtin_ctz(value);
+                state = nextpnr::Bits::ctz(value);
                 NPNR_ASSERT(selected_states[state]);
             } else if (number_selected == 0) {
                 // Group is unselected.
--- a/common/exclusive_state_groups.h
+++ b/common/exclusive_state_groups.h
@@ -1,0 +1,146 @@
+/*
+ *  nextpnr -- Next Generation Place and Route
+ *
+ *  Copyright (C) 2021  The SymbiFlow Authors.
+ *
+ *  Permission to use, copy, modify, and/or distribute this software for any
+ *  purpose with or without fee is hereby granted, provided that the above
+ *  copyright notice and this permission notice appear in all copies.
+ *
+ *  THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ *  WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ *  MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ *  ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ *  WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ *  ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ *  OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ *
+ */
+
+#ifndef EXCLUSIVE_STATE_GROUPS_H
+#define EXCLUSIVE_STATE_GROUPS_H
+
+#ifndef NEXTPNR_H
+#error Include after "nextpnr.h" only.
+#endif
+
+NEXTPNR_NAMESPACE_BEGIN
+
+// Implementation for exclusive state groups, used to implement generic
+// constraint system.
+template <size_t StateCount, typename StateType = int8_t, typename CountType = uint8_t> struct ExclusiveStateGroup
+{
+    ExclusiveStateGroup() : state(kNoSelected) { count.fill(0); }
+    struct Definition
+    {
+        IdString prefix;
+        IdString default_state;
+        std::vector<IdString> states;
+    };
+
+    static_assert(StateCount < std::numeric_limits<StateType>::max(), "StateType cannot store max StateType");
+    static_assert(std::numeric_limits<StateType>::is_signed, "StateType must be signed");
+
+    std::bitset<StateCount> selected_states;
+    StateType state;
+    std::array<CountType, StateCount> count;
+
+    static constexpr StateType kNoSelected = -1;
+    static constexpr StateType kOverConstrained = -2;
+
+    std::pair<bool, IdString> current_state(const Definition &definition) const
+    {
+        if (state <= 0) {
+            return std::make_pair(state == kNoSelected, definition.default_state);
+        } else {
+            NPNR_ASSERT(state <= definition.states.size());
+            return std::make_pair(true, definition.states[state]);
+        }
+    }
+
+    bool check_implies(int32_t next_state) const
+    {
+        // Implies can be satified if either that state is
+        // selected, or no state is currently selected.
+        return state == next_state || state == kNoSelected;
+    }
+
+    bool add_implies(int32_t next_state)
+    {
+        NPNR_ASSERT(next_state < StateCount);
+
+        // Increment and mark the state as selected.
+        count[next_state] += 1;
+        selected_states[next_state] = true;
+
+        if (state == next_state) {
+            // State was already selected, state group is still satified.
+            return true;
+        } else if (selected_states.count() == 1) {
+            // State was not select selected, state is now selected.
+            // State group is satified.
+            state = next_state;
+            return true;
+        } else {
+            // State group is now overconstrained.
+            state = kOverConstrained;
+            return false;
+        }
+    };
+
+    void remove_implies(int32_t next_state)
+    {
+        NPNR_ASSERT(next_state < StateCount);
+        NPNR_ASSERT(selected_states[next_state]);
+
+        count[next_state] -= 1;
+        NPNR_ASSERT(count[next_state] >= 0);
+
+        // Check if next_state is now unselected.
+        if (count[next_state] == 0) {
+            // next_state is not longer selected
+            selected_states[next_state] = false;
+
+            // Check whether the state group is now unselected or satified.
+            auto value = selected_states.to_ulong();
+            auto number_selected = __builtin_popcount(value);
+            if (number_selected == 1) {
+                // Group is no longer overconstrained.
+                state = __builtin_ctz(value);
+                NPNR_ASSERT(selected_states[state]);
+            } else if (number_selected == 0) {
+                // Group is unselected.
+                state = kNoSelected;
+            } else {
+                state = kOverConstrained;
+            }
+        }
+    }
+
+    template <typename StateRange> bool requires(const StateRange &state_range) const
+    {
+        if (state < 0) {
+            return false;
+        }
+
+        for (const auto required_state : state_range) {
+            if (state == required_state) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    void print_debug(const Context *ctx, IdString object, const Definition &definition) const;
+    void explain_implies(const Context *ctx, IdString object, IdString cell, const Definition &definition, BelId bel,
+                         int32_t next_state) const;
+
+    template <typename StateRange>
+    void explain_requires(const Context *ctx, IdString object, IdString cell, const Definition &definition, BelId bel,
+                          const StateRange state_range) const;
+};
+
+NEXTPNR_NAMESPACE_END
+
+#endif
--- a/fpga_interchange/arch.cc
+++ b/fpga_interchange/arch.cc
@@ -828,6 +828,18 @@ bool Arch::is_net_within_site(const NetInfo &net) const
     return true;
 }
 
+size_t Arch::get_cell_type_index(IdString cell_type) const
+{
+    const CellMapPOD &cell_map = *chip_info->cell_map;
+    int cell_offset = cell_type.index - cell_map.cell_names[0];
+    if((cell_offset < 0 || cell_offset >= cell_map.cell_names.ssize())) {
+        log_error("Cell %s is not a placable element.\n", cell_type.c_str(this));
+    }
+    NPNR_ASSERT(cell_map.cell_names[cell_offset] == cell_type.index);
+
+    return cell_offset;
+}
+
 // Instance constraint templates.
 template void Arch::ArchConstraints::bindBel(Arch::ArchConstraints::TagState *, const Arch::ConstraintRange);
 template void Arch::ArchConstraints::unbindBel(Arch::ArchConstraints::TagState *, const Arch::ConstraintRange);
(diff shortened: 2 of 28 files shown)


common/design_utils.cc
```diff
@@ -71,7 +71,9 @@ void print_utilisation(const Context *ctx)
     }
     std::map<IdString, int> available_types;
     for (auto bel : ctx->getBels()) {
-        available_types[ctx->getBelType(bel)]++;
+        if(!ctx->getBelHidden(bel)) {
```
gatecat, line 74: This will need to be added to the Arch API (but I think it's a good thing to add.)

common/exclusive_state_groups.h
```diff
@@ -0,0 +1,146 @@
...
+        NPNR_ASSERT(count[next_state] >= 0);
+
+        // Check if next_state is now unselected.
+        if (count[next_state] == 0) {
+            // next_state is not longer selected
+            selected_states[next_state] = false;
+
+            // Check whether the state group is now unselected or satified.
+            auto value = selected_states.to_ulong();
+            auto number_selected = __builtin_popcount(value);
```
gatecat, line 106: I suspect these are going to need fallbacks in order to support MSVC

fpga_interchange/arch.cc
```diff
@@ -573,4 +693,95 @@ const std::vector<std::string> Arch::availablePlacers = {"sa",
...
+    cell->cell_bel_pins.clear();
+
+    const CellBelMapPOD &cell_pin_map = chip_info->cell_map->cell_bel_map[mapping];
+
+    log_info("Mapping cell %s\n", cell->name.c_str(this));
+    for (const auto &pin_map : cell_pin_map.common_pins) {
+        IdString cell_pin(pin_map.cell_pin);
+        IdString bel_pin(pin_map.bel_pin);
+
+        if(cell_pin.str(this) == "GND") {
```
gatecat, line 709: You might want to think about using `constid`s like other arches (mixing fixed constids and db ones like nexus and nextpnr-xilinx) rather than string compares; this _probably_ isn't a very hot path but I suspect you will hit other paths where fast IdString compares/lookups are needed

fpga_interchange/arch.cc
```diff
@@ -573,4 +693,95 @@ const std::vector<std::string> Arch::availablePlacers = {"sa",
...
+            // FIXME: Tie this pin to the VCC net
+            continue;
+        }
+
+        auto &bel_pins = cell->cell_bel_pins[cell_pin];
+        log_info("%s => %s\n", cell_pin.c_str(this), bel_pin.c_str(this));
+        cell->pins[cell_pin] = bel_pin;
+        bel_pins.push_back(bel_pin);
+
+        // FIXME: Only have support for 1-to-1 pin in the router right now.
```
gatecat, line 723: in case you haven't seen it, #575

fpga_interchange/arch.cc
```diff
@@ -573,4 +693,95 @@ const std::vector<std::string> Arch::availablePlacers = {"sa",
...
+}
+
+std::string Arch::get_part() const {
+    // FIXME: Need a map between device / package / speed grade and part.
+    return chip_info->name.get() + args.package + "-1";
+}
+
+void Arch::map_port_pins(BelId bel, CellInfo *cell) const {
+    IdStringRange pins = getBelPins(bel);
+    NPNR_ASSERT(pins.begin() != pins.end());
```
gatecat, line 770: I think I've seen a few places where you are getting exactly one item from an iterator. It would be neater to have a generic function (e.g. in `util.h`) with the various asserts for this purpose

fpga_interchange/arch.h
```diff
@@ -696,12 +752,14 @@ struct Arch : ArchAPI<ArchRanges>
...
 
+    static constexpr size_t kMaxState = 8;
     struct TileStatus
     {
+        std::vector<ExclusiveStateGroup<kMaxState>> tags;
         std::vector<CellInfo *> boundcells;
     };
 
-    std::vector<TileStatus> tileStatus;
+    std::unordered_map<int32_t, TileStatus> tileStatus;
```
gatecat, line 762: I'm kind of curious of the performance tradeoff here

fpga_interchange/arch.h
```diff
@@ -758,39 +818,102 @@ struct Arch : ArchAPI<ArchRanges>
...
+            constraints.unbindBel(tile_status.tags.data(), get_cell_constraints(bel, cell->type));
+        }
+
         refreshUiBel(bel);
     }
 
-    bool checkBelAvail(BelId bel) const override { return tileStatus[bel.tile].boundcells[bel.index] == nullptr; }
+    bool checkBelAvail(BelId bel) const override
+    {
+        // FIXME: This could consult the constraint system to see if this BEL
```
gatecat, line 895: this should probably wait until we've had a more general think about placement validity - use of `checkBelAvail` and `getConflictingBelCell` for validity reasons isn't really something we've done before but does have some interesting applications, though would need some placer work to take advantage of

frontend/frontend_base.h
```diff
@@ -583,29 +583,16 @@ template <typename FrontendType> struct GenericFrontend
...
-            auto drv = net->driver;
-            if (drv.cell != nullptr) {
-                disconnect_port(ctx, drv.cell, drv.port);
-                drv.cell->ports[drv.port].net = nullptr;
-                connect_port(ctx, split_iobuf_i, drv.cell, drv.port);
-            }
-            connect_port(ctx, split_iobuf_i, iobuf, ctx->id("I"));
-            NPNR_ASSERT(net->driver.cell == nullptr);
-            connect_port(ctx, net, iobuf, ctx->id("O"));
+            iobuf->addInout(ctx->id("IO"));
```
gatecat, line 586: This will break other arches; I do think the whole IO buffer stuff is not the nicest part of nextpnr and is somewhat a legacy of the decision to support inferring IO buffers (including complex tristate stuff) in the ice40 and ecp5 arches. If you have any suggestions on improving this; I'd be interested to know (one option is an arch flag to disable the creation of the buffers altogether for arches that don't need them; then they can just look at `ctx->ports`)

fpga_interchange/xdc.cc
```diff
@@ -0,0 +1,160 @@
...
+    PortInfo * port_info = static_cast<PortInfo*>(object->internalRep.twoPtrValue.ptr2);
+    NPNR_ASSERT(port_info->net != nullptr);
+    CellInfo * cell = ctx->port_cells.at(port_info->name);
+
+    cell->attrs[ctx->id(property)] = Property(value);
+
+    return TCL_OK;
+}
+
+void Arch::parse_xdc(const std::string &filename) {
```
gatecat, line 141: As a bit of a note; a proper xdc implementation will need to override the default command handler to return "number" commands as themselves in square brackets; so that bus ports work without needing an escaped string (e.g. command `3` returns `[3]` so you can do `foo[3]`). It turns out the rules Vivado use for this are actually [quite interesting](https://twitter.com/whitequark/status/1256508818554597377)...

gatecat:
In general I'm happy with this; I suspect most of my review comments are things you know already but don't want them being forgotten.

Is there a need to check in the capnproto output; if we require the capnp library to link against then requiring the capnp compiler doesn't seem to onerous an extra requirement?

fpga_interchange/xdc.cc
litghost: Good catch, I'll fix that.

litghost:
> Is there a need to check in the capnproto output; if we require the capnp library to link against then requiring the capnp compiler doesn't seem to onerous an extra requirement?

We don't (and probably shouldn't) check in the capnp generated files.  My goal with this WIP PR is to get to a full end to end flow ASAP, so I could discover any surprises.  Once I get a working wire physical netlist output, I'll circle back and incrementally fix issues.

frontend/frontend_base.h
litghost: The problem as I see it right now is that nextpnr doesn't have a place to attach `port` properties, e.g. `PACKAGE_PIN`, `IOSTANDARD` (or equivalent).  The iobuf cells that are created here are a reasonable place for now.

I'll think on this today.  In general, we want the placer to be able to choose where IO ports and their buffers are in the fabric, and that does mean a placable object must exist.  I think that inout top level ports should be represented by 1 wire that connects to the tristate buffer and local pad objects (e.g. pulls).

As a short term fix, what do you think about gating the `$nextpnr_iobuf` behavior (e.g. 1 port named "IO" versus two ports named "I" and "O") behind an argument to `GenericFrontend`?

gatecat: > As a short term fix, what do you think about gating the $nextpnr_iobuf behavior (e.g. 1 port named "IO" versus two ports named "I" and "O") behind an argument to GenericFrontend?

SGTM

fpga_interchange/arch.h
litghost: I believe for smaller designs, this will be good. I think for large fabrics, this will need to be revisited.  Because of how I moved access to the `tileStatus` behind `get_tile_status` for all of the mutable cases, I believe this will be easy to change as needed once we have more test data.

fpga_interchange/arch.cc
litghost: Sure.

fpga_interchange/arch.cc
litghost: Right.  For now, it is probably best that I just debug 1-to-1 mappings (e.g. LUT's, FF's, LUT-RAM's, CARRY's). Just getting to the point where I can place and route those items, this arch will be in a good starting place!

gatecat: Yes that's sensible

gatecat:
> We don't (and probably shouldn't) check in the capnp generated files

I'm seeing things like ` fpga_interchange/References.capnp.c++` in this PR? Either way sounds like it's a temporary issue only.

common/design_utils.cc
litghost: This was done in #581

fpga_interchange/arch.cc
litghost: This was handled in #586

frontend/frontend_base.h
litghost: This was fixed in #579

fpga_interchange/xdc.cc
litghost: This was addressed in #582

fpga_interchange/arch.cc
litghost: Done!

common/exclusive_state_groups.h
litghost: Yep, I've opened a PR to address this here: https://github.com/YosysHQ/nextpnr/pull/589

litghost:
> > We don't (and probably shouldn't) check in the capnp generated files
> 
> I'm seeing things like ` fpga_interchange/References.capnp.c++` in this PR? Either way sounds like it's a temporary issue only.

Yep, this was addressed as part of #583

litghost:
I believe this is ready for review now.  I believe all of the stuff that can be easily separated has been. The two examples appear to work.  I'm moving onto constant network support, as that is the next step missing for something like a murax or picosoc class design (without LUT-RAM/CARRY).

fpga_interchange/site_router.cc
```diff
@@ -0,0 +1,753 @@
...
+                    node = node->parent;
+                } while (node != RouteNode::Node());
+
+                if (uncongestion_route != RouteNode::Node()) {
+                    break;
+                }
+            }
+
+            if (uncongestion_route != RouteNode::Node()) {
+                // Select a trivally uncongestion route if possible.
```
gatecat, line 638: nitpick - these comments could do with a spelling check

fpga_interchange/site_router.cc
```diff
@@ -0,0 +1,753 @@
...
+        return site_ok;
+    }
+
+    site_ok = route_site(ctx, &site_info);
+    if (verbose_site_router(ctx)) {
+        if (site_ok) {
+            site_info.remove_routed_sources();
+            NPNR_ASSERT(site_info.is_fully_routed());
+            log_info("Site %s is routable\n",
+                     ctx->chip_info->sites[ctx->chip_info->tiles[tile].sites[site]].name.get());
```
gatecat, line 743: I think a `get_site_name(tile, site)` helper function might be a good idea here

fpga_interchange/arch.h
```diff
@@ -818,12 +844,39 @@ struct Arch : ArchAPI<ArchRanges>
...
+    TileStatus &get_tile_status(int32_t tile)
+    {
+        auto result = tileStatus.emplace(tile, TileStatus());
+        if (result.second) {
+            auto &tile_type = chip_info->tile_types[chip_info->tiles[tile].type];
+            result.first->second.boundcells.resize(tile_type.bel_data.size());
+            result.first->second.tags.resize(default_tags.size());
+
+            result.first->second.sites.reserve(tile_type.number_sites);
+            for (size_t i = 0; i < tile_type.number_sites; ++i) {
```
gatecat, line 856: this is giving a mixed signed/unsigned warning

gatecat:
My attempt at testing this locally is failing during the patch step:

```
Traceback (most recent call last):
  File "/usr/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/david/ic_test/python-fpga-interchange/fpga_interchange/patch.py", line 72, in <module>
    main()
  File "/home/david/ic_test/python-fpga-interchange/fpga_interchange/patch.py", line 56, in main
    root_schema = get_schema(args.schema_dir, args.schema,
  File "/home/david/ic_test/python-fpga-interchange/fpga_interchange/convert.py", line 120, in get_schema
    schemas = Interchange(schema_dir)
  File "/home/david/ic_test/python-fpga-interchange/fpga_interchange/interchange_capnp.py", line 822, in __init__
    self.references_schema = capnp.load(
  File "capnp/lib/capnp.pyx", line 4034, in capnp.lib.capnp.load
  File "capnp/lib/capnp.pyx", line 3297, in capnp.lib.capnp.SchemaParser.load
capnp.lib.capnp.KjException: home/david/ic_test/RapidWright/interchange/References.capnp:15: failed: Import failed: /capnp/java.capnp
```

is this something you have seen before?

litghost:
> My attempt at testing this locally is failing during the patch step:
>
> is this something you have seen before?

Ya this is a frustrating thing.  If you compile and install https://github.com/capnproto/capnproto-java this error will be resolved.  I put in a hot patch for this when compiling the C++ capnp's (see here https://github.com/SymbiFlow/fpga-interchange-schema/blob/bce42208b80c239d749881603b4cdf41944c70ad/cmake/cxx_static/CMakeLists.txt#L8-L14) because it was straight forward.

gatecat:
Thanks, that worked, and the wire test output looks good! Is it worth adding a note to the fpga_interchange README about this issue?
Screenshot from 2021-02-18 16-49-41

litghost:
> Thanks, that worked, and the wire test output looks good! Is it worth adding a note to the fpga_interchange README about this issue?

Ya.  I need to update the README as well about what is working.

fpga_interchange/arch.h
litghost: Fixed.

fpga_interchange/site_router.cc
litghost: Fixed!

fpga_interchange/site_router.cc
litghost: Indeed, fixed it up!

litghost:
> > Thanks, that worked, and the wire test output looks good! Is it worth adding a note to the fpga_interchange README about this issue?
> 
> Ya. I need to update the README as well about what is working.

README is updated with more instructions and latest state of the FPGA interchange arch.

litghost:
I believe I've address the comments from the latest review.

PR #577, opened 2021-02-11
MachXO2 Backend

cr1901:
cc: @gatecat and @litghost

I was told to submit the _working_ backend as-is when ready. I'm aware that there have been some API changes in the past 2 months or so, and this backend needs to bring them in.


machxo2/arch.cc
```diff
@@ -0,0 +1,532 @@
...
+    NPNR_ASSERT(src != BelId());
+    NPNR_ASSERT(dst != BelId());
+
+    // TODO: Same deal applies here as with estimateDelay.
+    return (abs(dst.location.x - src.location.x) + abs(dst.location.y - src.location.y)) * (0.01 + 0.01);
+}
+
+bool Arch::getBudgetOverride(const NetInfo *net_info, const PortRef &sink, delay_t &budget) const { return false; }
+
+ArcBounds Arch::getRouteBoundingBox(WireId src, WireId dst) const
```
gatecat, line 418: This is going to cause `router2` to fail in some exciting way or other, it'd be better to assertion fail than return empty bounds here

machxo2/archdefs.h
```diff
@@ -0,0 +1,183 @@
...
+typedef IdString GroupId;
+typedef IdString DecalId;
+
+struct ArchNetInfo
+{
+};
+
+struct NetInfo;
+
+struct ArchCellInfo
```
gatecat, line 129: looks like the fields here are a leftover from nextpnr-generic that aren't needed

cr1901: Indeed, I didn't need `ArchCellInfo` for the initial port (though I'm sure I will need it once packing is halfway-intelligent).

gatecat:
Closing now #577 has been merged

PR #579, opened 2021-02-11
Add control to whether GenericFrontend splits IO ports.

litghost:


frontend/frontend_base.h
```diff
@@ -139,14 +139,15 @@ template <typename FrontendType> struct GenericFrontend
...
+    const bool split_io;
     using mod_dat_t = typename FrontendType::ModuleDataType;
     using mod_port_dat_t = typename FrontendType::ModulePortDataType;
     using cell_dat_t = typename FrontendType::CellDataType;
     using netname_dat_t = typename FrontendType::NetnameDataType;
     using bitvector_t = typename FrontendType::BitVectorDataType;
 
     std::unordered_map<IdString, ModuleInfo> mods;
-    std::unordered_map<IdString, const mod_dat_t &> mod_refs;
+    std::unordered_map<IdString, const mod_dat_t> mod_refs;
```
gatecat, line 150: can you explain the reason for this change? it doesn't seem linked to the rest of the PR, even though I suspect it is a good idea

common/nextpnr.h
```diff
@@ -833,6 +833,7 @@ struct BaseCtx
 
     // Top-level ports
     std::unordered_map<IdString, PortInfo> ports;
+    std::unordered_map<IdString, CellInfo *> port_cells;
```
gatecat, line 836: can you add a brief note about `port_cells` to `docs/netlist.md`?

frontend/frontend_base.h
litghost: Because `mod_dat_t` is simple `FrontendType::ModuleDataType`, the type is totally under the control of the `FrontendType`.  The json frontend already marks `ModuleDataType` as a reference, but the FPGA interchange frontend is instead using a lightwight value type.

See https://github.com/YosysHQ/nextpnr/blob/c956cae8244c094783edc7101fd0ca542c24e55b/frontend/json_frontend.cc#L37

If you want I can separate this change from this PR.

common/nextpnr.h
litghost: Done.  

There is a broader question about separating the root level cell namespace and the root level port namespace.  In Verilog I believe it is legal to have root level ports and root level cells with the same name, but this causes a failure in nextpnr.  I believe the "correct" long term answer is to have a separate namespace for root level ports and cell names, but I don't think this needs to be fixed right now.

gatecat: I think it could also make sense to use a new type (eg TopLevelPortInfo) other than PortInfo here that could store things like attributes and also partition pin information for partially reconfigurable designs. But this is definitely not for this PR but another day.

frontend/frontend_base.h
gatecat: That makes sense, I'm happy with that being part of this PR.

PR #585, opened 2021-02-16
Remove isValidBelForCell

gatecat:
This Arch API dates from when we were first working out how to implement placement validity checking, and in practice is little used by the core parts of placer1/HeAP and the Arch implementation involves a lot of duplication with isBelLocationValid.

In the short term; placement validity checking is better served by the combination of checkBelAvail and isValidBelForCellType before placement; followed by isBelLocationValid after placement (potentially after moving/swapping multiple cells).

Longer term, removing this API makes things a bit cleaner for a new validity checking API.


litghost:
This seems like a good change to me.

PR #586, opened 2021-02-16
Add Cell -> BEL Pin maps to FPGA interchange arch.

litghost:
This also expands the FPGA interchange Arch BBA to include placement constraints, but doesn't implement them yet.

### DIFF
--- a/fpga_interchange/arch.cc
+++ b/fpga_interchange/arch.cc
@@ -126,9 +126,17 @@ Arch::Arch(ArchArgs args) : args(args)
     } else {
         // Default to first package.
         NPNR_ASSERT(chip_info->packages.size() > 0);
-        IdString package_name(chip_info->packages[0].package);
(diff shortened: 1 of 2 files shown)


fpga_interchange/arch.cc
```diff
@@ -87,6 +109,43 @@ Arch::Arch(ArchArgs args) : args(args)
...
+                NPNR_ASSERT(package_index == -1);
+                package_index = i;
+            }
+        }
+
+        if (package_index == -1) {
+            log_error("Could not find package '%s' in chipdb.\n", this->args.package.c_str());
+        }
+    } else {
+        // Default to first package.
```
gatecat, line 127: ime, this behaviour has caused significant user confusion before when they forget to specify a package (see #304), it would probably be better to make `--package` required if there is more than one package in the database

litghost: Sure, done.

PR #589, opened 2021-02-17
Add a Bits utility library.

litghost:
This library captures use of __builtin_popcount and __builtin_ctz on
GCC/clang and hopefully handles the MSVC case.

### DIFF
--- a/common/bits.cc
+++ b/common/bits.cc
@@ -1,0 +1,51 @@
+/*
+ *  nextpnr -- Next Generation Place and Route
+ *
+ *  Copyright (c) 2013  Mike Pedersen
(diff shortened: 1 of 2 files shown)


common/bits.cc
```diff
@@ -0,0 +1,51 @@
+/*
+ *  nextpnr -- Next Generation Place and Route
+ *
+ *  Copyright (c) 2013  Mike Pedersen
```
gatecat, line 4: Just to check, this code is OK to be under the ISC license?

litghost: Ya, it was under MIT, hence the copyright notice.

PR #590, opened 2021-02-17
Expose ice40 arch placer-heap internal parameters.

cbalint13:
Expose to users more control over internal parameters.

* For now ice40 only and for the placer heap , but looking forward for more suggestions.
* It is useful for automated annealing see an early experiment: https://github.com/cbalint13/nextpnr-anneal

Thank you !


gatecat:
Thanks!

PR #591, opened 2021-02-19
Add constant network support to FPGA interchange arch

litghost:
This adds initial constant network support (along side changes in https://github.com/SymbiFlow/python-fpga-interchange/pull/23).  I believe logic needs to be added to combine GND/VCC cells/nets, but this doesn't appear to matter with the initial const wire designs.  Once I have an example netlist with multiple GND/VCC cells/nets, I'll add the remaining logic.


gatecat:
Approach here looks good. Some more work might be needed in the future to improve performance of constant routing where you have a single wire with a very large number of pips fanning off it; the backwards routing heuristic in router2 works well for this but doesn't currently deal with congestion which will fall back to regular forwards routing which will have to iterate over a lot of pips. 

This was more of a problem for xcup rather than xc7 in my nextpnr-xilinx experience, due to GND constants needing to come from a nearby LUT creating more possibilities for congestion (also xcup devices just being larger). One thing I did was to have a three level "global->row->tile" type structure for constant wires so the router could make targeted progress; but another solution would be to special-case this in the router, for example always using backwards routing for constants and extending the backwards router to work around congestion.

litghost:
> the backwards routing heuristic in router2 works well for this but doesn't currently deal with congestion which will fall back to regular forwards routing which will have to iterate over a lot of pips.

My plan is to have the site router push all constants to the site pins, which will address this explicitly.  Because the site router ensures a congestion free site routing, and dedicated site constant sources are preferred over routed constant sources, and because of site inverter <-> constant net relationships, handling this in the site router should avoid any congestion.

litghost:
> This was more of a problem for xcup rather than xc7 in my nextpnr-xilinx experience, due to GND constants needing to come from a nearby LUT creating more possibilities for congestion (also xcup devices just being larger). One thing I did was to have a three level "global->row->tile" type structure for constant wires so the router could make targeted progress; but another solution would be to special-case this in the router, for example always using backwards routing for constants and extending the backwards router to work around congestion.

Ya, this might cause some problems.  The current constant network does not have pseudo site pips to the LUT outputs, though it needs them.  xc7 graphs are easier in this regard.  I think this is a case of "crawl", "walk", "run", e.g. fix it once we can demostrate the issue.

Overall I think the constant network structure is transparent to the arch, and can be tweaked as needed as problems arise.

litghost:
@gatecat This is ready for review when you get a chance.  I've gotten a counter design mostly working, but I need to fix LUT rotation / port sharing before that is ready.

fpga_interchange/arch.cc
```diff
@@ -677,13 +696,33 @@ DecalXY Arch::getGroupDecal(GroupId pip) const { return {}; };
...
+    int src_x, src_y;
+    get_tile_x_y(net_info->driver.cell->bel.tile, &src_x, &src_y);
+
+    int dst_x, dst_y;
+    get_tile_x_y(sink.cell->bel.tile, &dst_x, &dst_y);
+
+    delay_t base = 30 * std::min(std::abs(dst_x - src_x), 18) + 10 * std::max(std::abs(dst_x - src_x) - 18, 0) +
+                   60 * std::min(std::abs(dst_y - src_y), 6) + 20 * std::max(std::abs(dst_y - src_y) - 6, 0) + 300;
+
+    base = (base * 3) / 2;
```
gatecat, line 724: for this lookahead to be more effective, I would also suggest returning a pip delay on the order of 100ps

fpga_interchange/dedicated_interconnect.cc
```diff
@@ -0,0 +1,785 @@
...
+            if (ctx->is_site_port(pip)) {
+                switch (node_to_expand.state) {
+                case IN_SOURCE_SITE:
+                    NPNR_ASSERT(wire_data.site == -1);
+                    next_node.state = IN_ROUTING;
+                    break;
+                case IN_ROUTING:
+                    NPNR_ASSERT(wire_data.site != -1);
+                    if (wire.tile == src_wire.tile && wire_data.site == src_wire_data.site) {
+                        // Dedicated routing won't have straight loops,
```
gatecat, line 133: Are we sure this is a universal truth? It does seem like a good heuristic, but I'm wondering if there are places where combinations of routethroughs/loopbacks might violate this.

fpga_interchange/dedicated_interconnect.cc
```diff
@@ -0,0 +1,785 @@
...
+                        log_info("BEL %s is not valid because pin %s cannot reach %s/%s\n", ctx->nameOfBel(driver_bel),
+                                 driver_bel_pin.c_str(ctx), ctx->nameOfBel(sink_bel), sink_bel_pin.c_str(ctx));
+                    }
+                    return false;
+                }
+
+                // Do detailed routing check to ensure driver can reach sink.
+                //
+                // FIXME: This might be too slow, but it handles a case on
+                // SLICEL.COUT -> SLICEL.CIN has delta_y = {1, 2}, but the
```
gatecat, line 244: This is going to be a problem when we start inserting relative placement constraints (which should give much better performance than essentially relying on trial and error) as these require fixed dx/dy values. nextpnr-xilinx punted on this but I think a way of specifying the "next" bel/site might make sense as a future API improvement.

fpga_interchange/dedicated_interconnect.cc
```diff
@@ -0,0 +1,785 @@
...
+    NPNR_ASSERT(bel != BelId());
+
+    for (const auto &port_pair : cell->ports) {
+        IdString port_name = port_pair.first;
+        NetInfo *net = port_pair.second.net;
+        if (net == nullptr) {
+            continue;
+        }
+
+        // This net doesn't have a driver, probably not valid?
```
gatecat, line 350: Undriven nets aren't necessarily a problem (for example if you connect a net to an input but don't drive it); in other arches we tend to skip over them.

fpga_interchange/dedicated_interconnect.cc
```diff
@@ -0,0 +1,785 @@
...
+    for (BelId bel : ctx->getBels()) {
+        const auto &bel_data = bel_info(ctx->chip_info, bel);
+        if (bel_data.category != BEL_CATEGORY_LOGIC) {
+            continue;
+        }
+        if (bel_data.synthetic) {
+            continue;
+        }
+
+        for (size_t i = 0; i < bel_data.num_bel_wires; ++i) {
```
gatecat, line 444: mixed signedness compare warning (I'm minded to tighten up the compiler flags so more of these things become CI failures, not sure what your opinion is on that)

fpga_interchange/fpga_interchange.cpp
```diff
@@ -248,6 +274,37 @@ static void emit_net(
...
+
+    wires_to_expand.push_back(root_wire);
+    while(!wires_to_expand.empty()) {
+        WireId wire = wires_to_expand.back();
+        wires_to_expand.pop_back();
+
+        auto downhill_iter = pip_downhill.find(wire);
+        if(downhill_iter == pip_downhill.end()) {
+            if(root_wire != wire) {
+                log_warning("Wire %s never entered the real fabric?\n",
```
gatecat, line 290: should this be a warning or an error?

gatecat:
my attempt to build the bba locally is failing with:

```
capnp.lib.capnp.KjException: home/david/nextpnr-master/fpga_interchange/examples/create_bba/build/fpga-interchange-schema/interchange/DeviceResources.capnp:546: failed: Union must have at least two members.
```

I think this is related to the LUT init stuff based on the line number; and checking out the commit before that PR was merged seems to work OK.

litghost:
> 
> 
> my attempt to build the bba locally is failing with:
> 
> ```
> capnp.lib.capnp.KjException: home/david/nextpnr-master/fpga_interchange/examples/create_bba/build/fpga-interchange-schema/interchange/DeviceResources.capnp:546: failed: Union must have at least two members.
> ```
> 
> I think this is related to the LUT init stuff based on the line number; and checking out the fpga-interchange-schema commit before that PR was merged seems to work OK.

Ya, I found this too and need to fix it.  Need to stand up a CI on `fpga-interchange-schema` to catch this kind of stuff.

fpga_interchange/fpga_interchange.cpp
litghost: This is a warning right now because the constant merging logic always emits the VCC/GND root, even if unused.  Then when walking that root, it won't enter the real fabric.  In the long term we can prune away an unused VCC/GND root and make this an error again.

fpga_interchange/dedicated_interconnect.cc
litghost: Please tighten up the flags.  We also should emit `-Wxxx` on local builds.  I'm using `clang` and it is capable of emitting the same warnings as the CI, but that isn't the current default (I believe).

fpga_interchange/arch.cc
litghost: Ok

fpga_interchange/dedicated_interconnect.cc
litghost: I'd need to see an example of where a straight loop forms a dedicated interconnect (e.g. placement dependent).  All the cases I know are site to another site.  Do you know of any counter examples?

The good news is if this logic fails on an arch, it will manifest eventually as a routing failure, which will generate a bug report, rather than an optimization failure.

fpga_interchange/dedicated_interconnect.cc
litghost: Agree that relative placement constraints should be emitted where possible.  Something that is tricky here is that the use of dedicated interconnect is not 1-to-1 with placement constraints.  Carry chains and DSP chains are good candidates for relative placement constraints, but the IFF/OFF cases are a good example where it should be up to the placer to determine if it wants to use the IFF/OFF.  It might make more sense to use a SLICEL/M FF than an IO FF depending on the shape and fanout of the signals.  Just because an IO FF is possible doesn't mean we should constraint the solution to that.

fpga_interchange/dedicated_interconnect.cc
litghost: I believe you mean to an output, but don't drive it.  A disconnected cell primitive input is a bug in most cases I can think of.  It must be either 1/0/signal, rather than "unknown".

fpga_interchange/fpga_interchange.cpp
gatecat: That's fair as a temporary thing; from an end-user point of view I don't much like the idea of a warning that isn't immediately end-user-actionable but I imagine it will be long gone by the time this is end-user-read.

fpga_interchange/dedicated_interconnect.cc
gatecat: Yes, this is an interesting thing to think about. Somewhat related is the dedicated LUT->FF routing which you want to use if you can; but definitely don't want to force all the time.

One idea I had in Nexus was a separate post-place pass that performs low-distance swaps only with the goal of using this dedicated interconnect (currently only LUT-FF, I've been meaning to look at doing similar with the dedicated fast LUT-LUT pattern that Nexus has too): https://github.com/YosysHQ/nextpnr/blob/master/nexus/post_place.cc

A generic version of that pass as a post-place optimisation to opportunistically exploit dedicated routing, by swapping things within a site or between adjacent sites, could definitely improve things from a routeability/timing point of view down the line.

litghost: > Yes, this is an interesting thing to think about. Somewhat related is the dedicated LUT->FF routing which you want to use if you can; but definitely don't want to force all the time.

Yep, and same story with LUT -> CARRY and CARRY -> FF placement.  In general, there needs to be a way to align site local placement to exploit specialized site routing to achieve maximal packing density.  Once I debug the FPGA arch site router for correctness, I believe this is going to be an important improvement during HeAP legalization to achieve really good results.

litghost:
> my attempt to build the bba locally is failing with:
> 
> ```
> capnp.lib.capnp.KjException: home/david/nextpnr-master/fpga_interchange/examples/create_bba/build/fpga-interchange-schema/interchange/DeviceResources.capnp:546: failed: Union must have at least two members.
> ```
> 
> I think this is related to the LUT init stuff based on the line number; and checking out the fpga-interchange-schema commit before that PR was merged seems to work OK.

I've fixed the bug and stood up a simple CI to find these issues with https://github.com/SymbiFlow/fpga-interchange-schema/pull/13

PR #595, opened 2021-02-23
Change CellInfo in getBelPinsForCellPin to be const.

litghost:


gatecat:
Don't forget `docs/archapi.md`, too (arguably we should be autogenerating this from comments in the long term now we have a more formal form of the API...)

litghost:
> Don't forget `docs/archapi.md`, too (arguably we should be autogenerating this from comments in the long term now we have a more formal form of the API...)

Will do!

litghost:
> Don't forget `docs/archapi.md`, too (arguably we should be autogenerating this from comments in the long term now we have a more formal form of the API...)

Fixed!

PR #597, opened 2021-02-23
Add dynamic bitarray to common library.

litghost:
Test PR is over here: https://github.com/YosysHQ/nextpnr-tests/pull/6

### DIFF
--- a/common/dynamic_bitarray.h
+++ b/common/dynamic_bitarray.h
@@ -39,10 +39,7 @@ template <typename Storage = std::vector<uint8_t>> class DynamicBitarray
         std::fill(storage.begin(), storage.end(), value ? std::numeric_limits<typename Storage::value_type>::max() : 0);
     }
 
-    constexpr size_t bits_per_value() const
-    {
-        return sizeof(typename Storage::value_type) * std::numeric_limits<typename Storage::value_type>::digits;
-    }
+    constexpr size_t bits_per_value() const { return std::numeric_limits<typename Storage::value_type>::digits; }
 
     bool get(size_t bit) const
     {
@@ -67,7 +64,7 @@ template <typename Storage = std::vector<uint8_t>> class DynamicBitarray
 
     void resize(size_t number_bits)
     {
-        size_t required_storage = (number_bits + bits_per_value()) / bits_per_value();
+        size_t required_storage = (number_bits + bits_per_value() - 1) / bits_per_value();
         storage.resize(required_storage);
     }
 
--- a/common/dynamic_bitarray.h
+++ b/common/dynamic_bitarray.h
@@ -1,0 +1,82 @@
+/*
+ *  nextpnr -- Next Generation Place and Route
+ *
+ *  Copyright (C) 2021  Symbiflow Authors
+ *
+ *  Permission to use, copy, modify, and/or distribute this software for any
+ *  purpose with or without fee is hereby granted, provided that the above
+ *  copyright notice and this permission notice appear in all copies.
+ *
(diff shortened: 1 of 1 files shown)


common/dynamic_bitarray.h
```diff
@@ -0,0 +1,82 @@
...
+#include <limits>
+#include <vector>
+
+// This class implements a simple dynamic bitarray, backed by some resizable
+// random access storage.  The default is to use a std::vector<uint8_t>.
+
+#ifndef DYNAMIC_BITARRAY_H
+#define DYNAMIC_BITARRAY_H
+
+namespace nextpnr {
```
gatecat, line 30: This should be NEXTPNR_NAMESPACE_BEGIN like in other files

common/dynamic_bitarray.h
```diff
@@ -0,0 +1,82 @@
...
+    static_assert(!std::numeric_limits<typename Storage::value_type>::is_signed, "Storage must be unsigned!");
+
+    void fill(bool value)
+    {
+        std::fill(storage.begin(), storage.end(), value ? std::numeric_limits<typename Storage::value_type>::max() : 0);
+    }
+
+    constexpr size_t bits_per_value() const
+    {
+        return sizeof(typename Storage::value_type) * std::numeric_limits<typename Storage::value_type>::digits;
```
gatecat, line 44: I'm not sure if multiplying sizeof by digits will be correct for non-8-bit types - I would expect just digits to be enough?

common/dynamic_bitarray.h
```diff
@@ -0,0 +1,82 @@
...
+        if (value) {
+            storage.at(element_index) |= (1 << bit_offset);
+        } else {
+            storage.at(element_index) &= ~(1 << bit_offset);
+        }
+    }
+
+    void resize(size_t number_bits)
+    {
+        size_t required_storage = (number_bits + bits_per_value()) / bits_per_value();
```
gatecat, line 70: Doesn't really matter but I think you could save a byte and do `bits_per_value()-1` here ?

common/dynamic_bitarray.h
litghost: Correct, I messed this up.  It used to be `sizeof(typename Storage::value_type) * CHAR_BIT` and then I brought in `std::numeric_limits`.

common/dynamic_bitarray.h
litghost: I'm avoiding depending on `nextpnr.h` where possible because it causes header nightmares where include order can cause terrible compiler errors.

gatecat: Yep, I can see that. The original idea for `NEXTPNR_NAMESPACE_BEGIN` was to enable multiple nextpnrs to be built with different namespaces and linked together but afaik no-one is using that (it would need a custom main if nothing else) but this has never happened so `namespace nextpnr` is OK for now.

common/dynamic_bitarray.h
litghost: I've fixed this and added a test about it.

common/dynamic_bitarray.h
litghost: I've fixed this and added a test about it.

gatecat (approved):
I've merged the test PR, can you bump the submodule so we get a CI run with the new tests?

litghost:
> I've merged the test PR, can you bump the submodule so we get a CI run with the new tests?

Sure, one minute!

litghost:
> > I've merged the test PR, can you bump the submodule so we get a CI run with the new tests?
> 
> Sure, one minute!

Done

PR #598, opened 2021-02-25
Tighten up compiler flags

gatecat:
This enables `-Wextra`; and also optionally `-Werror` (including enabling `-Werror` for CI). `-Werror` is not enabled by default, otherwise there is a poor out-of-the-box experience when someone inevitably builds with an older/newer compiler that has a different set of diagnostics and ends up with an error.


whitequark:
This broke WASI builds: https://github.com/YoWASP/nextpnr/runs/1983657117?check_suite_focus=true

PR #603, opened 2021-02-26
Prevent trival misplacements in placer1.

litghost:


gatecat:
I think you might need something similar for `bound` in the swapping-chains case: https://github.com/YosysHQ/nextpnr/blob/c9ee36b80529a8e7bb7e196e288d63aedad979a7/common/placer1.cc#L659

litghost:
> I think you might need something similar for `bound` in the swapping-chains case:
> 
> https://github.com/YosysHQ/nextpnr/blob/c9ee36b80529a8e7bb7e196e288d63aedad979a7/common/placer1.cc#L659

Good catch, fix is in.

PR #604, opened 2021-02-26
Add counter test for FPGA interchange

litghost:
This adds a very simple LUT rotation during site routing, which enables a valid counter test to be place and routed. Needs BBA changes from https://github.com/SymbiFlow/python-fpga-interchange/pull/26 and on top of fix from #603 

There is still a gap around LUT route through.  Both the site router and router need to emit psuedo-cells when using LUT route-throughs.


fpga_interchange/arch.cc
```diff
@@ -192,6 +193,50 @@ Arch::Arch(ArchArgs args) : args(args)
...
+
+    // Map lut cell types to their LutCellPOD
+    for (const LutCellPOD &lut_cell : chip_info->cell_map->lut_cells) {
+        IdString cell_type(lut_cell.cell);
+        auto result = lut_cells.emplace(cell_type, &lut_cell);
+        NPNR_ASSERT(result.second);
+    }
+
+    raw_bin_constant = std::regex("[01]+", std::regex_constants::ECMAScript | std::regex_constants::optimize);
+    verilog_bin_constant =
```
gatecat, line 235: I appreciate the priority is getting something working right now; but this kind of stuff is why I'd like to see something other than Verilog-constants-in-strings as the bitvector format long term.

fpga_interchange/arch.cc
```diff
@@ -666,8 +712,61 @@ bool Arch::route()
     } else {
         log_error("FPGA interchange architecture does not support router '%s'\n", router.c_str());
     }
+
+    if (result) {
+        // Fixup LUT vcc pins.
+        IdString vcc_net_name(chip_info->constants->vcc_net_name);
```
gatecat, line 718: I think it'd be neater to have this in another function like `Arch::fixup_lut_vcc` to make `route` a bit simpler

litghost: I agree with that, I'll refactor it.

gatecat:
This seems reasonable for now. aiui this is putting quite a bit of stuff into the site router that can be optimised later on; for example the majority of the time there won't be a need to check the equation if a simple check of the total number of unique inputs is satisfied (e.g. two LUT2s can always be combined into a LUT5/6 pair regardless of equation).

fpga_interchange/arch.cc
litghost: Changing that will require patching up https://github.com/Xilinx/RapidWright/blob/4ed91eab267c4476e4e2318f89864bc3d63cae69/src/com/xilinx/rapidwright/interchange/LogNetlistReader.java#L57-L80 .  I'm not saying that we shouldn't change it, just that there are couple moving parts here.  Because we want to stand up a direct FPGA interchange to FASM emitter (https://github.com/SymbiFlow/python-fpga-interchange/issues/27) we should probably cut over to something sooner rather than later.

It is worth noting that right now it appears that all bitstrings are coming into the LogicalNetlist as char vectors of `1` and `0` (e.g. "raw_bin_constant").  This appears to be accepted by RapidWright/Vivado for LUTs.  I am expecting this will be a problem for RAMB18E1 init strings.

litghost:
> This seems reasonable for now. aiui this is putting quite a bit of stuff into the site router that can be optimised later on; for example the majority of the time there won't be a need to check the equation if a simple check of the total number of unique inputs is satisfied (e.g. two LUT2s can always be combined into a LUT5/6 pair regardless of equation).

Total agreement.  This was the first general implementation I could think of.  I agree that this can be optimized as we continue.  I think my short term goal is to get to murax/picosoc class designs with LUT-RAMs, BRAMs, LUTs and FFs working before coming back and rewriting most of the site router for speed.  After LUT-RAMs are working, I'll likely test how big the gap is to get all the way to something like a Linux-capable LiteX (e.g. differential IO and more complicated IO logic), but I expect optimizations will be needed to reach that class of design.

I also want to stand up significantly better testing around the site router so that we can test specific packing scenarios and ensure they generate the correct result from `isBelValid`.

fpga_interchange/arch.cc
litghost: Done

gatecat:
Just to double check, are you OK with this being merged now?

litghost:
> Just to double check, are you OK with this being merged now?

Yep!

PR #605, opened 2021-02-26
Add placement sanity check in placer_heap.

litghost:
Also check return of placer1_refine.


gatecat:
For consistency, would you be willing to use the scoped lock in the router, too?

litghost:
> For consistency, would you be willing to use the scoped lock in the router, too?

Sure, done.

gatecat (approved):

PR #606, opened 2021-02-28
Gowin DFF fixes

pepijndevos:
Under the old generic target, only plain simple DFF were supported, but the new target supports all sorts of exotic reset modes and it turns out this was not tested very well at all.

This fix makes sure that no two DFF of different types can occupy the same slice.

There is at least 1 more bug of this kind, so feel free to leave this open, or merge and I'll make new PRs as I fix more bugs.


gatecat (approved):
LGTM

PR #607, opened 2021-03-01
Add absl::flat_hash_map.

litghost:
This lowers the CPU cost of using the flat wire map in router2, and should
use less memory as well.


gatecat:
Some quick benchmarks for total runtime; although fairly primitive it's enough I'm happy this is a clear improvement:

```
LiteX VexRiscv on CrossLink-NX VIP

Before: Runtime: min = 84.72s, avg = 110.00s, max = 147.48s
After: min = 82.29s, avg = 100.29s, max = 125.80s

LiteX VexRiscv on ECP5 Versa

Before: min = 102.34s, avg = 109.13s, max = 117.10s
After: min = 97.38s, avg = 107.32s, max = 112.96s

MARLANN on iCE40

Before: min = 17.51s, avg = 20.33s, max = 22.21s
After: min = 16.53s, avg = 19.80s, max = 21.89s
```

@whitequark can you double check this isn't going to cause any problems for the wasm build? It looks like a reasonable dependency to me and there might be other useful things in there too.

PR #612, opened 2021-03-08
router2: Fix vast perf drop when leaving bounding box

gatecat:


litghost:
I think this fix was in error?  The logic used to be "limit iteration count for is_bb = true, no limit on iteration count for is_bb = false".  This results in incorrectly reporting "ERROR: Failed to route arc" when the problem was that the router just needed to keep running.

In particular, the CI failures in #659 would not occur with this change removed.  There is an argument that a better lookahead function results in better routing, but I'd argue it's better for the router to take longer, and finish, than incorrectly report no route is possible?

litghost:
Ok, I've replicated the issue.  The solution is not to remove the `!is_bb ||` check, but stop the router if iter > to_explore **and a solution has been found** if !is_bb.  I'll submit a PR with the fix.

litghost:
> Ok, I've replicated the issue. The solution is not to remove the `!is_bb ||` check, but stop the router if iter > to_explore **and a solution has been found** if !is_bb. I'll submit a PR with the fix.
> 
> :edit: This isn't quiet right, because toexplore should lowered at
> 
> https://github.com/YosysHQ/nextpnr/blob/bb6079133c9b0de9db3e39735d160c1a161ec981/common/router2.cc#L729-L731
> 
> Still investing

My original theory was correct, solution in https://github.com/YosysHQ/nextpnr/pull/663

PR #613, opened 2021-03-08
Add basic support for Nexus DPHY hard IP

gatecat:
This hasn't been extensively tested yet, but in a quick test demo I can see the output data visibly change when covering the camera so it's at least getting something.

this is on top of #612 which was discovered during testing


siliconwitch:
Don't suppose you still have the test demo for this? 

Having some trouble wrapping my head around configuration for the DPHY block, and the minimum required wiring 🤔

gatecat:
Yeah, I did some tests with https://github.com/gatecat/litex-nexus-mipi and was able to get plausible packets out of the camera on the VIP board

PR #615, opened 2021-03-09
Add diagnostic prints to debug lookahead performance.

litghost:


gatecat:
For some strange reason CI doesn't seem to have picked this up. Can you try another force-push?

litghost:
> For some strange reason CI doesn't seem to have picked this up. Can you try another force-push?

Will do!

PR #617, opened 2021-03-10
Only depend on Abseil in threaded builds.

### LINKED ISSUE #616: Dependency on Abseil making WASM/YoWASP builds impossible
Abseil (with CMake, at least) has a [hard dependency](https://github.com/abseil/abseil-cpp/blob/master/CMakeLists.txt#L98) on threading, which breaks builds on platforms without threads. The upstream [worked around this](https://github.com/abseil/abseil-cpp/issues/693) for Bazel builds but not CMake. Unfortunately, the workaround (or its CMake equivalent) fixes the immediate issue (no `-pthread`) but it doesn't actually let the build complete on platforms without atomics, at least not for the library that nextpnr wants to use.

whitequark:
Abseil has a hard dependency on threads (not just in the build system but in many places in the base libraries), so there is no way to use it on WASI at the moment. Assuming that Abseil will only be used for speed and not because it provides a richer API, I suggest this workaround.

Fixes #616.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -274,8 +274,10 @@ foreach (family ${ARCH})
     # Include the family-specific CMakeFile
     include(${family}/family.cmake)
     foreach (target ${family_targets})
-        target_link_libraries(${target} PRIVATE absl::flat_hash_map)
-        target_link_libraries(${target} PRIVATE absl::flat_hash_set)
+        if (USE_THREADS)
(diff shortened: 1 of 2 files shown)


gatecat:
Thanks, this seems like a reasonable option to unbreak the build for now but I'll hold off adding any more dependency on abseil until we think about this more.

whitequark:
Thank you!

PR #619, opened 2021-03-12
Add CMake infrastructure for fpga interchange

acomodi:
This PR is an initial CMake infrastracture integration for the FPGA interchange.

CMake can generate targets to run:

- RW device generation
- chipdb bba generation
- tests

The tools used to do so must be pre-installed in the system (RW, python-fpga-interchange, etc.) and cach'd vars are used to point to them.


acomodi:
@litghost @gatecat FYI.

I marked as WIP as I still need to add targets to run archcheck and also because I get a segfault error during the physical netlist generation:

```
Program received signal SIGSEGV, Segmentation fault.
0x00005555557ec09c in nextpnr_fpga_interchange::RelPtr<nextpnr_fpga_interchange::RelSlice<nextpnr_fpga_interchange::RelPtr<char> > >::get (this=0x7ffffd7fd2bc) at /data/interchange/nextpnr/common/relptr.h:7
7           const T *get() const { return reinterpret_cast<const T *>(reinterpret_cast<const char *>(this) + offset); }
(gdb) backtrace
#0  0x00005555557ec09c in nextpnr_fpga_interchange::RelPtr<nextpnr_fpga_interchange::RelSlice<nextpnr_fpga_interchange::RelPtr<char> > >::get (this=0x7ffffd7fd2bc) at /data/interchange/nextpnr/common/relptr.h:7
#1  0x00005555557e97b6 in nextpnr_fpga_interchange::RelPtr<nextpnr_fpga_interchange::RelSlice<nextpnr_fpga_interchange::RelPtr<char> > >::operator* (this=0x7ffffd7fd2bc) at /data/interchange/nextpnr/common/relptr.h:11
#2  0x00005555557ddd7f in nextpnr_fpga_interchange::Arch::Arch (this=0x555555ce8a10, args=...) at /data/interchange/nextpnr/fpga_interchange/arch.cc:95
#3  0x000055555568b820 in nextpnr_fpga_interchange::Context::Context (this=0x555555ce8a10, args=...) at /data/interchange/nextpnr/common/nextpnr.h:1464
#4  0x00005555558c1296 in FpgaInterchangeCommandHandler::createContext (this=0x7fffffffd760, values=std::unordered_map with 0 elements) at /data/interchange/nextpnr/fpga_interchange/main.cc:83
#5  0x0000555555595b32 in nextpnr_fpga_interchange::CommandHandler::exec (this=0x7fffffffd760) at /data/interchange/nextpnr/common/command.cc:446
#6  0x00005555558c1926 in main (argc=11, argv=0x7fffffffdbd8) at /data/interchange/nextpnr/fpga_interchange/main.cc:116
```

I traced it back to this line https://github.com/YosysHQ/nextpnr/blob/bffaad477a541459e072339c28edf0087437c254/fpga_interchange/arch.cc#L95 Any idea on what might be the cause?

fpga_interchange/examples/chipdb.cmake
```diff
@@ -0,0 +1,112 @@
...
+            ${rapidwright_device_target}
+    )
+
+    add_custom_target(constraints-${part}-device DEPENDS ${constraints_device})
+
+    set(constraints_luts_device ${CMAKE_CURRENT_BINARY_DIR}/${part}_constraints_luts.device)
+    add_custom_command(
+        OUTPUT ${constraints_luts_device}
+        COMMAND
+            python3 -mfpga_interchange.patch
```
gatecat, line 73: other arches tend to use `${PYTHON_EXECUTABLE}` rather than `python3` here to allow it to be overridden in some cases if needed

fpga_interchange/examples/devices/xc7a35t/CMakeLists.txt
```diff
@@ -0,0 +1,2 @@
+generate_chipdb(part xc7a35tcsg324-1)
```
gatecat, line 1: somewhat incidental to this PR, but given the routing graph is the same, having one chipdb across all packages would make more sense imo

gatecat:
> I marked as WIP as I still need to add targets to run archcheck and also because I get a segfault error during the physical netlist generation:

could it be a mismatch between the structures in nextpnr and the bba generator? (for nextpnr-nexus I have a version field in the bba that I hopefully increment every time this changes as a crude attempt to catch stuff like this)

fpga_interchange/examples/chipdb.cmake
acomodi: Sure, done.

fpga_interchange/examples/devices/xc7a35t/CMakeLists.txt
acomodi: Definetely makes sense, I am not sure whether the part-specific data is required somewhere, but for now I changed it such that only one device is generated (for xc7a35t and xc7a50t, the common device is the 50t)

fpga_interchange/examples/chipdb.cmake
```diff
@@ -0,0 +1,134 @@
...
+    )
+
+    set(device ${create_rapidwright_device_db_device})
+    set(part ${create_rapidwright_device_db_part})
+    set(rapidwright_device_db ${CMAKE_CURRENT_BINARY_DIR}/${part}.device)
+    add_custom_command(
+        OUTPUT ${rapidwright_device_db}
+        COMMAND
+            RAPIDWRIGHT_PATH=${RAPIDWRIGHT_PATH}
+            ${RAPIDWRIGHT_PATH}/scripts/invoke_rapidwright.sh
```
Xiretza, line 21: Could the path to `invoke_rapidwright.sh` be put in a cmake variable? The whole concept of `RAPIDWRIGHT_PATH` kind of breaks down when rapidwright is packaged instead of just lying around in a local git repo, and it would be nice to be able to configure this sinlge entry point to fit the installation method.

fpga_interchange/examples/devices/xc7a35t/CMakeLists.txt
litghost: So because of how the device database is structured, the current code in python-fpga-interchange will generate a chipdb for all packages present in the input device database.  This does mean that if you request a xc7a35tcsg324-1, you get a database for all `xc7a35t` devices.

fpga_interchange/examples/chipdb.cmake
```diff
@@ -0,0 +1,134 @@
+function(create_rapidwright_device_db)
+    set(options)
```
litghost, line 2: Please add a documentation block at the top.

fpga_interchange/examples/chipdb.cmake
litghost: I agree with this.

fpga_interchange/examples/chipdb.cmake
```diff
@@ -0,0 +1,134 @@
...
+            ${part}
+        DEPENDS
+            ${RAPIDWRIGHT_PATH}/scripts/invoke_rapidwright.sh
+    )
+
+    add_custom_target(rapidwright-${device}-device DEPENDS ${rapidwright_device_db})
+    set_property(TARGET rapidwright-${device}-device PROPERTY LOCATION ${rapidwright_device_db})
+endfunction()
+
+function(create_patched_device_db)
```
litghost, line 32: So rather than 1 function that applies two specific functions, please make a generate function for applying patches.  That function can be invoked for each patch that is required.  For example, US+ will need different patches than 7-series, etc.

fpga_interchange/examples/chipdb.cmake
```diff
@@ -0,0 +1,134 @@
...
+
+    get_property(constrained_luts_device_loc TARGET constraints-luts-${device}-device PROPERTY LOCATION)
+    set(chipdb_bba ${chipdb_dir}/chipdb-${device}.bba)
+    add_custom_command(
+        OUTPUT ${chipdb_bba}
+        COMMAND
+           ${PYTHON_EXECUTABLE} -mfpga_interchange.nextpnr_emit
+                --schema_dir ${INTERCHANGE_SCHEMA_PATH}
+                --output_dir ${CMAKE_CURRENT_BINARY_DIR}
+                --bel_bucket_seeds ${PYTHON_INTERCHANGE_PATH}/test_data/series7_bel_buckets.yaml
```
litghost, line 117: BEL bucket seeds should be an argument, as they will be architecture specific.

fpga_interchange/examples/tests.cmake
```diff
@@ -0,0 +1,115 @@
...
+        COMMAND
+            ${PYTHON_EXECUTABLE} -mfpga_interchange.yosys_json
+                --schema_dir ${INTERCHANGE_SCHEMA_PATH}
+                --device ${device_loc}
+                --top ${top}
+                ${synth_json}
+                ${netlist}
+        DEPENDS
+            ${synth_json}
+            ${device_target}
```
litghost, line 72: Should depend on both the target and the file.

fpga_interchange/examples/tests/const_wire/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
...
+    xdc wire.xdc
+    sources wire.v
+)
+
+add_interchange_test(
+    name const_wire_arty
+    device xc7a50t
+    package csg324
+    tcl run.tcl
+    xdc wire.xdc
```
litghost, line 15: These will need different `XDC` because the pin definitions are different.

fpga_interchange/examples/chipdb.cmake
acomodi: Done

fpga_interchange/examples/chipdb.cmake
acomodi: Done

fpga_interchange/examples/chipdb.cmake
acomodi: Done

fpga_interchange/examples/chipdb.cmake
acomodi: Done

fpga_interchange/examples/tests.cmake
acomodi: Done

fpga_interchange/examples/tests/const_wire/CMakeLists.txt
acomodi: fixed

fpga_interchange/examples/chipdb.cmake
```diff
@@ -0,0 +1,187 @@
...
+            ${PYTHON_EXECUTABLE} -mfpga_interchange.patch
+                --schema_dir ${INTERCHANGE_SCHEMA_PATH}
+                --schema device
+                --patch_path ${patch_path}
+                --patch_format ${patch_format}
+                ${input_device_loc}
+                ${patch_data}
+                ${output_device_file}
+        DEPENDS
+            ${input_device}
```
litghost, line 94: Should be depending on patch data?

fpga_interchange/examples/tests.cmake
```diff
@@ -0,0 +1,124 @@
...
+    )
+
+    add_custom_target(test-${family}-${name}-phys DEPENDS ${phys})
+
+    set(dcp ${CMAKE_CURRENT_BINARY_DIR}/${name}.dcp)
+    add_custom_command(
+        OUTPUT ${dcp}
+        COMMAND
+            RAPIDWRIGHT_PATH=${RAPIDWRIGHT_PATH}
+            ${RAPIDWRIGHT_PATH}/scripts/invoke_rapidwright.sh
```
Xiretza, line 111: (ditto on `${INVOKE_RAPIDWRIGHT}`)

fpga_interchange/examples/chipdb.cmake
```diff
@@ -0,0 +1,187 @@
...
+        COMMAND
+            RAPIDWRIGHT_PATH=${RAPIDWRIGHT_PATH}
+            ${INVOKE_RAPIDWRIGHT}
+            com.xilinx.rapidwright.interchange.DeviceResourcesExample
+            ${part}
+        DEPENDS
+            ${INVOKE_RAPIDWRIGHT}
+    )
+
+    add_custom_target(rapidwright-${device}-device DEPENDS ${rapidwright_device_db})
```
litghost, line 40: There needs to an option to see the output target name to a variable in parent scope.

fpga_interchange/examples/chipdb.cmake
```diff
@@ -0,0 +1,187 @@
...
+        DEPENDS
+            constraints-luts-${device}-device
+            ${constraints_luts_device_loc}
+    )
+
+    add_custom_target(chipdb-${device}-bba DEPENDS ${chipdb_bba})
+    add_dependencies(chipdb-${family}-bbas chipdb-${device}-bba)
+endfunction()
+
+set(chipdb_dir ${CMAKE_CURRENT_BINARY_DIR}/${family}/chipdb)
```
litghost, line 182: These lines do not belong here.  This file should only have CMake functions.

fpga_interchange/examples/chipdb.cmake
```diff
@@ -0,0 +1,187 @@
...
+                --device ${constraints_luts_device_loc}
+        COMMAND
+            mv ${CMAKE_CURRENT_BINARY_DIR}/chipdb.bba ${chipdb_bba}
+        DEPENDS
+            constraints-luts-${device}-device
+            ${constraints_luts_device_loc}
+    )
+
+    add_custom_target(chipdb-${device}-bba DEPENDS ${chipdb_bba})
+    add_dependencies(chipdb-${family}-bbas chipdb-${device}-bba)
```
litghost, line 179: Let's leave this out

fpga_interchange/examples/chipdb.cmake
```diff
@@ -0,0 +1,187 @@
...
+        DEPENDS
+            ${input_device}
+            ${input_device_loc}
+    )
+
+    add_custom_target(${patch_name}-${device}-device DEPENDS ${output_device_file})
+    set_property(TARGET ${patch_name}-${device}-device PROPERTY LOCATION ${output_device_file})
+endfunction()
+
+function(generate_chipdb)
```
litghost, line 102: This is specific to 7-series, so name as such?

litghost: I don't think this function is right.  I think generate chipdb should only be device database -> chipdb.bba

fpga_interchange/examples/chipdb.cmake
acomodi: fixed

fpga_interchange/examples/tests.cmake
acomodi: fixed

fpga_interchange/examples/chipdb.cmake
acomodi: I put these in `family.cmake`

fpga_interchange/examples/chipdb.cmake
acomodi: Separated in a xc7 specific (which calls RW and the patch funcs) and a generic generate_db which takes as input a device_target

fpga_interchange/examples/chipdb.cmake
```diff
@@ -0,0 +1,235 @@
...
+            ${part}
+        DEPENDS
+            ${INVOKE_RAPIDWRIGHT}
+    )
+
+    add_custom_target(rapidwright-${device}-device DEPENDS ${rapidwright_device_db})
+    set_property(TARGET rapidwright-${device}-device PROPERTY LOCATION ${rapidwright_device_db})
+
+    if (output_target)
+        set(output_target_name rapidwright-${device}-device PARENT_SCOPE)
```
litghost, line 49: I believe this should `set(${output_target} rapidwright-${device}-device PARENT_SCOPE)`?

fpga_interchange/examples/chipdb.cmake
```diff
@@ -0,0 +1,235 @@
...
+    #
+    # Generates a device database from RapidWright
+    #
+    # If output_target is specified, the output_target_name variable
+    # is set to the generated output_device_file target.
+    #
+    # Targets generated:
+    #   - rapidwright-<device>-device
+
+    set(options output_target)
```
litghost, line 18: `output_target` should be a 1 value arg.

fpga_interchange/examples/chipdb.cmake
acomodi: Done

fpga_interchange/examples/chipdb.cmake
acomodi: Removed

fpga_interchange/examples/chipdb.cmake
acomodi: Done

fpga_interchange/examples/chipdb.cmake
acomodi: Done

acomodi:
> could it be a mismatch between the structures in nextpnr and the bba generator?

@gatecat I missed to add the `bba -> bin` generation target with bbasm. With the `chipdb.bin` it worked fine.

I have added archcheck targets as well, both generic and with custom test_data

fpga_interchange/examples/tests/counter/remap.v
```diff
@@ -0,0 +1,11 @@
+module INV(input I, output O);
```
gatecat, line 1: should this be specific to `counter`? it looks more general than that. longer term of course this should be something that nextpnr can deal with (at a glance it looks like the interchange format could specify a rule like this already?)

acomodi: > should this be specific to counter?

Not sure about this actually, maybe @litghost can give some more info here.

On a side note though, I have just run the counter test and it is currently failing with the following:

```
ERROR: Failed to route arc 65.0 of net 'GLOBAL_LOGIC1', from $CONSTANTS_X0Y0.$CONSTANTS/$VCC_SOURCE to SLICE_X1Y37.SLICEL/A4.
```

Maybe for the time being we can avoid generating a target for this test.

litghost: > should this be specific to `counter`? it looks more general than that. longer term of course this should be something that nextpnr can deal with (at a glance it looks like the interchange format could specify a rule like this already?)

So there is a set of cells from Yosys that need what Vivado calls `retargeting`.  These are generally cells from older architecture libraries that don't exist in the new arches. This is a different process than macro expansion.  Another example of targeting is `FD` cell to `FDRE` cell.

Macro expansion only exists at the physical netlist level.  The logical netlist has the unexpanded macro (example `OBUFTDS` and `RAM64X1D`).  Re-targeting happens before logical netlist, and is a straight replacement.  Both `INV` -> `LUT1` and `FD` -> `FDRE` are examples of re-targetting.  Something that is important to note here is that re-targetting only happens on the unexpanded logical netlist. The `OBUFTDS` is an example of a macro primitive that contains 3 cells, one of which is an `INV`.  The `INV` that is part of the `OBUFTDS` macro can only be placed at `IOB33S/O_ININV`.

So to pop the stack, I believe re-targetting should be done prior to when creating the logical netlist (which is where I put it in the flow), which runs before nextpnr.  Macro expansion **must** be done within the place and route tool, because it is not visible at the logical netlist level at all, and so macro expansion would belong within nextpnr (but hasn't been done yet).

litghost: Examples of retargetting from Vivado:
```
test.bit
create_cell -ref INV test1
WARNING: [Coretcl 2-1024] Master cell 'INV' is not supported by the current part and has been retargeted to 'LUT1'.
test1
create_cell -ref BUF test2
WARNING: [Coretcl 2-1024] Master cell 'BUF' is not supported by the current part and has been retargeted to 'LUT1'.
test2
create_cell -ref FD test3
WARNING: [Coretcl 2-1024] Master cell 'FD' is not supported by the current part and has been retargeted to 'FDRE'.
test3
```

gatecat: Thanks for the explanation! I think this might be something we want in upstream Yosys, eventually.

litghost:
@acomodi Did I missing, but I think there is not README anymore?

acomodi:
> Did I missing it, but I think there is not README anymore?

@litghost The README of the fpga_interchange is still there actually (https://github.com/acomodi/nextpnr/tree/add-cmake-infra-fpga-interchange/fpga_interchange). What I deleted was the create_bba directory which contained the README to build the chipdb. Should I re-insert that one in the examples?

litghost:
> > Did I missing it, but I think there is not README anymore?
> 
> @litghost The README of the fpga_interchange is still there actually (https://github.com/acomodi/nextpnr/tree/add-cmake-infra-fpga-interchange/fpga_interchange). What I deleted was the create_bba directory which contained the README to build the chipdb. Should I re-insert that one in the examples?

But the instructions from that README are now gone?  There needs to be instructions to actually installing dependencies, etc.  Because that stuff happens before / outside of CMake.  To be clear, it is totally fine that the dependencies are installed outside of CMake, but there needs to be clear instructions on it.

litghost:
> > > Did I missing it, but I think there is not README anymore?
> > 
> > 
> > @litghost The README of the fpga_interchange is still there actually (https://github.com/acomodi/nextpnr/tree/add-cmake-infra-fpga-interchange/fpga_interchange). What I deleted was the create_bba directory which contained the README to build the chipdb. Should I re-insert that one in the examples?
> 
> But the instructions from that README are now gone? There needs to be instructions to actually installing dependencies, etc. Because that stuff happens before / outside of CMake. To be clear, it is totally fine that the dependencies are installed outside of CMake, but there needs to be clear instructions on it.

To be clear, what we need is some instructions on how to build the example designs from scratch on a clean install.

acomodi:
> To be clear, what we need is some instructions on how to build the example designs from scratch on a clean install.

@litghost I re-added the README.md with the instructions in the [examples directory](https://github.com/acomodi/nextpnr/tree/add-cmake-infra-fpga-interchange/fpga_interchange/examples)

fpga_interchange/examples/README.md
```diff
@@ -0,0 +1,67 @@
...
+Install python-fpga-interchange if not already installed:
+```
+git clone https://github.com/SymbiFlow/python-fpga-interchange.git
+cd python-fpga-interchange.git
+python -m pip install -e .
+```
+
+Clone RapidWright, if not already cloned:
+```
+git clone https://github.com/Xilinx/RapidWright.git
```
litghost, line 37: This is missing:
```
make update_jars
```

acomodi: Fixed

litghost:
@acomodi I think this looks pretty good.  Can you please add a basic GH actions job to build a wire and counter use the new CMake stuff?  I've asked @gatecat, and they indicated it would be ok.

acomodi:
@litghost @gatecat [GH action CI](https://github.com/acomodi/nextpnr/runs/2123162401) went green

.github/workflows/interchange_ci.yml
```diff
@@ -0,0 +1,27 @@
...
+      run: |
+        sudo apt-get update
+        sudo apt-get install git make cmake libboost-all-dev python3-dev libeigen3-dev tcl-dev clang bison flex swig
+
+    - name: Execute build script
+      run: stdbuf -i0 -o0 -e0 ./.github/ci/build_interchange.sh
+
+    - name: Execute test script
+      run: |
+        cd build && make all-fpga_interchange-tests -j`nproc`
```
litghost, line 27: Can you run the equivalent of `make -C create_bba test` in the new system?

litghost: Sorry, I mean `make -C archcheck check`

fpga_interchange/examples/tests.cmake
```diff
@@ -0,0 +1,125 @@
...
+            OUT_JSON=${synth_json}
+            TECHMAP=${techmap}
+            yosys -c ${tcl}
+        DEPENDS ${sources}
+    )
+
+    add_custom_target(test-${family}-${name}-json DEPENDS ${synth_json})
+
+    # Logical Netlist
+    set(device_target constraints-luts-${device}-device)
```
litghost, line 67: This `constraints-luts-` bits is not right?  Can you please revise?  You are assuming things that are equiv to 7-series flow, which is the kind of stuff what we do not want.

fpga_interchange/examples/tests.cmake
```diff
@@ -0,0 +1,125 @@
...
+        OUTPUT ${synth_json}
+        COMMAND
+            SOURCES=${sources}
+            OUT_JSON=${synth_json}
+            TECHMAP=${techmap}
+            yosys -c ${tcl}
+        DEPENDS ${sources}
+    )
+
+    add_custom_target(test-${family}-${name}-json DEPENDS ${synth_json})
```
litghost, line 64: Where did family variable come from?

litghost: In general no variables should be used like this unless they come from an argument to the function.

fpga_interchange/examples/tests.cmake
```diff
@@ -0,0 +1,125 @@
...
+
+    set(chipdb_target chipdb-${device}-bin)
+
+    # Physical Netlist
+    set(phys ${CMAKE_CURRENT_BINARY_DIR}/${name}.phys)
+    add_custom_command(
+        OUTPUT ${phys}
+        COMMAND
+            nextpnr-fpga_interchange
+                --chipdb ${chipdb_dir}/chipdb-${device}.bin
```
litghost, line 96: Rather than assume where the chipdb binary is located, why not get it from the chipdb_target, like you did with the device database?

fpga_interchange/examples/tests.cmake
litghost: I think `device_target` should probably be `device-${device}` and chipdb target should just be a property of the device target.

fpga_interchange/examples/tests.cmake
litghost: Ah, `family` was from the `nextpnr` root. Please add a comment to that affect.

fpga_interchange/examples/chipdb.cmake
```diff
@@ -0,0 +1,290 @@
...
+        device ${device}
+        patch_name constraints-luts
+        patch_path lutDefinitions
+        patch_format yaml
+        patch_data ${PYTHON_INTERCHANGE_PATH}/test_data/series7_luts.yaml
+        input_device ${constraints_device}
+        output_target constraints_luts_device
+    )
+
+    set(device_target ${constraints_luts_device} PARENT_SCOPE)
```
litghost, line 179: This should be an argument

litghost: Just like https://github.com/YosysHQ/nextpnr/pull/619/files#diff-96fd1153eb6fb79f848282c1f05b0a3e0f984461c4ec540413bac18f12d00a0fR116-R118

fpga_interchange/examples/tests.cmake
```diff
@@ -0,0 +1,125 @@
...
+            ${device_loc}
+    )
+
+    add_custom_target(test-${family}-${name}-netlist DEPENDS ${netlist})
+
+    set(chipdb_target chipdb-${device}-bin)
+
+    # Physical Netlist
+    set(phys ${CMAKE_CURRENT_BINARY_DIR}/${name}.phys)
+    add_custom_command(
```
litghost, line 92: Need targets for generating logical and physical netlist YAML.

fpga_interchange/examples/tests.cmake
litghost: I'd prefer we pass the family variable in.  See basically add a `oneValueArg`, `family`.

And then when invoking `add_interchange_test` add `family ${family}`.

fpga_interchange/family.cmake
```diff
@@ -5,8 +5,26 @@ endif()
...
+
 add_subdirectory(3rdparty/fpga-interchange-schema/cmake/cxx_static)
 
+include(${family}/examples/chipdb.cmake)
+include(${family}/examples/tests.cmake)
+
+set(chipdb_dir ${CMAKE_CURRENT_BINARY_DIR}/${family}/chipdb)
+file(MAKE_DIRECTORY ${chipdb_dir})
+
+set(root_dir ${CMAKE_CURRENT_SOURCE_DIR})
```
litghost, line 22: Why is `root_dir` needed?  This seems like a bad idea?

fpga_interchange/examples/chipdb.cmake
```diff
@@ -0,0 +1,290 @@
...
+    add_custom_command(
+        OUTPUT ${chipdb_bba}
+        COMMAND
+            ${PYTHON_EXECUTABLE} -mfpga_interchange.nextpnr_emit
+                --schema_dir ${INTERCHANGE_SCHEMA_PATH}
+                --output_dir ${CMAKE_CURRENT_BINARY_DIR}
+                --bel_bucket_seeds ${bel_bucket_seeds}
+                --device ${device_loc}
+        COMMAND
+            mv ${CMAKE_CURRENT_BINARY_DIR}/chipdb.bba ${chipdb_bba}
```
litghost, line 232: This is a dangerous solution, because in a parallel build this could result in a file collision.  I recommend we run `nextpnr_emit` in a unique working directory and use it in place, rather than moving it.

fpga_interchange/family.cmake
acomodi: This was only required to get the location of the `check_arch_api.py` for the archcheck target. Maybe I can find another way to get that script

fpga_interchange/examples/tests.cmake
acomodi: Given that the family will alwasy be fpga_interchange (if we can assume this), might be worth avoiding adding additional arguments and have targets like `test-fpga-interchange-${name}-json`

fpga_interchange/family.cmake
litghost: In that case, you should be using `${PROJECT_SOURCE_DIR}` or `${nextpnr_SOURCE_DIR}`.

fpga_interchange/examples/tests.cmake
litghost: In the long run, we are probably going to want other roll up targets.  For example, all A7 tests, all Z7 tests, all Nexus tests, all Zynq US+ tests, etc.  If you make the `family` argument explicit, there is an easier attachment point.

fpga_interchange/examples/chipdb.cmake
acomodi: Done. Each device already had its own directory, so now, the bba will just be called chipdb.bba

fpga_interchange/family.cmake
acomodi: Done

fpga_interchange/examples/tests.cmake
acomodi: Done

fpga_interchange/examples/chipdb.cmake
acomodi: Done

fpga_interchange/examples/tests.cmake
acomodi: Done. I have added a target which contains info on:

- Device location
- Device target
- Chipdb bin location
- Chipdb bin target

The "container" target is `device-${device}`

fpga_interchange/examples/tests.cmake
acomodi: All right, done

.github/workflows/interchange_ci.yml
acomodi: Sure, I have added an `all-${family}-archcheck-tests` target to accumulate all the archcheck tests dependencies.

fpga_interchange/examples/chipdb.cmake
```diff
@@ -0,0 +1,314 @@
...
+        device-${device}
+        PROPERTIES
+            DEVICE_LOC ${device_loc}
+            DEVICE_TARGET ${device_target}
+            CHIPDB_BIN_LOC ${chipdb_bin}
+            CHIPDB_BIN_TARGET chipdb-${device}-bin
+    )
+
+    # Generate architecture check target
+    set(test_data_source ${CMAKE_CURRENT_SOURCE_DIR}/test_data.yaml)
```
litghost, line 275: Why is this symlink needed?  I believe this will prevent in-source builds from working properly.

fpga_interchange/examples/devices/xc7a35t/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+generate_xc7_device_db(
+    device xc7a35t
+    part xc7a35tcsg324-1
+    device_target target
```
litghost, line 4: Please name the output variable better than `target`.  Maybe `xc7a35t_device`?

fpga_interchange/examples/devices/xc7a35t/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+)
+
+generate_chipdb(
+    family ${family}
+    device xc7a35t
+    part xc7a35tcsg324-1
+    device_target ${target}
+    bel_bucket_seeds ${PYTHON_INTERCHANGE_PATH}/test_data/series7_bel_buckets.yaml
+    package csg324
```
litghost, line 13: The package here is only used for testing the chipdb.  Maybe rename the argument to "test_package" and mention this in the README.  The key here is that this chipdb has more packages than just csg324, and this looks confusing right now.

fpga_interchange/examples/chipdb.cmake
```diff
@@ -0,0 +1,314 @@
...
+
+    if(DEFINED device_target)
+        set(${device_target} ${constraints_luts_device} PARENT_SCOPE)
+    endif()
+endfunction()
+
+function(generate_chipdb)
+    # ~~~
+    # generate_chipdb(
+    #    family <family>
```
litghost, line 189: Please add short comments describing each argument to all of these CMake functions.

fpga_interchange/examples/tests.cmake
```diff
@@ -0,0 +1,164 @@
+function(add_interchange_test)
+    # ~~~
+    # add_interchange_test(
+    #    name <name>
+    #    family <family>
+    #    device <common device>
+    #    package <package>
+    #    tcl <tcl>
+    #    xdc <xdc>
+    #    top <top name>
```
litghost, line 10: `top` is optional, should be in `[]`?

fpga_interchange/examples/tests.cmake
```diff
@@ -0,0 +1,164 @@
...
+        OUTPUT ${dcp}
+        COMMAND
+            RAPIDWRIGHT_PATH=${RAPIDWRIGHT_PATH}
+            ${INVOKE_RAPIDWRIGHT}
+            com.xilinx.rapidwright.interchange.PhysicalNetlistToDcp
+            ${netlist} ${phys} ${xdc} ${dcp}
+        DEPENDS
+            ${INVOKE_RAPIDWRIGHT}
+            ${phys}
+            ${netlist}
```
litghost, line 159: Should also depend on XDC file?

litghost:
@gatecat I believe this PR is ready for another review from you.

fpga_interchange/examples/chipdb.cmake
acomodi: This is required as otherwise we get:

```
Traceback (most recent call last):
  File "/data/interchange/nextpnr/python/check_arch_api.py", line 90, in <module>
    check_arch_api(ctx)
  File "/data/interchange/nextpnr/python/check_arch_api.py", line 64, in check_arch_api
    with open('test_data.yaml', 'r') as f:
FileNotFoundError: [Errno 2] No such file or directory: 'test_data.yaml'
ERROR: Error occurred while executing Python script /data/interchange/nextpnr/python/check_arch_api.py
0 warnings, 1 error
fpga_interchange/examples/devices/xc7a35t/CMakeFiles/chipdb-xc7a35t-bin-check-test-data.dir/build.make:57: recipe for target 'fpga_interchange/examples/devices/xc7a35t/CMakeFiles/chipdb-xc7a35t-bin-check-test-data' failed
```

test_data.yaml needs to reside in the build directory, hence the symlink

fpga_interchange/examples/devices/xc7a35t/CMakeLists.txt
acomodi: Sure, done

fpga_interchange/examples/chipdb.cmake
litghost: Just have the working directory of the command in the source directory where the tile is?

fpga_interchange/examples/devices/xc7a35t/CMakeLists.txt
acomodi: Done

fpga_interchange/examples/chipdb.cmake
acomodi: Done

fpga_interchange/examples/tests.cmake
acomodi: Done

fpga_interchange/examples/tests.cmake
acomodi: Yes, done

fpga_interchange/examples/chipdb.cmake
acomodi: Right, done

fpga_interchange/examples/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+```
+mkdir build
+cd build
+cmake .. --DARCH=fpga_interchange -DRAPIDWRIGHT_PATH=<RapidWright path> -DINTERCHANGE_SCHEMA_PATH=<fpga-interchange-schema path> -DPYTHON_INTERCHANGE_PATH=<python-fpga-interchange path>
+```
+
+To build the xc7a35t architecture, run:
+```
+make chipdb-xc7a35t-bin
```
litghost, line 56: @acomodi Is this the right target name now?

litghost (approved):
LGTM, assuming GH action build is green.  @acomodi Please link to green build once complete.

fpga_interchange/examples/README.md
acomodi: Yes, this is still correct. This target is stored as property of the `device-${device}` target now, alongside with the file location and also the interchange device (target and location as well).

litghost:
CI failed, but this is a bug that I've been working on: 
```
ERROR: post-placement validity check failed for Bel 'SLICE_X1Y22.SLICEL/B5FF' (cell '$auto$simplemap.cc:527:simplemap_adff_sdff$1506')
ERROR: Placing design failed.
0 warnings, 2 errors
make[3]: *** [fpga_interchange/examples/tests/counter/CMakeFiles/test-fpga_interchange-counter_basys3-dcp.dir/build.make:88: fpga_interchange/examples/tests/counter/counter_basys3.phys] Error 255
```

acomodi:
> CI failed, but this is a bug that I've been working on:

I have temporarily disabled the counter test so that we have a green CI

PR #620, opened 2021-03-12
Add support for partially routed nets from the placer in router2.

litghost:


gatecat:
CI got blown out by GitHub issues earlier but I've triggered a rebuild; otherwise LGTM

PR #621, opened 2021-03-12
Split nextpnr.h to allow for linear inclusion.

litghost:
"nextpnr.h" is no longer the god header.  Important improvements:

 - Functions in log.h can be used without including
   BaseCtx/Arch/Context. This means that log_X functions can be called
   without included "nextpnr.h"

 - NPNR_ASSERT can be used without including "nextpnr.h" by including
   "nextpnr_assertions.h".  This allows NPNR_ASSERT to be used safely in
   any header file.

 - Types defined in "archdefs.h" are now available without including
   BaseCtx/Arch/Context.  This means that utility classes that will be
   used inside of BaseCtx/Arch/Context can be defined safely in a
   self-contained header.


gatecat:
Thanks, this looks like a great improvement. I will give it a proper review over the weekend.

litghost:
> Thanks, this looks like a great improvement. I will give it a proper review over the weekend.

I suspect I'll need to fixup many of the arches.

litghost:
> > Thanks, this looks like a great improvement. I will give it a proper review over the weekend.
> 
> I suspect I'll need to fixup many of the arches.

Build is passing now.

common/deterministic_rng.h
```diff
@@ -0,0 +1,103 @@
...
+ *  WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ *  MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ *  ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ *  WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ *  ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ *  OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ *
+ */
+
+#include <algorithm>
```
gatecat, line 21: nitpick but I think these should be after the include guard

gatecat:
I think with this PR places using `namespace nextpnr` like bits.h can now go back to `NEXTPNR_NAMESPACE_BEGIN` for consistency

litghost:
> I think with this PR places using `namespace nextpnr` like bits.h can now go back to `NEXTPNR_NAMESPACE_BEGIN` for consistency

I agree, but I'd like to fix that in a follow up PR, rather than in this one.

common/deterministic_rng.h
litghost: Will fix.

gatecat:
> I agree, but I'd like to fix that in a follow up PR, rather than in this one.

No problem, just checking it wasn't forgotten about.

common/deterministic_rng.h
litghost: Fixed

PR #626, opened 2021-03-16
Add missing includes to fix WASI build

whitequark:
These were omitted in #621 (some because they are not exercised in non-WASI builds, some because libstdc++ is not very hygienic with headers).


gatecat:
Thanks!

PR #627, opened 2021-03-16
WIP/DNM: Murax-capable FPGA interchange WIP PR

litghost:
This PR has the current progress to get FPGA interchange to P&R murax/picosoc class designs.

This PR adds:
 - Generic lookahead that will work for timing and non-timing driven architectures
 - A completely re-written site router that should be robust
 - Pseudo pips are now handled in the `Arch` based on changes from https://github.com/SymbiFlow/python-fpga-interchange/pull/32
 - Routing chosen by site router is now put into the `Arch` prior to running the general router to avoid router1 or router2 having to de-congest pseudo pips.  Because site routing is fixed during general routing, `checkPipAvail` correct marks that pseudo pips are available or unavailable.
 - Fixed some latent bugs for BRAM's and parameter driven cell -> BEL mapping

Known remaining issues:
 - [x] Some pseudo pip data is incorrect (see https://github.com/Xilinx/RapidWright/issues/138) resulting in valid designs being rejected
 - [ ] Lookahead data is excessive redundant, resulting in a larger memory footprint than is reasonable
 - [x] Lookahead data should be serialized and re-used rather than recomputed every time
 - [ ] LUT rotation in site router is consuming excessive amount of time (~45% of detailed site routing check on murax).  This needs to be optimized
 - [ ] Unused cell ports currently generate warnings, even if those nets are just tied to GND/VCC
 - [x] Site router will allow VCC nets to enter site ports for GND signals, and vise-versa.  This will result in unroutable designs if the site router does this
 - [x] Site router does not understand inverting pips, and may select an inverted route, which will change the design incorrectly.

Not all issues need to be fixed in order to get murax/picosoc working.  Incorrect pseudo pip data is likely the only remaining blocking issue, but until that is cleared, it is hard to tell how many of the others will be problematic.


fpga_interchange/arch.cc
```diff
@@ -895,6 +1011,28 @@ const std::vector<std::string> Arch::availablePlacers = {"sa",
...
 const std::vector<std::string> Arch::availableRouters = {"router1", "router2"};
 
+bool parameter_compare(const std::string &database_param, const std::string &netlist_param)
+{
+
+    bool result = false;
+    try {
+        size_t idx;
+        int database_value = std::stoi(database_param, &idx, 0);
+        int netlist_value = std::stoi(netlist_param, &idx, 2);
```
gatecat, line 1021: this looks a bit fragile to me, but I guess it's OK for a first go

fpga_interchange/arch.h
```diff
@@ -1303,29 +573,33 @@ struct Arch : ArchAPI<ArchRanges>
...
         get_tile_loc(pip.tile, &loc);
         loc.z = 0;
         return loc;
     }
 
-    uint32_t getPipChecksum(PipId pip) const override { return pip.index; }
+    uint32_t getPipChecksum(PipId pip) const final { return pip.index; }
 
-    WireId getPipSrcWire(PipId pip) const override
+    WireId getPipSrcWire(PipId pip) const final NPNR_ATTRIBUTE(__always_inline__)
```
gatecat, line 586: can you create a `NPNR_ALWAYS_INLINE` macro? that way someone can do a MSVC version if they care enough

fpga_interchange/examples/murax/Murax.v
```diff
@@ -0,0 +1,4558 @@
+// Generator : SpinalHDL v1.1.5    git head : 0310b2489a097f2b9de5535e02192d9ddd2764ae
```
gatecat, line 1: Not sure if it might be better to have Murax located externally?

fpga_interchange/sampler.h
```diff
@@ -0,0 +1,67 @@
...
+
+#ifndef SAMPLER_H_
+#define SAMPLER_H_
+
+#include <cstdint>
+#include <functional>
+#include <stdexcept>
+#include <vector>
+
+namespace nextpnr {
```
gatecat, line 29: this should be using NEXTPNR_NAMESPACE_BEGIN

fpga_interchange/lookahead.cc
```diff
@@ -0,0 +1,1195 @@
...
+    absl::flat_hash_set<TypeWireId> types_deferred;
+
+    // Walk each tile type, and expand all non-site wires in the tile.
+    // Wires that are nodes will expand as if the node type is the first node
+    // in the wire.
+    //
+    // Wires that only have 1 output pip are deferred until the next loop,
+    // because generally those wires will get explored via another wire.
+    // The deferred will be expanded if this assumption doesn't hold.
+#if defined(NEXTPNR_USE_TBB) // Run parallely
```
gatecat, line 723: the slightly intricate use of macros here is going to make it very easily to accidentally break one of the TBB/non-TBB builds, I'm not sure if there is a good solution for this and ideally we test both in CI eventually

fpga_interchange/lookahead.cc
```diff
@@ -0,0 +1,1195 @@
...
+        expand_deferred_routing_graph(ctx, rng, tile_sampler, wire_type, &types_explored, &all_tiles_storage,
+                                      &best_path);
+    }
+
+    auto end = std::chrono::high_resolution_clock::now();
+    if (ctx->verbose) {
+        log_info("Done with expansion, dt %02fs\n", std::chrono::duration<float>(end - start).count());
+    }
+
+    FILE *lookahead_data = fopen("lookahead.csv", "w");
```
gatecat, line 878: It looks like this is unconditionally writing a file in the current directory? this should probably be behind some kind of conditional

fpga_interchange/site_arch.cc
```diff
@@ -0,0 +1,415 @@
...
+            // Because this net is driven from a source out of the site,
+            // no out of site sink is required.
+            continue;
+        }
+
+        // Examine net to determine if it has any users not in this site.
+        bool net_used_out_of_site = false;
+        for (const PortRef &user : net->users) {
+            if (user.cell == nullptr) {
+                // This is pretty weird!
```
gatecat, line 168: is this a case you are seeing? I don't think it's something that should ever be happening

fpga_interchange/site_arch.h
```diff
@@ -0,0 +1,990 @@
...
+            dst_net->wires.erase(iter);
+        }
+    }
+
+    void archcheck();
+
+    bool is_pip_synthetic(const SitePip &pip) const NPNR_ATTRIBUTE(__always_inline__);
+};
+
+struct SitePipDownhillIterator
```
gatecat, line 455: It seems at a glance that there's quite a bit of duplication between these iterators that could be tidied up?

fpga_interchange/cost_map.cc
```diff
@@ -0,0 +1,356 @@
...
+    }
+
+    const auto &delay_matrix = iter->second;
+
+    int32_t off_x = delay_matrix.offset.first + (dst_x - src_x);
+    int32_t off_y = delay_matrix.offset.second + (dst_y - src_y);
+
+    int32_t closest_x = off_x;
+    int32_t closest_y = off_y;
+    if (closest_x < 0) {
```
gatecat, line 101: using `std::max` and `std::min` would be neater here

fpga_interchange/site_routing_cache.h
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    static SiteRoutingKey make(const SiteArch *ctx, const SiteNetInfo &site_net);
+};
+
+NEXTPNR_NAMESPACE_END
+
+template <> struct std::hash<NEXTPNR_NAMESPACE_PREFIX SiteRoutingKey>
+{
+    std::size_t operator()(const NEXTPNR_NAMESPACE_PREFIX SiteRoutingKey &key) const noexcept
```
gatecat, line 97: could this use a precomputed hash for better performance like `TypeWireSet` ?

fpga_interchange/sampler.cc
```diff
@@ -0,0 +1,174 @@
...
+ */
+
+#include "sampler.h"
+#include <algorithm>
+#include <cmath>
+#include <stdexcept>
+
+namespace nextpnr {
+
+static size_t partition_x(std::vector<size_t>::iterator begin, std::vector<size_t>::iterator end,
```
gatecat, line 28: a single `partition` function with an `axis` argument would reduce duplication between `partition_x` and `partition_y`

fpga_interchange/site_routing_cache.h
litghost: Believe or not, once I implemented the site routing cache, this was no longer in the hot path.  Remaining `isBelLocationValid` hot path is (values from murax run):
 - is_cell_valid_constraints: 11.84s
 - remap_luts: 87s
 - route_site: 33.18s
 - SiteArch: 16.56s

Once I fix the remaining issues preventing the design from getting to a DCP, I have good ideas on how to improve most of those numbers.

fpga_interchange/site_arch.cc
litghost: I'll change this to a NPNR_ASSERT then.  The problem with places where `Type *` appear, you can never be sure if a nullptr is allowed, or what.

fpga_interchange/lookahead.cc
litghost: Oh, this can be deleted.  This was a debugging logic I left in to visualize the output.

fpga_interchange/sampler.h
litghost: Yep

fpga_interchange/examples/murax/Murax.v
litghost: 100% agree.

fpga_interchange/arch.cc
litghost: Exactly.  Parameter handling probably needs a lot of polish once the big fish are fried.

fpga_interchange/arch.h
litghost: Will do.

fpga_interchange/site_arch.h
litghost: So `SitePipDownhillIterator` is basically the hottest part of the site router expansion loop, and it is very sensitivity.  I'm not sure how to both reduce the duplication, but also not increase the runtime.  For better or worse, I think `SitePipDownhillIterator` needs to be seperate from the other two iterators because of how critical path it ends up being to the site router.

`PipDownhillIterator` is in a similiar boat, but I haven't taken the time to optimize it as much.

fpga_interchange/sampler.h
litghost: Fixed

fpga_interchange/arch.h
litghost: Fixed

fpga_interchange/lookahead.cc
litghost: I should refactor this to minimize the differences between the `#if defined` stuff

fpga_interchange/site_arch.cc
litghost: Fixed.

fpga_interchange/cost_map.cc
litghost: Fixed

fpga_interchange/lookahead.cc
litghost: I've moved this to a function, but disabled it for now.

fpga_interchange/lookahead.cc
litghost: I've refactored this so that most code compiles in both paths, and the differences are small and fairly local.

litghost:
I was able to get a variant on this PR to a DCP, and have to fix the errors found there.  However I'm getting pretty close.

litghost:
I have a DCP that only has two remaining issues, both related to BRAM:
 - Parameter values don't match what Vivado is expecting, e.g. I need to fix https://github.com/SymbiFlow/fpga-interchange-schema/issues/12 now :(
 - Constant routing for BRAM is missing

I'll likely start splitting up the changes in the PR so we can get them merged as I work in parallel to address the issues listed above.  I know how to fix both of the above issues, so at this point is just a matter of time.

litghost:
Remaining known issues for murax:
 - [ ] Merge parameter updates (https://github.com/Xilinx/RapidWright/pull/141 / https://github.com/SymbiFlow/python-fpga-interchange/pull/37 / https://github.com/YosysHQ/nextpnr/pull/639) and test 
 - [ ] Handle "best" constant routing.  This will require a DB update with site pip inversion metadata
 - [ ] Need to block invalid LUT route-thru's

PR #628, opened 2021-03-17
fpga_interchange: add more devices

acomodi:

This PR adds some additional devices apart from the 35T.

Locally, the wire test fails during routing for the z010 and a200t with the following:

```
ERROR: Failed to route arc 0.0 of net '$iopadmap$i', from IOB_X0Y61.IOB33/INBUF_EN_OUT to IOB_X0Y54.IOB33/OUSED_OUT.
```

Apart from the routing issue, the following would need to be done:

- [x] Add test_data for the added archs
- [x] Enable new archs for each test (we might have a new CMake function to reduce code-duplication for each new similar test)


acomodi:
@litghost How the `test_data.yml` is written/generated? I am trying to add one for the z010 device

litghost:
> @litghost How the `test_data.yml` is written/generated? I am trying to add one for the z010 device

By hand.

fpga_interchange/examples/devices/xc7a200t/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+    part xc7a200tsbg484-1
+    device_target xc7a200t_target
+)
+
+generate_chipdb(
+    family ${family}
+    device xc7a200t
+    part xc7a200tsbg484-1
+    device_target ${xc7a200t_target}
+    bel_bucket_seeds ${PYTHON_INTERCHANGE_PATH}/test_data/series7_bel_buckets.yaml
```
litghost, line 12: This needs to be updated now that #644 merged, see https://github.com/YosysHQ/nextpnr/pull/644/commits/720f64ea6017fa3e42f80eba64cc615c64788914#diff-e8ebf6961ee292d34a9c55086f82687a83189a2846de6c4a1102f391c8a27168R12

acomodi: Fixed

acomodi:
@litghost I have added an `add_board` function so that tests can be easier added in one function call only, instead of multiple times. I believe this is ready to review.

litghost (approved):
Looks pretty good.  Let's see what happens with the CI.

litghost:
@acomodi Looks like runtime is spiking.  I think we want to split the jobs?  Do you know if there is a way to share the build products, and split on part type?

acomodi:
> @acomodi Looks like runtime is spiking. I think we want to split the jobs? Do you know if there is a way to share the build products, and split on part type?

@litghost Yeah, even though CI completed, it took quite some time (~1h20m), so I think that, if doable, splitting jobs across different devices makes sense. I'll see what I can do

gatecat:
75m for the xc7a200t is a pretty long time for feedback, it would definitely be nice to see if we can speed things up so it doesn't exceed the 50 odd minutes that Cirrus is taking for the main CI (which I'd like to speed up too long term, 30 minutes being the goal)

litghost:
> 75m for the xc7a200t is a pretty long time for feedback, it would definitely be nice to see if we can speed things up so it doesn't exceed the 50 odd minutes that Cirrus is taking for the main CI (which I'd like to speed up too long term, 30 minutes being the goal)

@acomodi I think what we want to do here is a forked build (if possible).  Something like:
 - In parallel:
   - Build C/C++ dependencies
   - Run invoke RapidWright, patch db and generate bba (text) (for each device)
 - Wait for previous jobs to finish
 - In parallel:
  - Run each device using products from earlier steps

This may also be a candidate for using ccache, if it isn't too bad.

litghost:
Also we are building `nextpnr-fpga_interchange` for each device, which is repeated/wasted work.  Again, ccache may help here:

It's currently taking ~21 minutes to build `nextpnr-fpga_interchange`, which seems high, but that is probably down to have a small number of cores.

litghost:
https://github.com/actions/cache might be useful here

acomodi:
Indeed, I am trying to set up caching so that we do not have to rebuild. I am looking at https://github.com/actions/cache which might be of help in this case.

I bumped into some issues about passing dependencies over other jobs.

litghost:
Google also lead to https://github.com/hendrikmuhs/ccache-action

acomodi:
@litghost @gatecat After some experiments I came up with the solution in https://github.com/YosysHQ/nextpnr/pull/628/commits/96448ace817a8042153734b8ba2da7b0410bbdb3 which uses ccache.

Basically the workflow is divided in 2 initial jobs which build independent tooling:
- Yosys
- Capnp and nextpnr-fpga_interchange

Than, one job for each device is spawned and the tools are re-built, using the ccache therefore taking a few minutes to compile.

In following CI runs, also the first 2 initial jobs should compile much faster using pre-existing ccache.

> Run invoke RapidWright, patch db and generate bba (text) (for each device)

This is not straightforward to do as we would need to store the artifacts of this job and download them in the Run-Tests job, as GH actions does not preserve the environment between one job to the other.
I might find another solution for this but at the moment I managed to cut the build times by ~half.

gatecat (approved):
ccache changes LGTM, subject to a final green CI

acomodi:
> subject to a final green CI

@gatecat even with a green CI it may be worth to re-run the jobs, to see if the ccache is doing the right job.

litghost:
> @litghost @gatecat After some experiments I came up with the solution in [3f9b52f](https://github.com/YosysHQ/nextpnr/commit/3f9b52fef4e0f92202655bd7693937919aa81915) which uses ccache.

This looks really good.  A200T finished in 42 minutes.  Let's see how it does with ccache!

acomodi:
Here's the latest CI which has ccache filled from previous runs https://github.com/YosysHQ/nextpnr/pull/628/checks?check_run_id=2194459871

litghost:
Build time dropped from 20m to 7m with ccache, which is a pretty nice win.  Suspect A200T will be done in ~20m.  Not bad as a starting place.

litghost:
@acomodi There is a small parallelism limitation right now in that we serialize the arch-check stuff and the example building.  As a follow on PR, can you create a target that depends on the two targets we explicitly invoke, so that the build system will do the arch check and examples in parallel?  It will save another couple minutes.  Definitely do this change in a new PR, this one is good to merge as is.

PR #631, opened 2021-03-18
Update root CMake with some additional features

litghost:
- Allow `gui_${family}` library to be linked against libraries that expose headers.  This is important if `archdefs`/etc use headers
 - Enable IPO on builds
 - Allow optional linking against `libprofiler`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,6 +18,7 @@ option(COVERAGE "Add code coverage info" OFF)
 option(STATIC_BUILD "Create static build" OFF)
 option(EXTERNAL_CHIPDB "Create build with pre-built chipdb binaries" OFF)
 option(WERROR "pass -Werror to compiler (used for CI)" OFF)
+option(PROFILER "Link against libprofiler" OFF)
 
 set(CMAKE_INTERPROCEDURAL_OPTIMIZATION TRUE)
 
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -1,6 +1,14 @@
 cmake_minimum_required(VERSION 3.5)
 project(nextpnr CXX C)
 
+# Allow family.cmake add additional dependencies to gui_${family}.
+cmake_policy(SET CMP0079 NEW)
+
+# Enable IPO support.
```
gatecat, line 7: Is IPO intended to be part of this patch?

litghost: I can remove it.  There is also some linking against `libprofiler` in here too.  I can seperate it out.

litghost: I was planning on adding it.

litghost: I've split up the commits with clear commit descriptions, and updated the PR description and title to avoid confusion.

gatecat: Thanks!

whitequark:
> * Enable IPO on builds

This PR does not just *enable* IPO on builds. It *requires* IPO support or [fails the entire build](https://github.com/YoWASP/nextpnr/runs/2144539713?check_suite_focus=true). (I think the error I linked to is actually a CMake bug, but do you really want to completely forbid nextpnr builds without IPO?)

PR #632, opened 2021-03-18
Add "checkPipAvailForNet" to Arch API.

litghost:
This is important for distinguishing valid pseudo pips in the FPGA
interchange arch. This also avoids a double or triple lookup of
pip->net map.

### DIFF
--- a/docs/archapi.md
+++ b/docs/archapi.md
@@ -398,6 +398,13 @@ pip to a net.
 
 *BaseArch default: returns `getBoundPipNet(pip) == nullptr`*
 
+### bool checkPipAvailForNet(PipId pip, NetInfo *net) const
+
+Returns true if the given pip is available to be bound to a net, or if the
+pip is already bound to that net.
+
+*BaseArch default: returns `getBoundPipNet(pip) == nullptr || getBoundPipNet(pip) == net`*
+
 ### NetInfo \*getBoundPipNet(PipId pip) const
 
 Return the net this pip is bound to.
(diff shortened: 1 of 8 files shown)


gatecat:
LGTM other than the CI failure. I think this API might find use in guiding global routing away from using excessive local routing resources, too.

litghost:
> LGTM other than the CI failure. I think this API might find use in guiding global routing away from using excessive local routing resources, too.

I think I've fixed the CI failure, let's see if anything else crops up.  Probably good to run your benchmarks and make sure this either has no impact or improves designs that use router2.

gatecat:
Looks like the FPGA interchange CI is failing now - I'll check the benchmarks over the weekend.

litghost:
> Looks like the FPGA interchange CI is failing now - I'll check the benchmarks over the weekend.

I think I know why.  I've replicated the failure locally.  The branch derived from https://github.com/YosysHQ/nextpnr/pull/627 isn't failing, so I likely know what the problem is.

litghost:
> > Looks like the FPGA interchange CI is failing now - I'll check the benchmarks over the weekend.
> 
> I think I know why. I've replicated the failure locally. The branch derived from #627 isn't failing, so I likely know what the problem is.

I’ve fixed the problem, CI is green.

docs/archapi.md
```diff
@@ -398,6 +398,13 @@ pip to a net.
 
 *BaseArch default: returns `getBoundPipNet(pip) == nullptr`*
 
+### bool checkPipAvailForNet(PipId pip, NetInfo *net) const
+
+Returns true if the given pip is available to be bound to a net, or is if the
```
gatecat, line 403: nitpick - should be `if` not `is if`

litghost: Fixed

gatecat (approved):

PR #633, opened 2021-03-19
cmake: Use IPO only if supported

gatecat:
Based on the CMake docs:

> Set <result> variable to YES if IPO is supported by the compiler and NO otherwise. If this option is not given then the command will issue a fatal error if IPO is not supported.

I think this should fix the WASM issue but haven't actually tested (is there an easy way to test locally?)


whitequark:
> is there an easy way to test locally?

I couldn't actually reproduce it locally yet, probably because CI uses cmake 3.19 and I only have cmake 3.18 locally.

whitequark (approved):
Thanks!

PR #634, opened 2021-03-20
Add getBelPinType to Python interface.

litghost:


common/arch_pybindings_shared.h
```diff
@@ -60,6 +60,8 @@ fn_wrapper_0a<Context, decltype(&Context::getBels), &Context::getBels, wrap_cont
 
 fn_wrapper_2a<Context, decltype(&Context::getBelPinWire), &Context::getBelPinWire, conv_to_str<WireId>,
               conv_from_str<BelId>, conv_from_str<IdString>>::def_wrap(ctx_cls, "getBelPinWire");
+fn_wrapper_2a<Context, decltype(&Context::getBelPinType), &Context::getBelPinType, conv_to_str<PortType>,
```
gatecat, line 63: we should be mapping `PortType` to a Python enum already: https://github.com/YosysHQ/nextpnr/blob/master/common/pybindings.cc#L127-L131

so I think `pass_through` would be enough with no need for the string conversion stuff

litghost: Right!  I'll test that now.

litghost: Yep, that worked as expected.  PR has been updated.

PR #636, opened 2021-03-20
Add pseudo pip data

litghost:
PR is based on #635, part 2 of many for breaking up PR #627.

This is relatively small delta from #635, mostly just a chip db schema update and some logic to start using that new pseudo pip data.

PR #637, opened 2021-03-20
Refine site router

litghost:
Builds on #636.  PR 3 of many splitting up #627.  This is a more significant change than the previous two PRs.

The new site router should be robust to most situations, and isn't significantly slower with the use of caching.


fpga_interchange/site_router.cc
```diff
@@ -39,669 +49,597 @@ void SiteRouter::unbindBel(CellInfo *cell)
...
+                     std::vector<SitePip>::const_iterator pips_end)
+{
+    for (auto iter = pips_begin; iter != pips_end; ++iter) {
+        ctx->unbindPip(*iter);
+    }
+}
 
-        std::vector<Node> nodes_to_expand;
-        nodes_to_expand.push_back(node);
+bool find_solution_via_backtrack(SiteArch *ctx, std::vector<PossibleSolutions> solutions,
```
gatecat, line 402: are `solutions` and `sinks_to_solutions` being passed-by-value rather than const references for a reason?

fpga_interchange/site_routing_storage.h
```diff
@@ -0,0 +1,153 @@
...
+        return Node(const_cast<RouteNodeStorage *>(this), idx);
+    }
+
+    // Return all node from the current owner to the free list.
+    void free_nodes(std::vector<size_t> &other_free_list)
+    {
+        free_list.insert(free_list.end(), other_free_list.begin(), other_free_list.end());
+
+        // Hand out nodes in ascending order.
+        std::sort(free_list.rbegin(), free_list.rbegin());
```
gatecat, line 132: is `rbegin()` twice here correct?

litghost: Nope!  Good eye.

fpga_interchange/site_router.cc
litghost: Good catch, let me think through what is going on.  I think this was not causing terrible performance before because this was a tail call, but that is not a reason to leave it like this.

fpga_interchange/site_routing_storage.h
litghost: Fixed.  Given that sorting `[begin, begin)` is a no-op, I'll remove this for now.

fpga_interchange/site_router.cc
litghost: Fixed.

PR #638, opened 2021-03-20
Correct some bugs in writing of physical netlist w.r.t. site sources.

litghost:
Local site sources should have their driving BEL pin included in the net
so that the site wire is driven by an output BEL pin.

Also handle cases where a cell pin is missing from the cell -> BEL map.  This is common, but weird.  For now generate a warning.  Improve logic with further iteration.

### DIFF
--- a/fpga_interchange/fpga_interchange.cpp
+++ b/fpga_interchange/fpga_interchange.cpp
@@ -114,14 +114,12 @@ static PhysicalNetlist::PhysNetlist::RouteBranch::Builder emit_branch(
             WireId src_wire = ctx->getPipSrcWire(pip);
             WireId dst_wire = ctx->getPipDstWire(pip);
 
-            IdString src_pin;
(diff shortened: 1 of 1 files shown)


fpga_interchange/fpga_interchange.cpp
```diff
@@ -345,6 +410,15 @@ void FpgaInterchange::write_physical_netlist(const Context * ctx, const std::str
 
     for(auto & cell_name : placed_cells) {
         const CellInfo & cell = *ctx->cells.at(cell_name);
+
+        if(cell.bel == BelId()) {
+            continue;
+        }
+
+        if(!ctx->isBelLocationValid(cell.bel)) {
```
gatecat, line 418: this seems more like some kind of error would be appropriate than just skipping over the cell?

litghost: Fair enough.

litghost: Fixed.

PR #639, opened 2021-03-22
Update parameter processing based on new DeviceResources metadata

litghost:
This requires a new release of the `python-fpga-interchange` lib, which hasn't been cut yet.

PR #640, opened 2021-03-23
Initial inverter logic for FPGA interchange

litghost:
Initial version of inverter logic.

For now just implements some inspection capabilities, and the site router (for now) avoids inverted paths.

PR extends #639.  This requires a new release of the `python-fpga-interchange` lib, which hasn't been cut yet.

### DIFF
--- a/fpga_interchange/chipdb.h
+++ b/fpga_interchange/chipdb.h
@@ -34,7 +34,7 @@ NEXTPNR_NAMESPACE_BEGIN
  * kExpectedChipInfoVersion
  */
 
-static constexpr int32_t kExpectedChipInfoVersion = 3;
+static constexpr int32_t kExpectedChipInfoVersion = 4;
 
 // Flattened site indexing.
 //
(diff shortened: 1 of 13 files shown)


fpga_interchange/chipdb.h
```diff
@@ -306,6 +328,14 @@ inline const SiteInstInfoPOD &site_inst_info(const ChipInfoPOD *chip_info, int32
     return chip_info->sites[chip_info->tiles[tile].sites[site]];
 }
 
+enum SyntheticType {
+    NOT_SYNTH = 0,
+    SYNTH_SIGNAL = 1,
```
gatecat, line 333: is there a particular use case intended for non-gnd/vcc synthetic signals, out of curiosity?

litghost: One of the techniques we’ve been working on is static fabric subregions, which is related to out of context builds.  We might use synth signals to represents those

litghost:
CI is green!

PR #641, opened 2021-03-23
Initial lookahead for FPGA interchange.

litghost:
This PR is the last PR split from #627.

Currently the lookahead is disabled by default because of the time to
compute and RAM usage.  However it does appear to work reasonably well
in testing.  Further effort is required to lower RAM usage after initial
computation, and explore trade-off for cheaper time to compute.


fpga_interchange/arch.cc
```diff
@@ -83,13 +90,30 @@ void IdString::initialize_arch(const BaseCtx *ctx) {}
 
 static const ChipInfoPOD *get_chip_info(const RelPtr<ChipInfoPOD> *ptr) { return ptr->get(); }
 
+static std::string sha1_hash(const char *data, size_t size)
+{
+    boost::uuids::detail::sha1 hasher;
+    hasher.process_bytes(data, size);
+
+    unsigned int digest[5];
```
gatecat, line 98: nitpick, can you use `uint32_t` here as is convention in nextpnr when an explicit width is desired

fpga_interchange/lookahead.cc
```diff
@@ -0,0 +1,1522 @@
...
+
+    if (ctx->args.rebuild_lookahead || !read_lookahead(chipdb_hash, lookahead_filename)) {
+        build_lookahead(ctx, rng);
+        if (!ctx->args.dont_write_lookahead) {
+            write_lookahead(chipdb_hash, lookahead_filename);
+        }
+    }
+}
+
+bool safe_add_i32(int32_t a, int32_t b, int32_t *out) { return !__builtin_add_overflow(a, b, out); }
```
gatecat, line 1159: this will need a MSVC/other compiler fallback

gatecat:
In terms of tradeoff, it'd definitely be interesting to compare this approach with the approach that nextpnr-xilinx uses - essentially a simple fixed model for routing delay, plus site pin to routing location expansion done dynamically for used pins only inbetween place and route.

fpga_interchange/main.cc
```diff
@@ -55,6 +55,8 @@ po::options_description FpgaInterchangeCommandHandler::getArchOptions()
     specific.add_options()("netlist", po::value<std::string>(), "FPGA interchange logical netlist to read");
     specific.add_options()("phys", po::value<std::string>(), "FPGA interchange Physical netlist to write");
     specific.add_options()("package", po::value<std::string>(), "Package to use");
+    specific.add_options()("rebuild_lookahead", "Ignore lookahead cache and rebuild");
```
gatecat, line 58: nextpnr convention is to use `-`, not `_`, in multi-word arguments

fpga_interchange/lookahead.cc
```diff
@@ -0,0 +1,1522 @@
...
+
+            RoutingNode next_node;
+            next_node.wire_to_expand = new_wire;
+            next_node.cost = node.cost + ctx->getPipDelay(pip).maxDelay() + ctx->getWireDelay(new_wire).maxDelay();
+
+            if (ctx->is_site_port(pip)) {
+                // Done with expansion, record the path if cheaper.
+                // Only the first path to each wire will be the cheapest.
+                TypeWireId route_to;
+                route_to.type = ctx->chip_info->tiles[pip.tile].type;
```
gatecat, line 104: is there a reason to be reaching deep into `chip_info` here rather than using the `TypeWireId` constructor like elsewhere?

fpga_interchange/arch.cc
litghost: So it looks like `unsigned int` is the type that is used by `boost::uuids::detail::sha1`, see https://www.boost.org/doc/libs/1_70_0/boost/uuid/detail/sha1.hpp .  Do you still want me to change it to `uint32_t` in that case?

litghost: I'm changing this to `boost::uuids::detail::sha1::digest_type`.

gatecat: `boost::uuids::detail::sha1::digest_type` seems reasonable

fpga_interchange/lookahead.cc
litghost: Yes, because if `src_index` is a node, we don't want `TypeWireId` to be the node, we want it to be the local tile wire connected to the source of the pip.

litghost: Same story with line 201.

fpga_interchange/main.cc
litghost: Fixed

fpga_interchange/lookahead.cc
litghost: Fixed

fpga_interchange/lookahead.cc
litghost: Done.  I took a look at boost's safe int option, and it is basically undocumented, so not a good option here.

litghost:
> In terms of tradeoff, it'd definitely be interesting to compare this approach with the approach that nextpnr-xilinx uses - essentially a simple fixed model for routing delay, plus site pin to routing location expansion done dynamically for used pins only inbetween place and route.

I suspect that `nextpnr-xilinx` approach is superior in terms of runtime performance (better memory locality and footprint and less hash table lookups), and likely is of adequate quality for A* purposes.  The problem is that it isn't generalized.  This lookahead is purely data driven, so in theory any architecture works, both timing and non-timing.  Further refinement can and should be done to:
 - Reduce the redundancy in the `CostMap` data, potentially reducing the entire matrix down to a parametric equation
 - Reduce the amount of exploration (dynamically?) during the lookahead computation

This lookahead is likely not going to stand the test of time upon further iteration, but I needed something in place to lower router runtimes for murax and designs of increasing complexity.

fpga_interchange/arch.cc
litghost: Fixed

gatecat:
> Reduce the redundancy in the CostMap data, potentially reducing the entire matrix down to a parametric equation

This is definitely the most interesting approach to me, as this is basically something that can be done once (say as an overnight job on a big machine, even) and then become part of the furniture, so to speak. Perhaps this can even fit into the machine learning buzzword :)

gatecat (approved):

litghost:
CI is green!

PR #645, opened 2021-03-23
FPGA interchange: Add counter and ram tests

litghost:
- Small fix to `get_net_type` that was sometimes poisoning the cache
 - Change to placer_heap legalization order and loop.


litghost:
Issues for these designs to work:
 - [ ] Implement "best constant" routing, because of https://github.com/Xilinx/RapidWright/issues/142
 - [ ] Some unused RAMB BEL ports need to be tied to GND/VCC.  Investigate why they are not correctly getting tied?
 - [ ] Investigate placer1 sometimes generating invalid placements (see https://github.com/YosysHQ/nextpnr/pull/645/checks?check_run_id=2178413535)
 - [ ] Need to block invalid LUT route-thru's, otherwise could result in impossible LUT packing.

After these designs work, I believe a murax design should also work.

litghost:
https://github.com/YosysHQ/nextpnr/pull/645/commits/83f4759d9accca301692b3f1f1d3acba593ef66d isn't a pretty solution, but it does appear to work.  @gatecat Idea's for making it better?

litghost:
> [83f4759](https://github.com/YosysHQ/nextpnr/commit/83f4759d9accca301692b3f1f1d3acba593ef66d) isn't a pretty solution, but it does appear to work. @gatecat Idea's for making it better?

To expound, at this point the site router can preferentially select for routes that have the right inversion.  However, because the net splits at the inverter, there is an open question of which net is on the other side of the inverter.  For example, if the cell pin in question is bound to GND, then the net on the output of the inverter is the `GLOBAL_LOGIC0` net.  However the net model doesn't allow for there to be multiple drivers of the same net.  In this case, we've "created" a new `GLOBAL_LOGIC0` from no where.  My answer for now is to simply "mask" the BEL pins away from the router if they have been dealt with in the site router, and then "unmask" afterwards.

gatecat:
What about creating a new net for the connection between the inverter and the relevant cell/bel pin? I appreciate this will probably make netlist generation and ripping up the inverters a bit more complicated.

gatecat:
> This solution isn't ideal because eventually we might want to "create" GLOBAL_LOGIC0 net's from the outputs of LUT's. From the router perspective, those will look like "new" sources, unless we add a way to "fake" connections for the global constant works. A bit of a mess really.

nextpnr-xilinx did the latter and had pips from the constant network to LUT outputs (I think I just did this for one or two LUTs per SLICE to avoid too many pips)

litghost:
> What about creating a new net for the connection between the inverter and the relevant cell/bel pin? I appreciate this will probably make netlist generation and ripping up the inverters a bit more complicated.

I thought about this.  This couples poorly with the current 1 to many cell port / BEL pin model.  When there is a 1 to many cell port / BEL pin and there are multiple inverters (which is the case for a RAMB36E1), then each inverter output would be it's own net.  The current net model is that each cell port has 1 net.  Bomb, model breaks.  `*sad face*`

litghost:
> > This solution isn't ideal because eventually we might want to "create" GLOBAL_LOGIC0 net's from the outputs of LUT's. From the router perspective, those will look like "new" sources, unless we add a way to "fake" connections for the global constant works. A bit of a mess really.
> 
> nextpnr-xilinx did the latter and had pips from the constant network to LUT outputs (I think I just did this for one or two LUTs per SLICE to avoid too many pips)

And this probably still makes sense for the LUT outputs, where we add a site ports (and maybe a site pip) to each LUT output site wire for VCC/GND.  There would need to be some logic to handle checking LUT equations if the site port / site pip is legal.

gatecat:
A final alternate suggestion is to route the pre-inverter constant all the way to the cell/bel pins (so the pins are connected to the "wrong" value from a logical point of view) and have some other special property to keep track of the inversion. This might break in cases where the presence of the inverter actually makes requires different routing beyond the inversion site pip but I don't know if those matter.

fpga_interchange/site_router.cc
```diff
@@ -404,8 +405,92 @@ void remove_solution(SiteArch *ctx, std::vector<SitePip>::const_iterator pips_be
...
+
+        NPNR_ASSERT(lhs.net == rhs.net);
+
+        PhysicalNetlist::PhysNetlist::NetType net_type = ctx->ctx->get_net_type(lhs.net->net);
+        if (net_type == PhysicalNetlist::PhysNetlist::NetType::SIGNAL) {
+            return non_inverting_preference(lhs, rhs);
+        }
+
+        // All GND/VCC nets use out of site sources.  Local constant sources
+        // are still connected via syntetic edges to the global GND/VCC
```
gatecat, line 457: nitpick - spelling

common/placer_heap.cc
```diff
@@ -841,7 +841,31 @@ class HeAPPlacer
         int ripup_radius = 2;
         int total_iters = 0;
         int total_iters_noreset = 0;
-        while (!remaining.empty()) {
+
+        bool need_legalize = require_validity;
+
+        /* First spread without legality checking, then check at the end! */
```
gatecat, line 847: I would ideally like a bit more investigation as to why this is needed, as I can't see an obvious reason.  It _could_ be the issue we've discussed before about how ripping up a cell can turn a legal placement illegal but I'm not sure if legaliser should ever be hitting that case.

litghost:
> A final alternate suggestion is to route the pre-inverter constant all the way to the cell/bel pins (so the pins are connected to the "wrong" value from a logical point of view) and have some other special property to keep track of the inversion. This might break in cases where the presence of the inverter actually makes requires different routing beyond the inversion site pip but I don't know if those matter.

Good thought, but at least in the 7-series fabric, the GND route consumes the `GFAN0`/`GFAN1` resource, versus the VCC route which does not.  I'll think about this some more after I get the counter and RAM designs working.

common/placer_heap.cc
litghost: I agree. I think I'll finish getting the counter/RAM designs generating a useful DCP and then circle back and investigate this issue.

gatecat:
> Good thought, but at least in the 7-series fabric, the GND route consumes the GFAN0/GFAN1 resource, versus the VCC route which does not. I'll think about this some more after I get the counter and RAM designs working.

The idea would be the pin is connected directly to the constant that should be routed to it, which means the netlist is logically wrong (because the inversion is now metadata) but physically correct

litghost:
> > Good thought, but at least in the 7-series fabric, the GND route consumes the GFAN0/GFAN1 resource, versus the VCC route which does not. I'll think about this some more after I get the counter and RAM designs working.
> 
> The idea would be the pin is connected directly to the constant that should be routed to it, which means the netlist is logically wrong (because the inversion is now metadata) but physically correct

Ah, sure.

litghost:
> * Some unused RAMB BEL ports need to be tied to GND/VCC.  Investigate why they are not correctly getting tied?

Issue filed with RapidWright for this one: https://github.com/Xilinx/RapidWright/issues/143 .  I've examined the `DeviceResources` file from RapidWright, and the problem appears to be in RapidWright.  Unclear if the Java API is broken, or if the bug exists in `DeviceResourcesWriter`.  I'll continue investigating to see if I can fix it in the open portion of RapidWright.

gatecat:
Not an issue for now, but something in the LUT code that I note is that it assumes Vcc is always the best constant for unused LUT inputs. This is true for Xilinx; Nexus and ECP5 but is not true for iCE40 and MachXO2 where unused LUT inputs are pulled low and connecting to Vcc would require explicit use of routing resources.

litghost:
> Not an issue for now, but something in the LUT code that I note is that it assumes Vcc is always the best constant for unused LUT inputs. This is true for Xilinx; Nexus and ECP5 but is not true for iCE40 and MachXO2 where unused LUT inputs are pulled low and connecting to Vcc would require explicit use of routing resources.

Right.  I'll put a FIXME.   The new method `prefered_constant_net_type` was specifically intented to answer the question "for this site pin, what is the easiest constant to route".  I believe it should be straight forward to fixup the logic here.

litghost:
> Not an issue for now, but something in the LUT code that I note is that it assumes Vcc is always the best constant for unused LUT inputs. This is true for Xilinx; Nexus and ECP5 but is not true for iCE40 and MachXO2 where unused LUT inputs are pulled low and connecting to Vcc would require explicit use of routing resources.

As a sanity check, looks like the Nexus arch has a LUT5/LUT4x2 fracture, with the high address pin being "M0/SEL".  It also appears that tying "M0/SEL" to GND causes puts it in LUT4x2 (versus something akin to a LUT5_2).  Assuming this understanding is right, I think implementing `prefered_constant_net_type` and using it in the LUT logic should be enough to clear this up!

gatecat:
> As a sanity check, looks like the Nexus arch has a LUT5/LUT4x2 fracture, with the high address pin being "M0/SEL". It also appears that tying "M0/SEL" to GND puts the hardware into a LUT4x2 config (versus something akin to a LUT5_2 config if "M0/SEL" is a signal net). Assuming this understanding is right, I think implementing prefered_constant_net_type and using it in the LUT logic should be enough to clear this up!

No, the M input is essentially the same as the Xilinx X input - used as the selection for the MUX2 when combining two LUT4s into a LUT5, but can also be used if driving the FF D input from fabric rather than from the LUT. Whether or not the MUX2 is used depends not on the M input but whether or not you use the 'F' outputs from the LUTs or the OFX output from the MUX.

This diagram might clear things up a  bit?

Screenshot from 2021-03-25 15-47-31

litghost:
> No, the M input is essentially the same as the Xilinx X input - used as the selection for the MUX2 when combining two LUT4s into a LUT5, but can also be used if driving the FF D input from fabric rather than from the LUT. Whether or not the MUX2 is used depends not on the M input but whether or not you use the 'F' outputs from the LUTs or the OFX output from the MUX.
> 
> This diagram might clear things up a bit?

Right. The MUX2 is more akin to the F7MUX than it is to the F6MUX found in Xilinx.  Ok.

gatecat:
> Right. The MUX2 is more akin to the F7MUX than it is to the F6MUX found in Xilinx. Ok.

Correct

common/placer_heap.cc
litghost: I think I have a replicating test case where the current site router fails to find a solution, but a solution is possible.  Looks like a bug in the backtrace algorithm.  Debugging now, hopefully have a fix soon.

litghost: Bug had to do with unrolling bad solutions, fix incoming.

fpga_interchange/site_router.cc
litghost: Yep, fixed!

common/placer_heap.cc
litghost: Fix is in https://github.com/YosysHQ/nextpnr/pull/645/commits/2a56539162e5d25e064a5940db75e8eb571ae917

litghost:
I think this PR is ready.  I believe the counter design should reliably P&R now.  The RAM design P&R's, but because of https://github.com/Xilinx/RapidWright/issues/143 , some of the constants aren't routed.  I believe both the RAM and murax designs will work after https://github.com/Xilinx/RapidWright/issues/143 is fixed (and https://github.com/Xilinx/RapidWright/pull/150 is merged).

litghost:
@gatecat This PR hasn't had a comment in a bit since I made [this comment requesting a final review](https://github.com/YosysHQ/nextpnr/pull/645#issuecomment-807806883).  Is there something I forgot to update in this PR?

gatecat:
Sorry, seems like I missed that comment at the time.

litghost:
> Sorry, seems like I missed that comment at the time.

No problem.  That's why I asked.

PR #649, opened 2021-03-26
interchange: add archcheck tests to all-device-test target

acomodi:


gatecat (approved):

acomodi:
CI went green. I have added also an installed cache of yosys, so that, if yosys version remains unchanged, it does not get rebuilt at all.

gatecat:
LGTM, thanks!

Issue #652, opened 2021-03-29
Placer BEL assumption violation


litghost:
There have been some intermittent placer failures with the new FPGA interchange 7-series implementation (see https://github.com/YosysHQ/nextpnr/pull/619#issuecomment-800658910 and https://github.com/YosysHQ/nextpnr/pull/651#issuecomment-809673534).  I've traced the failure down to the following assumption that both placer1 and placer_heap make which is:

 - If the placer starts with a valid design, and always makes valid swaps (e.g. new location is legal and displaced cell moves to old location and that is also legal), then the final design is valid

There appears to be a counter example in the 7-series architecture that means that additional logic is required.  Here is the counter example:

First the circuit being packed:
```
wire a1, a2, a3, a4;
wire o5;
LUT2 lut_a5lut(.I0(a1), .I1(a2), .O(o5));
wire o6;
LUT2 lut_a6lut(.I0(a3), .I1(a4), .O(o6));

wire clk;
wire a5q;
FDRE a5ff(.C(clk), .R(1'b0), .D(o6), .Q(a5q));
wire aq;
FDRE aff(.C(clk), .R(1'b0), .D(o5), .Q(aq));

# Provide a sink outside of the site
FDRE off(.D(a5q));
FDRE o2ff(.D(aq));

```

This circuit can be legal packed as:
```
lut_a5lut => A5LUT
lut_a6lut => A6LUT
aff => AFF
a5ff => A5FF
```

This is legal because the `aff` uses the `AQ` output port.  However if the `aff` cell is unplaced, the entire site becomes unroutable due to congestion at the `AMUX`.  For clarity, the current placer placer1 and placer_heap algorithm will swap `aff` to an empty location FF if able.

For reference here is the TCL to create this simple circuit in Vivado and demonstrate the issue:
```
link_design -part xc7a35tcsg324-1

set lut_a5lut [create_cell -ref LUT2 lut_a5lut]
set lut_a6lut [create_cell -ref LUT2 lut_a6lut]
set a5ff [create_cell -ref FDRE a5ff]
set aff [create_cell -ref FDRE aff]
set off [create_cell -ref FDRE off]
set o2ff [create_cell -ref FDRE o2ff]

set clk [create_net clk]
set o5 [create_net o5]
set o6 [create_net o6]
set a1 [create_net a1]
set a2 [create_net a2]
set a3 [create_net a3]
set a4 [create_net a4]
set aq [create_net aq]
set a5q [create_net a5q]

[...]

PR #653, opened 2021-03-29
[interchange] Fix site pip check for drivers.

litghost:
Previous code allowed router to enter sites with no sinks.


gatecat:
LGTM, assuming the CI failures are due to #652 and this is OK to be merged?

litghost:
> LGTM, assuming the CI failures are due to #652 and this is OK to be merged?

I've rebased after the fix for #652, and it is green now

PR #655, opened 2021-03-30
interchange: Fix illegal placements

gatecat:
Alternative fix to #652


common/placer1.cc
```diff
@@ -552,7 +552,7 @@ class SAPlacer
             add_move_cell(moveChange, other_cell, newBel);
         }
 
-        if (!ctx->isBelLocationValid(newBel) || ((other_cell != nullptr && !ctx->isBelLocationValid(oldBel)))) {
+        if (!ctx->isBelLocationValid(newBel) || !ctx->isBelLocationValid(oldBel)) {
```
litghost, line 555: I think it is worth adding a comment why this changed.  I could see it changing back otherwise.

fpga_interchange/arch.h
```diff
@@ -825,10 +823,9 @@ struct Arch : ArchAPI<ArchRanges>
             if (!is_cell_valid_constraints(cell, tile_status, explain_constraints)) {
                 return false;
             }
-
-            auto &bel_data = bel_info(chip_info, bel);
-            return get_site_status(tile_status, bel_data).checkSiteRouting(getCtx(), tile_status);
         }
+        auto &bel_data = bel_info(chip_info, bel);
+        return get_site_status(tile_status, bel_data).checkSiteRouting(getCtx(), tile_status);
```
litghost, line 828: Need a comment here about why we are checking site status for an unbound BEL.  I need to check that the SiteRouter does the right thing on an empty site.

litghost:
I think a similar change is required for the legalization loop in placer_heap.  Let me find it.

litghost:
> I think a similar change is required for the legalization loop in placer_heap. Let me find it.

I think the placer_heap legalization didn't have this issue. LGTM, merge once green!

litghost (approved):
LGTM

PR #659, opened 2021-04-01
[interchange] Pseudo pip fixes

litghost:
Builds on #658.  @gatecat May break this PR further if you'd like.

### DIFF
--- a/fpga_interchange/pseudo_pip_model.cc
+++ b/fpga_interchange/pseudo_pip_model.cc
@@ -1,0 +1,470 @@
+/*
+ *  nextpnr -- Next Generation Place and Route
+ *
+ *  Copyright (C) 2021  Symbiflow Authors
+ *
+ *
+ *  Permission to use, copy, modify, and/or distribute this software for any
+ *  purpose with or without fee is hereby granted, provided that the above
+ *  copyright notice and this permission notice appear in all copies.
+ *
+ *  THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ *  WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ *  MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ *  ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ *  WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ *  ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ *  OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ *
+ */
+
+#include "pseudo_pip_model.h"
+
+#include "context.h"
+
+//#define DEBUG_PSEUDO_PIP
+
+NEXTPNR_NAMESPACE_BEGIN
+
+void PseudoPipData::init_tile_type(const Context *ctx, int32_t tile_type) {
+    if(max_pseudo_pip_for_tile_type.count(tile_type)) {
+        return;
+    }
+
+    const TileTypeInfoPOD & type_data = ctx->chip_info->tile_types[tile_type];
+    int32_t max_pseudo_pip_index = -1;
+    for(int32_t pip_idx = 0; pip_idx < type_data.pip_data.ssize(); ++pip_idx) {
+        const PipInfoPOD & pip_data = type_data.pip_data[pip_idx];
+        if(pip_data.pseudo_cell_wires.size() == 0) {
+            continue;
+        }
+
+        if(pip_idx > max_pseudo_pip_index) {
+            max_pseudo_pip_index = pip_idx;
+        }
+
+        HashTables::HashSet<size_t> sites;
+        std::vector<PseudoPipBel> pseudo_pip_bels;
+        for(int32_t wire_index : pip_data.pseudo_cell_wires) {
+            const TileWireInfoPOD &wire_data = type_data.wire_data[wire_index];
+            if(wire_data.site == -1) {
+                continue;
+            }
+
+            // Only use primary site types for psuedo pips
+            //
+            // Note: This assumption may be too restrictive.  If so, then 
+            // need to update database generators to provide 
+            // pseudo_cell_wires for each site type, not just the primary.
+            if(wire_data.site_variant != -1) {
+                continue;
+            }
+
+            sites.emplace(wire_data.site);
+
+            int32_t driver_bel = -1;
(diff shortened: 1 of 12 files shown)


litghost:
This is the set of fixes needed to reliably P&R murax.  There is still an issue with clock routing, looking it now.

litghost:
@gatecat Does the nexus interchange emitter not supply a single input cell type in https://github.com/SymbiFlow/fpga-interchange-schema/blob/4b3269148653b4e73c1c18c86d3c31d6d29e8453/interchange/DeviceResources.capnp#L594 ?

litghost:
Per https://github.com/YosysHQ/yosys/blob/0ccc7229c02449618209a20d66b9fa07e3ea79f2/techlibs/nexus/cells_map.v#L65 it looks like there is only `INV` and `LUT4` cells.  Having other LUT widths would simplify unused pin analysis when doing smarter LUT rotation, but it isn't strictly speaking needed.  Thoughts?

gatecat:
> @gatecat Does the nexus interchange emitter not supply a single input cell type in https://github.com/SymbiFlow/fpga-interchange-schema/blob/4b3269148653b4e73c1c18c86d3c31d6d29e8453/interchange/DeviceResources.capnp#L594 ?

No, this isn't being done yet as it wasn't previously needed but I can fix that easily enough

gatecat:
> Per https://github.com/YosysHQ/yosys/blob/0ccc7229c02449618209a20d66b9fa07e3ea79f2/techlibs/nexus/cells_map.v#L65 it looks like there is only `INV` and `LUT4` cells. Having other LUT widths would simplify unused pin analysis when doing smarter LUT rotation, but it isn't strictly speaking needed. Thoughts?

My understanding is that the vendor tools also only use LUT4, and this was done to maximise netlist compatibility with them (similar to iCE40 only using SB_LUT4).

litghost:
> > Per https://github.com/YosysHQ/yosys/blob/0ccc7229c02449618209a20d66b9fa07e3ea79f2/techlibs/nexus/cells_map.v#L65 it looks like there is only `INV` and `LUT4` cells. Having other LUT widths would simplify unused pin analysis when doing smarter LUT rotation, but it isn't strictly speaking needed. Thoughts?
> 
> My understanding is that the vendor tools also only use LUT4, and this was done to maximise netlist compatibility with them (similar to iCE40 only using SB_LUT4).

Ok.  I think I can make the use of the wire lut cell type optional unless a "complicated" lut sharing schema is used.  Given that nexus and ice40 doesn't use particularly complicated LUT sharing, the wire lut cell type likely isn't important.  7-series, US+ and Versal have tighter LUT coupling (especially for SLICEM CLB's), so having the wire lut is more important.  Versal in particular is a weird case, in that it has dedicated upper and lower O5 outputs and the 7-series/US+ O6 type outputs.

fpga_interchange/pseudo_pip_model.cc
```diff
@@ -0,0 +1,475 @@
...
+        cell.lut_cell.pins.push_back(IdString(ctx->wire_lut->input_pins[0]));
+        cell.lut_cell.equation.resize(2);
+        cell.lut_cell.equation.set(0, false);
+        cell.lut_cell.equation.set(1, true);
+
+        // Map LUT input to input wire used by pseudo pip.
+        IdString input_bel_pin(bel_data.ports[bel.input_bel_pin]);
+        cell.cell_bel_pins[IdString(ctx->wire_lut->input_pins[0])].push_back(input_bel_pin);
+
+        lut_mappers[bel_data.lut_element].cells.push_back(&cell);
```
gatecat, line 382: this seems fragile at a glance, couldn't subsequent `emplace_back`s invalidate `&cell`?

gatecat: never mind, I missed the call to `reserve`

fpga_interchange/pseudo_pip_model.cc
```diff
@@ -0,0 +1,475 @@
...
+    //  - Trivially marking other pseudo pips as unavailable if it requires
+    //    logic BELs used by active pseudo pips (or bound by cells).
+    //  - Determine if remaining pseudo pips can be legally placed.  This
+    //    generally consists of:
+    //     - Checking LUT element
+    //     - FIXME: Checking constraints (when metadata is available)
+
+    const std::vector<int32_t> pseudo_pips_for_site = site_to_pseudo_pips.at(site);
+
+    // unused_pseudo_pips is a scratch array kept in the model, but is intended to be used during
```
gatecat, line 301: I think a word is missing from the end of this comment

litghost: Yep.  Fixed.

litghost:
For reference, this PR routes a valid murax, except for the BUFG output net (which I mentioned above).  If I let Vivado re-route the BUFG output net, then the design closes with a frequency ~25 MHz.  Not great, but not bad for having no timing information at all.  Now that this PR is green, I'll see if I can figure out why the BUFG output net ends up so mangled.  I strongly suspect the bounding box is the issue, but I'll need to debug it to be more confident.

@gatecat What I'm seeing with the BUFG output net is that the route does just fine with one of the CMTs, but then tries to route south across the CMT boundary, which messes the hold timing completely.

gatecat:
I suspect the global routing issues are probably around the notional locations of global nodes. While a hypothetically ideal lookahead could probably solve this, router1/2 are not optimised for global routing and are almost guaranteed to go wrong in more complex cases.

In practice I have found the best option for global routing is a dedicated backwards BFS pass that forces use only of global routes after leaving the destination tile (or nearabouts).  For UltraScale+ the rules get even more complex, as you first have to route to a single origin point using "ROUTE" type globals and then fan out using "DISTR" type globals in order to get skew low enough (iirc), or you'll have hold time failures (which incidentally nextpnr cannot currently identify or fix).

fpga_interchange/pseudo_pip_model.h
```diff
@@ -0,0 +1,108 @@
...
+    const std::vector<size_t> &get_possible_sites_for_pip(const Context *ctx, PipId pip) const;
+
+    // Get list of BELs the pseudo pip uses, and how it routes through them.
+    //
+    // This does **not** include site ports or site pips.
+    const std::vector<PseudoPipBel> &get_logic_bels_for_pip(const Context *ctx, int32_t site, PipId pip) const;
+
+    HashTables::HashMap<int32_t, size_t> max_pseudo_pip_for_tile_type;
+    HashTables::HashMap<std::pair<int32_t, int32_t>, std::vector<size_t>> possibles_sites_for_pip;
+    HashTables::HashMap<std::tuple<int32_t, int32_t, int32_t>, std::vector<PseudoPipBel>> logic_bels_for_pip;
```
gatecat, line 67: by three elements I think a custom key type is nicer than a tuple

litghost: Done.

litghost:
@gatecat I believe this should be ready to merge now.  I'll debug the BUFG routing and submit a fix in a follow on PR.

gatecat (approved):

litghost:
CI is green.

PR #663, opened 2021-04-06
Fix bug in router2 where router may give up too early.

litghost:
Was introduced in #612.  The logic before was intended to prevent the
router from terminating early when not using a bounding box, but the fix
in #612 simply removed that, meaning that the router might terminate
early incorrectly.  The solution here is to only use the toexplore
hysteric once a solution is found.

### DIFF
--- a/common/router2.cc
+++ b/common/router2.cc
@@ -654,7 +654,18 @@ struct Router2
         bool debug_arc = /*usr.cell->type.str(ctx).find("RAMB") != std::string::npos && (usr.port ==
                             ctx->id("ADDRATIEHIGH0") || usr.port == ctx->id("ADDRARDADDRL0"))*/
                 false;
-        while (!t.queue.empty() && iter < toexplore) {
+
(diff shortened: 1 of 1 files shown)


common/router2.cc
```diff
@@ -654,7 +654,18 @@ struct Router2
...
+
+        // When running without a bounding box, the toexplore limit should be
+        // suspended until a solution is reached.  Once a solution is found,
+        // the toexplore limit should be used again to prevent requiring the
+        // router to drain the routing queue.
+        //
+        // Note that is it important that the must_drain_queue be set to true
+        // when running without a bb to ensure that a routing failure is
+        // because there is not route, rather than just because the toexplore
+        // hysteric is incorrect.
```
gatecat, line 666: nitpick - heuristic

litghost: Yep, fixed.

gatecat (approved):

PR #664, opened 2021-04-06
interchange/nexus: Add counter example

gatecat:
Current issues:
 - needs gatecat/prjoxide#5 and SymbiFlow/python-fpga-interchange#53
 - a remap rule is currently being used to deal with the missing LUT input ports, that otherwise cause the interchange to fail. we should probably be auto-creating missing ports with their default value at some point (the Yosys JSON to logical netlist script?)
 - inversion is not implemented
 - ~~routing currently fails because it can't route a 0 (I'm not actually sure where this 0 comes from as Yosys isn't creating any VLO cells), this should be fixed using a combination of implementing inversion and 0-driver PIPs for LUT outputs~~ fixed for now with a slightly more intricate remap rule


litghost:
>     * a remap rule is currently being used to deal with the missing LUT input ports, that otherwise cause the interchange to fail. we should probably be auto-creating missing ports with their default value at some point (the Yosys JSON to logical netlist script?)

So I think this is a good discussion to have.  I would argue that if the relevant cell port is present in the Cell -> BEL pin map, then **something** needs to be connected to that cell port coming out of Yosys (e.g. before the Yosys JSON to logical netlist step).  `z`/`x` is not a legally routable signal, turns out!  In arch-defs, we typically use `setundef  -zero`.   In this case, maybe something like:
```
setundef -one t:LUT4 i:*
```
would be good?

gatecat:
I think the solution here is more complex than `setundef` as that doesn't work if ports are missing from the instantiation altogether. Although we could add the ports to the techmap rules, this still doesn't help for user instantiated LUTs missing ports (which the vendor tools allow). 

A Yosys pass that autocreates empty ports based on the blackbox, so that setundef can do its thing, might be the solution here.

litghost:
> I think the solution here is more complex than `setundef` as that doesn't work if ports are missing from the instantiation altogether. Although we could add the ports to the techmap rules, this still doesn't help for user instantiated LUTs missing ports (which the vendor tools allow).

All true.

> A Yosys pass that autocreates empty ports based on the blackbox, so that setundef can do its thing, might be the solution here.

This seems like the right answer.  An auxiliary question is how do simulation tools handles these cases?  I imagine verilator and iverilog doesn't like these partially connected cells either.  But maybe they have some special handling for this type of stuff.

gatecat:
The unconnected ports will become either x or z in simulation, iirc, I believe some or the sim models do have special case handling for cases where this matters (the LUT models are written to ignore undefined inputs that don't affect the output in any case)

litghost:
> The unconnected ports will become either x or z in simulation, iirc, I believe some or the sim models do have special case handling for cases where this matters (the LUT models are written to ignore undefined inputs that don't affect the output in any case)

So it is worth noting that I don't believe the yosys nexus cells_sim LUT4 is written to handle this properly.  But I'm not an expert in reading Verilog.

gatecat:
`?:` is specified to have a defined output when both branches are the same even if the select input is undefined, so the model as written should be OK provided the LUT function doesn't depend on disconnected inputs.

litghost:
> `?:` is specified to have a defined output when both branches are the same even if the select input is undefined, so the model as written should be OK provided the LUT function doesn't depend on disconnected inputs.

Ok, thanks.

gatecat:
Currently this is blocked on default cell pin issues, I had initially done YosysHQ/yosys#2722 as a solution but now I realise this isn't enough for all cases and I have created SymbiFlow/fpga-interchange-schema#35 for further discussion.

PR #665, opened 2021-04-07
Add CMake option to enable IPO (enabled by default).

cr1901:
For developers, the compile times of LTO may not be worth the performance benefit. Add an option to manually disable LTO for developers.


gatecat (approved):

mmicko:
Agree on this one, also some cross compile environment are not able to handle LTO correctly

PR #668, opened 2021-04-09
interchange: Disambiguate cell and bel pins when creating Vcc ties

gatecat:
The pins created for tieing to Vcc were being named after the bel pin, relying on the fact that Xilinx names cell and bel pins differently for LUTs. This isn't true for Nexus devices which uses the same names for both, and was causing a failure as a result.

This uses a "PHYS_" prefix that's highly unlikely to appear in a cell pin name to disambiguate.

cc @litghost


litghost:
Lgtm

PR #669, opened 2021-04-09
interchange: Pin prjoxide commit in CI

gatecat:
This way things can be merged into master that necessitate other changes without breaking the CI.

Ideally this would use tags long term but prjoxide isn't quite stable enough yet, this should suffice for now.

cc @acomodi


acomodi (approved):
LGTM. Might be worth to add a `FIXME` or `TODO` comment as a reminder to transition to use tags revisions similarly to what python-interchange has.

Issue #672, opened 2021-04-12
FastBels returns pointers to vector elements which later become invalid


adamgreig:
I hit this adding some cellGroups to ECP5 for the DSP; in HeAP's `CutSpreader` ctor [it calls](https://github.com/YosysHQ/nextpnr/blob/master/common/placer_heap.cc#L1153) FastBels' `getBelsForBelBucket` and stores the returned pointer into `fast_bels_by_partition_type` in `fb`, and does this for each bucket/cell type. Unfortunately each call to `getBelsForBelBucket` [calls](https://github.com/YosysHQ/nextpnr/blob/master/common/fast_bels.h#L162) `addBelBucket` which [resizes](https://github.com/YosysHQ/nextpnr/blob/master/common/fast_bels.h#L104) `fast_bels_by_partition_type`, invalidating the previous pointers.

For now I have a temporary workaround in `CutSpreader` to call `getBelsForBelBucket` for each bucket first, to ensure the vector is fully constructed, and then re-do the loop actually storing the results. I'm not sure what the best fix is though; should `FastBels` be returning pointers which it will likely invalidate? If so I guess maybe `CutSpreader` should copy out the bels instead of storing the pointer?

```diff
diff --git a/common/placer_heap.cc b/common/placer_heap.cc
index 042f3046..329c21ac 100644
--- a/common/placer_heap.cc
+++ b/common/placer_heap.cc
@@ -1145,6 +1145,13 @@ class HeAPPlacer
       public:
         CutSpreader(HeAPPlacer *p, const std::unordered_set<BelBucketId> &buckets) : p(p), ctx(p->ctx), buckets(buckets)
         {
+            // Call getBelsForBelBucket for each bucket to ensure the
+            // fast_bels_by_partition_type vector doesn't get resized
+            // after we store pointers into it.
+            for (BelBucketId bucket : sorted(buckets)) {
+                FastBels::FastBelsData *fast_bels;
+                p->fast_bels.getBelsForBelBucket(bucket, &fast_bels);
+            }
             // Get fast BELs data for all buckets being Cut/Spread.
             size_t idx = 0;
             for (BelBucketId bucket : sorted(buckets)) {
```

PR #673, opened 2021-04-12
fast_bels: Don't return pointer that might become invalid

gatecat:
@adamgreig, does this look like a reasonable solution to #672  ?


adamgreig (approved):
LGTM and I've checked it fixes the problem I was having on ECP5, thanks!

PR #677, opened 2021-04-13
interchange: Allow pseudo-cells with no input pins

gatecat:
These are used for the LUT-as-GND-driver pseudo-pips in the Nexus arch, which will probably be required for UltraScale too.

Anyone want to give this a review? @acomodi or @tmichalak ?

### DIFF
--- a/fpga_interchange/pseudo_pip_model.cc
+++ b/fpga_interchange/pseudo_pip_model.cc
@@ -156,7 +156,6 @@ void PseudoPipData::init_tile_type(const Context *ctx, int32_t tile_type)
                 }
 
                 NPNR_ASSERT(output_bel_pin == bel.output_bel_pin);
-                NPNR_ASSERT(input_bel_pin != -1);
                 bel.input_bel_pin = input_bel_pin;
 
                 logic_bels_for_pip[LogicBelKey{tile_type, pip_idx, site}].push_back(bel);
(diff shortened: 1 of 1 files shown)


fpga_interchange/pseudo_pip_model.cc
```diff
@@ -378,13 +377,22 @@ void PseudoPipModel::update_site(const Context *ctx, size_t site)
...
 
-        // Map LUT input to input wire used by pseudo pip.
-        IdString input_bel_pin(bel_data.ports[bel.input_bel_pin]);
-        cell.cell_bel_pins[IdString(ctx->wire_lut->input_pins[0])].push_back(input_bel_pin);
+        if (bel.input_bel_pin == -1) {
+            // FIXME: currently assume that LUT route-throughs with no input pins are GND drivers as this is all we need
+            // for Nexus/Xilinx where Vcc is readily available and cheap This won't be true for other arches
+            cell.lut_cell.equation.resize(2);
+            cell.lut_cell.equation.set(0, false);
+            cell.lut_cell.equation.set(1, false);
```
acomodi, line 386: This allows having constant GND generated by the LUT, is this correct?

fpga_interchange/pseudo_pip_model.cc
```diff
@@ -440,7 +448,7 @@ void PseudoPipModel::update_site(const Context *ctx, size_t site)
         // that pseudo pip violates the LUT element equation.
         for (const PseudoPipBel &bel : bels) {
             const BelInfoPOD &bel_data = type_data.bel_data[bel.bel_index];
-            if (bel_data.lut_element == -1) {
+            if (bel_data.lut_element == -1 || bel.input_bel_pin == -1) {
```
acomodi, line 451: I am not sure about this, but shouldn't the no-input pseudo cell be checked as well?

acomodi (commented):
Comments in

fpga_interchange/pseudo_pip_model.cc
gatecat: as it stands yes, see the comment this is the main use case for LUT pseudo-cells with no input pins. supporting anything else will need more work on the schema side, I think

fpga_interchange/pseudo_pip_model.cc
gatecat: hmm, possibly, I think I need to understand how `lut_wires_unavailable` works better

gatecat: I've now added a check that in this case, any of the bel pins is not blocked by the LUT equation, which I think is at least more correct than no check at all.

fpga_interchange/pseudo_pip_model.cc
acomodi: All right. As I am dealing with the LUT-thru FASM generation, I wonder how to express and more importantly identify this case. 

For instance, at the moment I am implementing the logic to identify the LUT-thrus by examining whether a LUT bel is traversed entirely in the same physical net, meaning that the LUT was indeed used as a route-thru.

In this case, however, I assume (maybe wrongly) that the net will start from the pseudoCell output, in this case, the LUT output, to reach the various sinks, therefore making the above logic unusable, as the input pins of the LUT appears as unused.

gatecat: I think this special case could be identified as the net being a ground/vcc net on the LUT output, which should force the LUT function for that net to all-zeros or all-ones accordingly.

acomodi: Ah, right, cause the net will be a const net, therefore yes, this case can be actually easily identified. Thanks for the clarification.

acomodi (approved):
LGTM

PR #678, opened 2021-04-13
interchange: add FASM generation target and clean-up tests

acomodi:

This PR adds a target to generate the FASM output.


fpga_interchange/examples/tests.cmake
```diff
@@ -266,8 +275,37 @@ function(add_interchange_test)
...
+        set(fasm ${CMAKE_CURRENT_BINARY_DIR}/${name}.fasm)
+        add_custom_command(
+            OUTPUT ${fasm}
+            COMMAND
+                ${PYTHON_EXECUTABLE} -mfpga_interchange.fasm_generator
+                    --schema_dir ${INTERCHANGE_SCHEMA_PATH}
+                    --family ${device_family}
+                    ${device_loc}
+                    ${netlist}
+                    ${phys} > ${fasm}
```
gatecat, line 299: doing redirection like this tends to cause problems if the FASM generator hits an error, as an empty file will be created and so `make` won't automatically re-run the fasm generator when running the makefile again.

I think the best option here is to add an output file argument to the FASM generator.

acomodi: Right, I'll make the corresponding changes to the FASM generator and update this once the new tag is there

gatecat (approved):

PR #688, opened 2021-04-27
New cluster API

gatecat:
This PR describes a new way of representing clusters to give considerably more flexibility to the architecture as to how cells are placed relative to each other; compared to the existing `constr_{x, y, z}` approach.

In particular, this approach can enable architectures to represent:
 - carry chains with discontinuities in the relative coordinates, for example due to clocking rows
 - IO-IOLOGIC connectivity which may be better defined by the routing graph than by relative coordinates
 - intra-site dedicated routing where the `z` coordinates aren't meaningful, and using bel names or routing to define constraints makes more sense.

Currently this PR only contains the API specification and not any implementation. I'd really appreciate some feedback before I start working on implementation, including a default implementation similar to the existing API for current arches.

cc @acomodi @kowalewskijan @mkurc-ant for any thoughts?

### DIFF
--- a/common/arch_api.h
+++ b/common/arch_api.h
@@ -142,8 +142,8 @@ template <typename R> struct ArchAPI : BaseCtx
     // Cluster methods
     virtual CellInfo *getClusterRootCell(ClusterId cluster) const = 0;
     virtual ArcBounds getClusterBounds(ClusterId cluster) const = 0;
-    virtual Loc getClusterOffset(CellInfo *cell) const = 0;
-    virtual bool isClusterStrict(CellInfo *cell) const = 0;
+    virtual Loc getClusterOffset(const CellInfo *cell) const = 0;
+    virtual bool isClusterStrict(const CellInfo *cell) const = 0;
     virtual bool getClusterPlacement(ClusterId cluster, BelId root_bel,
                                      std::vector<std::pair<CellInfo *, BelId>> &placement) const = 0;
     // Flow methods
--- a/common/arch_api.h
+++ b/common/arch_api.h
@@ -139,6 +139,13 @@ template <typename R> struct ArchAPI : BaseCtx
     virtual typename R::CellTypeRangeT getCellTypes() const = 0;
     virtual typename R::BelBucketRangeT getBelBuckets() const = 0;
     virtual typename R::BucketBelRangeT getBelsInBucket(BelBucketId bucket) const = 0;
+    // Cluster methods
+    virtual CellInfo *getClusterRootCell(ClusterId cluster) const = 0;
+    virtual ArcBounds getClusterBounds(ClusterId cluster) const = 0;
+    virtual Loc getClusterOffset(ClusterId cluster, CellInfo *cell) const = 0;
+    virtual bool isClusterStrict(CellInfo *cell) const = 0;
+    virtual bool getClusterPlacement(ClusterId cluster, BelId root_bel,
+                                     std::vector<std::pair<CellInfo *, BelId>> &placement) const = 0;
     // Flow methods
     virtual bool pack() = 0;
     virtual bool place() = 0;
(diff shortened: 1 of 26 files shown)


acomodi (commented):
What might be missing is a description of how to generate and fill the various clusters in the first place. I imagine there will be a step prior to placement, that will build and generate the set of possible clusters, which will have placement priority over the single cells which are not part of any cluster. This step will group up all the cells that may or need to belong to a cluster, and this will be determined by the rules in the device database.
E.g. (cluster rule: `CARRY4.COUT connects to CARRY4.CIN`, where CARRY4 need are two different cells ofc).

There might be a definition of an API method to add cells to a cluster and to generate an initial cluster with a root cell, if the cell is a good candidate to be the root one.

> IO-IOLOGIC connectivity which may be better defined by the routing graph than by relative coordinates

By the above you mean getting the information through the connectivity of the two pins in two cells that need to be placed in the same cluster right? For instance, by the carry chain rule above, we know that a COUT is always connected to CIN, so the routing graph is will be used to get the location or the offset of the next CARRY4 in the chain?

gatecat:
> There might be a definition of an API method to add cells to a cluster and to generate an initial cluster with a root cell, if the cell is a good candidate to be the root one.

The idea of this new approach is that the Arch API only provides methods for querying clusters. Creating clusters, and the backing data structures to store them, will be left down to the arches themselves. There will be a `BaseArch` default implementation similar to the existing structures but the interchange arch will use its own more complex structures.

gatecat:
> By the above you mean getting the information through the connectivity of the two pins in two cells that need to be placed in the same cluster right? For instance, by the carry chain rule above, we know that a COUT is always connected to CIN, so the routing graph is will be used to get the location or the offset of the next CARRY4 in the chain?

Yes, this would be one data source the arch could use to provide the cluster placement. This way the discontinuities in the fabric can be considered, in a way that fixed offsets doesn't allow.

kowalewskijan:
I wonder how the carry-chain split might be solved with the cluster solution. In the sense, we have a design with a carry-chain exceeding the coords of a single carry-chain cluster. We could either have one cluster and an algo to determine how the end of one carry-chain is connected to the beginning of the next one (probably by picking the adjacent cell to the root cell?) or we might have a root/parent cluster which will point to the next cluster etc.

acomodi:
@kowalewskijan I guess we can provide rules on how to behave in such cases, and the algorithm can use different rules based on a priority order. For instance, keeping the carry chain as an example:
- the direct `COUT --> CIN` connection between two carries is the priority 0 rule;
- as a case where the rule cannot be applied anymore is hit, the next rule in the priority list is used, and, instead of a direct `COUT --> CIN`. In this case, the rule can be using the CO3 pin that gets to the general interconnect and place the cell in an adjacent location, depending also on the number of carries remaining in the cluster;
- from there, the first priority rule is applied again and so on

mkurc-ant:
IMO this API is fine. I assume that all the details regarding cluster generation and then constraining their placement either as whole clusters or individual cells is to be handled by arch specific code.

Also as the others said it would be nice to have an API for cluster splitting and merging. But lets focus on one thing at a time.

gatecat:
There are two separate issues with the Xilinx carry chain, that I think are confusing things.

1. where the carry chain crosses the `HCLK` row, the relative y delta is 2 rather than 1 - see screenshot below. This is something that the new API is explicitly designed to deal with; as instead of fixed x and y offsets exact bels are given by `getClusterPlacement`. To reiterate, the offsets and cluster size in `getClusterBounds` and `getClusterOffset` are approximations to guide analytical placement and do not need to take into account edge cases like this; detail placement will use the bels given by `getClusterPlacement` which no longer need to follow any regular structure.
Screenshot from 2021-04-28 10-40-48

2. dealing with carry chains that are too long to be placed using dedicated routing. In this case, clusters longer than the maximum size should never be being created by the arch in the first place; this is not something for this API to deal with. In the interchange format, we might want a maximum cluster size type field if we don't want to deal with this by interpreting the routing graph.

Hopefully this clears things up a bit?

PR #689, opened 2021-04-29
ECP5 ALU54B placement support

adamgreig:
This PR uses cellGroups to allow nextpnr to place an ECP5 ALU54B correctly adjacent to the connected MULT18X18Ds, allowing use of the ALU without having to manually place all components.

Manual placement of both the ALU and MULT still works, but we detect trying to manually place only the MULT(s) or only the ALU and error out, since I found only specifying half the locations did not work with the placer.

There's a couple other relevant commits included:

* 9538954 adds a missing parameter `REG_OPCODEOP1_1_CLK` which is already in the prjtrellis db and yosys cell
* d4c6882 adds the placement constraints described above
* b6c608e adds a few checks to the placer to detect invalid MULT+ALU connections; all these scenarios are rejected outright by Diamond but currently ignored by nextpnr, though I'm not sure if the resulting bitstream would be valid anyway since many are fixed connections
* d3a6cf3 fixes an issue where if the output registers are set, the output register is always driven by the MULT even if an ALU is in use

I'll probably have some more changes to come (here and to the prjtrellis db), but this PR is enough to get automatically placed and functional multiply-accumulate working, and hopefully is easier to review in isolation.

I've tested this on an LFE5U-45F successfully and also compared a number of bitstreams to what Diamond generates; there are still a few bits different but mostly insignificant (settings for unused clocks, the ALU OPCODE inverters which nextpnr doesn't use, etc).

### DIFF
--- a/ecp5/arch.cc
+++ b/ecp5/arch.cc
@@ -574,6 +574,11 @@ bool Arch::place()
         PlacerHeapCfg cfg(getCtx());
         cfg.criticalityExponent = 4;
         cfg.ioBufTypes.insert(id_TRELLIS_IO);
+
+        cfg.cellGroups.emplace_back();
+        cfg.cellGroups.back().insert(id_MULT18X18D);
+        cfg.cellGroups.back().insert(id_ALU54B);
+
         if (!placer_heap(getCtx(), cfg))
             return false;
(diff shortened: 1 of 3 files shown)


gatecat:
Thanks!

PR #693, opened 2021-05-04
interchange: Add more global cell info

gatecat:
This is hopefully the last prerequisite PR before actual global routing can be added to the interchange arch.

c.f. SymbiFlow/python-fpga-interchange#80


gatecat:
Now superseded by #694

PR #694, opened 2021-05-04
interchange: Initial global routing implementation

gatecat:
This currently implements very simple BFS-based global routing for clock buffer outputs (other features are WIP). It produces plausible results on the counter test but hits some issues on the RAM test relating to site pin conflicts, currently.

It's on top of #693.


gatecat:
It appears the issues with the ram test aren't actually specific to this PR but affect master too, see SymbiFlow/nextpnr#287

gatecat:
## Results with the RAM test

Before this patch, skew on critical path =  -1.093ns
Screenshot from 2021-05-04 15-17-57

After this patch, skew on critical path = -0.018ns
Screenshot from 2021-05-04 15-18-35

PR #696, opened 2021-05-06
nexus: Promote CE/LSR globals

gatecat:
This should be one of several optimisations to improve current nexus route-ability issues. It's currently a WIP, it will fail altogether on some designs due to PCLKCIB congestion (this is something of a latent defect that this patch will force to be addressed...)


Lunaphied:
Closing as stale - no meaningful activity in 5 years (last commit Sep 2021), was already in draft and never picked up.

PR #697, opened 2021-05-06
router2: Dynamicly expand bounding box based on congestion

gatecat:
Also, use the net bounding box rather than the arc bounding box to encourage more resource re-use. This might still need some more tweaks, but it is hopefully a not-worsening overall.


tcal-x:
@gatecat I'm seeing a huge improvement with this!

PR #700, opened 2021-05-10
interchange: arch: do not allow site pips within sites

acomodi:
During general routing, the only site pips that can be allowed are those which connect a site wire to the routing interface. Or, in more general terms, all the PIPs connecting a driver to the routing interface.

This patch though might be too restrictive when dealing with architectures that require more than one site PIPs to route from a driver within a site to the routing interface (which is something that should be allowed in the interchange).

After all, backtracking to validate a path might still be required, as https://github.com/YosysHQ/nextpnr/pull/695 is not enough to cover the faulty edge-cases, where a series of PIPs were used to completely route thru a site.

### DIFF
--- a/fpga_interchange/site_router.cc
+++ b/fpga_interchange/site_router.cc
@@ -1070,6 +1070,47 @@ static void block_lut_outputs(SiteArch *site_arch,
     }
 }
 
+// Recursively visit downhill PIPs until a SITE_PORT_SINK is reached.
+// Marks all PIPs for all valid paths.
+static bool visit_downhill_pips(const SiteArch *site_arch, const SiteWire &site_wire, std::vector<PipId> &valid_pips) {
+    bool valid_path_exists = false;
+    for (SitePip site_pip : site_arch->getPipsDownhill(site_wire)) {
+        const SiteWire &dst_wire = site_arch->getPipDstWire(site_pip);
+        if (dst_wire.type == SiteWire::SITE_PORT_SINK) {
+            valid_pips.push_back(site_pip.pip);
+            return true;
+        }
+
+        bool path_ok = visit_downhill_pips(site_arch, dst_wire, valid_pips);
+        valid_path_exists |= path_ok;
+
+        if (path_ok) {
+            valid_pips.push_back(site_pip.pip);
+        }
+    }
+
+    return valid_path_exists;
+}
+
+// Checks all downhill PIPs starting from driver wires.
+// All valid PIPs are stored and returned in a vector.
+static std::vector<PipId> check_downhill_pips(Context *ctx, const SiteArch *site_arch) {
+    auto &cells_in_site = site_arch->site_info->cells_in_site;
+
+    std::vector<PipId> valid_pips;
+    for (auto &net_pair : site_arch->nets) {
+        NetInfo *net = net_pair.first;
+        const SiteNetInfo *site_net = &net_pair.second;
+
+        if (net->driver.cell && cells_in_site.count(net->driver.cell)) {
+            const SiteWire &site_wire = site_net->driver;
+
+            visit_downhill_pips(site_arch, site_wire, valid_pips);
+        }
+    }
+    return valid_pips;
+}
+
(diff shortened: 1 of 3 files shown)


gatecat:
One option would be to store legal site pips that are in the input/output cones of bels used by the net after site routing, then the check doesn't need to do the backtracking itself and just consults that set of pips.

acomodi:
>  This will likely cause some failures as I don't believe the site router pre binds routing in all cases 

Yeah, I believe this too. I imagine that at the moment this is not causing any failures as we do not hit edge cases in the tests. In fact AFAIU, this check was to allow, for instance, routing from a FF output through an OUTMUX and then to the general interconnect. If there were other site pips to be traversed, than the path to route from the FF to the output would be illegal, even though it should be explored.

> One option would be to store legal site pips that are in the input/output cones of bels used by the net after site routing, then the check doesn't need to do the backtracking itself and just consults that set of pips.

Hmm right. If I understand this correctly, this would mean checking all downhill site pips from a BEL output (as the input should be already pre-binded) that lead to the site output ports and mark all those site PIPs as valid, correct?

acomodi:
@gatecat I have implemented the suggested solution, all the driver wires downhill PIP's that enable reaching an output site port are stored and than visited when determining a validity of a PIP.

I think the solution might be improved to store some data in the `site_router_cache` as there may be quite some repetitive pattern. In fact the valid PIPs will always be the same for a given site wire.

gatecat:
I think it should be possible to improve performance by doing it when we bind site routing, in between placement and routing, rather than throughout placement at the point of checking the validity of site routing?

But yeah, overall the site routing stuff could doubtless be a lot faster.

acomodi:
At the moment the list of valid PIPs is already built in between placement and routing, during the bindSiteRouting step, so it shouldn't impact placement run-time.

I'll investigate some optimizations to this patch, and also check the impact on the murax design (and whether all the edge cases are solved).

gatecat:
Ah, sorry, I misread the misread the GitHub diff, that's fine then.

fpga_interchange/site_router.cc
```diff
@@ -1070,6 +1070,52 @@ static void block_lut_outputs(SiteArch *site_arch,
...
+        const SiteNetInfo *site_net = &net_pair.second;
+
+        if (net->driver.cell && cells_in_site.count(net->driver.cell)) {
+            const SiteWire &site_wire = site_net->driver;
+
+            bool result = visit_downhill_pips(site_arch, site_wire, valid_pips);
+
+            // There needs to be at least one path that leads
+            // from a cell driver to a site output sink
+            NPNR_ASSERT(result);
```
gatecat, line 1112: this isn't necessarily true in two cases:
 - 1. all of the users of the net are contained within the site
 - 2. the net has no users (this is a trivial case and just needs `&& !net->users.empty()` added to the if statement)

acomodi: Added fix for case ii.

For i. I might be missing something here, but I guess that, if a site wire corresponding to a net driver got assigned to a net, that would mean that that net was already handled during site routing, therefore there is no need to check for it's valid PIPs.

gatecat: > For i. I might be missing something here, but I guess that, if a site wire corresponding to a net driver got assigned to a net, that would mean that that net was already handled during site routing, therefore there is no need to check for it's valid PIPs

It's true that we don't care about the set of valid PIPs; but that means we also need to ignore the net and not have an assertion that is likely to fail (because it doesn't matter if we find PIPs to a site pin or not).

acomodi: Hmm, right, I wonder if this assertion needs to be there at all at this point. As the router would fail as well if there are no PIPs to route to a site pin, with or without this assertion. Given that there are drivers within a site that may not have routes to a site pin at all, and these drivers are not handled by the router (I think).

gatecat: Yeah I think it's safest just to remove the assertion for now.

gatecat (approved):

PR #702, opened 2021-05-11
interchange: Fix bounding box computation

gatecat:
@acomodi this should fix the most acute part of the regression coming from #697 - at least it does for the RAM test, I haven't actually tried Murax. There are definitely some other issues with routing though, that I'm continuing to investigate.

### DIFF
--- a/common/router2.cc
+++ b/common/router2.cc
@@ -175,7 +175,7 @@ struct Router2
                     nets.at(i).bb.x0 = std::min(nets.at(i).bb.x0, ad.bb.x0);
                     nets.at(i).bb.x1 = std::max(nets.at(i).bb.x1, ad.bb.x1);
                     nets.at(i).bb.y0 = std::min(nets.at(i).bb.y0, ad.bb.y0);
-                    nets.at(i).bb.y1 = std::max(nets.at(i).bb.x1, ad.bb.y1);
+                    nets.at(i).bb.y1 = std::max(nets.at(i).bb.y1, ad.bb.y1);
                 }
                 // Add location to centroid sum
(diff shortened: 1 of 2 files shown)


gatecat:
> Regarding the fix itself, do I understand correctly that, without this, the bounding box was always bound to the x0 = y0 = 0 point, therefore being quite large w.r.t. the required bb?

Correct - this shouldn't have been causing such terrible performance on its own right, so there are probably still some other issues going on too (like bad lookahead, and perhaps further bounding box issues).

gatecat:
I'm still seeing some other issues here with Murax, in particular a large number of routes failing the bounding box test, and am investigating.

gatecat:
The last commit seems to fix most of the problems - routing is still slower than it should be, I suspect that the last remaining issues are around lookahead (which probably isn't worth spending too much time on until we've made some more progress with timing data), bad bounding boxes in certain complex cases like IO, and poor placement quality (which is on my longer term TODO list)

gatecat:
> Yeah, I see SA placement is currently what's taking most of the time. I also believe that it could greatly improve as soon as we support carry chains as well, but at the moment I think it is good enough, thanks for the fixes.

afaik the current bottlenecks according to litghost were the amount of LUT mapping that the placer does, and constantly resetting the cell/bel pin maps. there should be low hanging fruit to improve both of these once we have a bit more spare time and ideally a test framework

gatecat:
for the record, these is the most expensive parts of placement legalisation (will apply to both SA and HeAP) according to callgrind:

Screenshot from 2021-05-11 14-45-50

PR #703, opened 2021-05-11
WIP: fpga-interchange bels chains support

kowalewskijan:
This PR provides support for fpga-interchange BELs chains using cluster API.

TODOs:
- [x] - prepare functionality to build clusters of given chain type
- [x] - provide cluster API for placer
- [ ] - test and enable carry chains

Related PRs/Issue:
- https://github.com/SymbiFlow/python-fpga-interchange/pull/70
- https://github.com/SymbiFlow/fpga-interchange-schema/pull/45
- https://github.com/SymbiFlow/nextpnr/issues/262

### DIFF
--- a/fpga_interchange/arch_pack_chains.cc
+++ b/fpga_interchange/arch_pack_chains.cc
@@ -96,6 +96,12 @@ bool Arch::getClusterPlacement(ClusterId cluster, BelId root_bel,
             if(bel_found) {
                 break;
             }
+            // Revert next_loc changes
+            if (coord == CHAIN_X_COORD) {
+                next_loc.x -= cfg.second;
+            } else {
+                next_loc.y -= cfg.second;
+            }
         }
         if (!bel_found) {
             // TODO: improve this message
--- a/fpga_interchange/arch_pack_chains.cc
+++ b/fpga_interchange/arch_pack_chains.cc
@@ -208,9 +208,11 @@ void Arch::prepare_cluster(const BelChainPOD *chain)
             if (snk_cell_type != cell->type || cell->ports.find(snk_cell_port) == cell->ports.end()) {
                 continue;
             }
-
             PortRef driver = cell->ports[snk_cell_port].net->driver;
-            if (driver.cell->type == id_GND || driver.cell->type == id_VCC) {
+            if (driver.cell == nullptr) {
+                goto handle_root;
+            } else if (driver.cell->type != cell->type) {
+            handle_root:
                 // We hit a root cell
                 cell->cluster.set(ctx, cell->name.str(ctx));
                 roots.push_back(cell);
--- a/fpga_interchange/arch_pack_chains.cc
+++ b/fpga_interchange/arch_pack_chains.cc
@@ -1,0 +1,237 @@
+/*
+ *  nextpnr -- Next Generation Place and Route
+ *
+ *  Copyright (C) 2021  Symbiflow Authors
+ *
+ *
+ *  Permission to use, copy, modify, and/or distribute this software for any
+ *  purpose with or without fee is hereby granted, provided that the above
+ *  copyright notice and this permission notice appear in all copies.
+ *
+ *  THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ *  WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ *  MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ *  ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ *  WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ *  ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ *  OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ *
+ */
+
+#include "design_utils.h"
+#include "log.h"
+#include "nextpnr.h"
+#include "arch.h"
+#include "util.h"
+
+#include <boost/algorithm/string.hpp>
+#include <queue>
+
+NEXTPNR_NAMESPACE_BEGIN
+
+CellInfo* Arch::getClusterRootCell(ClusterId cluster) const
+{
+    NPNR_ASSERT(cluster != ClusterId());
+    CellInfo *ci = nullptr;
+    for (auto cell : sorted(getCtx()->cells)) {
+        ci = cell.second;
+        if (ci->cluster == cluster && cluster_roots.find(ci->name) != cluster_roots.end()) {
+            break;
+        }
+    }
+    return ci;
+}
+
+bool getClusterPlacement(ClusterId cluster, BelId root_bel,
+                         std::vector<std::pair<CellInfo *, BelId>> &placement) const
+{
+    // Load coord config if available?
+    // Place cluster
+    NPNR_ASSERT_FALSE("unimplemented");
+}
+
+// TODO
+//ArcBounds getClusterBounds(ClusterId cluster) const
+//{
+//}
+//Loc getClusterOffset(const CellInfo *cell) const
+//{
+//}
+//bool isClusterStrict(const CellInfo *cell) const
+//{
+//}
+
+void dump_chains(const ChipInfoPOD *chip_info, Context *ctx)
+{
+    std::unordered_map<IdString, const BelChainPOD *> bel_chain_prototypes;
+    for (size_t i = 0; i < chip_info->bel_chains.size(); ++i) {
+        const auto &bel_chain = chip_info->bel_chains[i];
+        IdString bel_chain_name(bel_chain.name);
+        bel_chain_prototypes.emplace(bel_chain_name, &bel_chain);
+        log_info("BEL chain '%s' loaded! Parameters:\n", bel_chain_name.c_str(ctx));
+        log_info("  - sites:\n");
+            for (auto site : bel_chain.sites) {
+                IdString site_name(site);
+                log_info("      - %s\n", site_name.c_str(ctx));
+            }
+        log_info("  - cells:\n");
+            for (auto cell : bel_chain.cells) {
+                IdString cell_name(cell);
+                log_info("      - %s\n", cell_name.c_str(ctx));
+            }
+        log_info("  - patterns:\n");
+            for (auto pattern : bel_chain.chain_patterns) {
+                IdString source(pattern.source);
+                IdString sink(pattern.sink);
+                log_info("      - %s -> %s\n", source.c_str(ctx), sink.c_str(ctx));
+            }
+        log_info("  - coord_configs:\n");
+            for (auto cfg : bel_chain.chain_coord_configs) {
+                log_info("      - coord: %d | step: %d\n", cfg.coord, cfg.step);
+            }
+    }
+}
+
+void Arch::prepare_cluster(const BelChainPOD *chain)
+{
+    Context *ctx = getCtx();
+
+    // Get chainable cells
+    std::vector<CellInfo *> chainable_cells;
+    for (auto cell : sorted(ctx->cells)) {
+        CellInfo *ci = cell.second;
+        for (auto chain_cell : chain->cells) {
+            IdString chain_cell_type(chain_cell);
+            if (ci->type == chain_cell_type) {
+                chainable_cells.push_back(ci);
+            }
+        }
+    }
+
+    // Prepare sources and sinks pairs from provided patterns
+    // pair: <cell_type, cell_port>
+    std::vector<std::pair<IdString, IdString>> sources, sinks;
+    std::string delimiter = ".";
+    for (auto pattern : chain->chain_patterns) {
+        IdString source(pattern.source);
+        IdString sink(pattern.sink);
+        IdString source_type;
+        IdString sink_type;
+        std::string source_str = source.c_str(ctx);
+        std::string sink_str = sink.c_str(ctx);
+        std::string source_type_str;
+        std::string sink_type_str;
+
+        source_type_str = source_str.substr(0, source_str.find(delimiter));
+        source_str = source_str.substr(source_str.find(delimiter) + 1, source_str.back());
+        source_type.set(ctx, source_type_str);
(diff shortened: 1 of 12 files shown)


fpga_interchange/arch_pack_chains.cc
```diff
@@ -0,0 +1,237 @@
...
+#include <boost/algorithm/string.hpp>
+#include <queue>
+
+NEXTPNR_NAMESPACE_BEGIN
+
+CellInfo* Arch::getClusterRootCell(ClusterId cluster) const
+{
+    NPNR_ASSERT(cluster != ClusterId());
+    CellInfo *ci = nullptr;
+    for (auto cell : sorted(getCtx()->cells)) {
```
gatecat, line 36: this is likely to introduce O(n²) performance - there should be a way of going from `ClusterId` to a root cell much faster than this (in the other arches `ClusterId` is actually the name of the root cell, so it ends up being a lookup in `ctx->cells`)

fpga_interchange/arch_pack_chains.cc
```diff
@@ -0,0 +1,237 @@
...
+        IdString source(pattern.source);
+        IdString sink(pattern.sink);
+        IdString source_type;
+        IdString sink_type;
+        std::string source_str = source.c_str(ctx);
+        std::string sink_str = sink.c_str(ctx);
+        std::string source_type_str;
+        std::string sink_type_str;
+
+        source_type_str = source_str.substr(0, source_str.find(delimiter));
```
gatecat, line 125: it might be a bit cleaner to do the splitting, and conversion to IdString, inside the bba generation rather than at runtime

fpga_interchange/chipdb.h
```diff
@@ -346,6 +371,7 @@ NPNR_PACKED_STRUCT(struct ChipInfoPOD {
     RelSlice<PackagePOD> packages;
     RelSlice<WireTypePOD> wire_types;
     RelSlice<GlobalCellPOD> global_cells;
+    RelSlice<BelChainPOD> bel_chains;
```
gatecat, line 374: chipdb changes should always be bumping the chipdb version, too: https://github.com/YosysHQ/nextpnr/blob/ea45971dcc81983d5c952e36c762b1f922dfe9f7/fpga_interchange/chipdb.h#L37

kowalewskijan:
@clavin-xlnx : I tried recently to build basys3-counter with carries, but it seems that there is a problem with finding proper BEL type for cells from a cluster (carry chain in this case). So for the cell: `$auto$alumacc.cc:485:replace_alu$1388.slice[0].carry4` the placer_heap tries to find available BELs using [getBelsForCellType](https://github.com/YosysHQ/nextpnr/blob/master/common/fast_bels.h#L147-L160). The results contain only BELs of type `OR2L` and `IDDR` instead of expected `CARRY4`.

It seems that current implementation of [isValidBelForCellType](https://github.com/YosysHQ/nextpnr/blob/master/fpga_interchange/arch.h#L792-L795) for the interchange checks if BEL category is `BEL_CATEGORY_LOGIC`. I did some debugging and it appeared that `CARRY4` BEL is defined as `BEL_CATEGORY_SITE_PORT`. AFAIK the BEL category comes from RapidWright, hence the question why the `CARRY4` is categorized as `SITE_PORT` rather than `LOGIC`? Is it expected to be like that, therefore we should change the validation process somehow to better handle these categories?

gatecat:
`CARRY4` definitely shouldn't be a site port. I will investigate where this is going wrong once I've finished what I'm currently looking at.

gatecat:
@kowalewskijan I've attempted to test this using:

```
    for (auto bel : getBels()) {
        if (getBelType(bel) != id("CARRY4"))
            continue;
        log_info("%s %d\n", getCtx()->nameOfBel(bel), bel_info(chip_info, bel).category);
    }
```

added to `Arch::init()` and all the bels have a category of 0 (`BEL_CATEGORY_LOGIC`) as expected.

I think your chipdb is messed up, perhaps because your bba generator and nextpnr data structures got out of sync (this is one of the reasons why it's important always to bump `kExpectedChipInfoVersion` when making any change)

fpga_interchange/arch_pack_chains.cc
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    log_info("Cluster placement for root cell: '%s'\n", root_cell->name.c_str(ctx));
+
+    return true;
+}
+
+ArcBounds Arch::getClusterBounds(ClusterId cluster) const
+{
+    ArcBounds bounds(0, 0, 0, 0);
```
gatecat, line 117: This is going to hit a similar bug to the one I fixed in 7a1a95a2d6ba7c9383f75016409aea1f613db547

fpga_interchange/arch_pack_chains.cc
```diff
@@ -0,0 +1,294 @@
...
+
+    for (auto cell : sorted(getCtx()->cells)) {
```
gatecat, line 119: I think you can look up in `cluster_to_coord_configs`, you don't need to iterate over all the cells

fpga_interchange/arch_pack_chains.cc
```diff
@@ -0,0 +1,294 @@
...
+        if (ci->cluster == cluster) {
+            bounds.x0 = std::min(bounds.x0, cell_loc.x);
+            bounds.y0 = std::min(bounds.y0, cell_loc.y);
+            bounds.x1 = std::max(bounds.x1, cell_loc.x);
+            bounds.y1 = std::max(bounds.y1, cell_loc.y);
+        }
+    }
+}
+
+Loc Arch::getClusterOffset(const CellInfo *cell) const
```
gatecat, line 131: This function should provide an approximate offset - it doesn't have to be exactly correct, it's just a hint to the placer - even if the cluster isn't yet placed and therefore `cell->bel` is still `BelId`()`

kowalewskijan:
@gatecat : during cluster placement I've stumbled upon the case in which the root cell driver is unplaced - in my case it was GND. I hacked it for now by allowing GND and VCC to be unplaced. I was thinking about it and should we just store which drivers were unplaced at the placement time and then check if they were actually placed after the placement flow? ([Related code](https://github.com/YosysHQ/nextpnr/pull/703/files#diff-58d51ad8ce8e0bbda06fdac1343c9676a2d2cea94443380c58602c2bb855c006))

gatecat:
While there might be another issue going on here; I think in this case clusters should never be including GND/VCC - similar to anything else that's not a carry but is driving carry in.

kowalewskijan:
In my case GND is a driver for a root cell but it's not a part of cluster in a sense GND cell does not have cluster_id. Yet still the check in the placer will fail on [isBelLocationValid](https://github.com/YosysHQ/nextpnr/blob/master/common/placer_heap.cc#L1035) because the GND is not placed at the time.
Here are some logs (I added cluster_id info about driver additionally just to be sure):
```
Info: Cluster '$auto$alumacc.cc:485:replace_alu$1388.slice[0].carry4' [CARRY4] Placement:                                                                                                                          
Info: Cell '$auto$alumacc.cc:485:replace_alu$1388.slice[0].carry4' placed on (14, 133, 48)                                                                                                                         
Info: Cell '$auto$alumacc.cc:485:replace_alu$1388.slice[1].carry4' placed on (14, 134, 48)               
Info: Cell '$auto$alumacc.cc:485:replace_alu$1388.slice[2].carry4' placed on (14, 135, 48)               
Info: Cell '$auto$alumacc.cc:485:replace_alu$1388.slice[3].carry4' placed on (14, 136, 48)               
Info: Cell '$auto$alumacc.cc:485:replace_alu$1388.slice[4].carry4' placed on (14, 137, 48)               
Info: Cell '$auto$alumacc.cc:485:replace_alu$1388.slice[5].carry4' placed on (14, 138, 48)
Info: Cell '$auto$alumacc.cc:485:replace_alu$1388.slice[6].carry4' placed on (14, 139, 48)
Info: BEL SLICE_X4Y22.SLICEL/CARRY4 is not valid because driver cell $__gnd_6/G is not placed | cluster id of driver:
```
If I get it right, the[ constants are placed](https://github.com/YosysHQ/nextpnr/blob/master/fpga_interchange/arch.cc#L842) just before routing, so until then the GND driver for a carry-chain root cell will remain unplaced, is it correct?

gatecat:
I think you need to modify the `isBelLocationValid` check so it isn't trying to check the GND in the first place.

kowalewskijan:
@gatecat : currently I'm able to place the carry-chain cluster, however the placement is taking a tremendous time for the counter example.
Here is the report:
```
Info: HeAP Placer Time: 497.38s                                                                                                                                                                                    
Info:   of which solving equations: 0.00s                                                                                                                                                                          
Info:   of which spreading cells: 0.03s                                                                                                                                                                            
Info:   of which strict legalisation: 496.40s
```
It seems that the cluster is placed and then the placement of DFFs is taking most of the time (it loops in [find_solution_via_backtrack](https://github.com/YosysHQ/nextpnr/blob/master/fpga_interchange/site_router.cc#L606-L684)). In regard to your comment about the bounding box fix, I rebased the branch to be up-to-date just in case. Do you have any hint what may be happening?

gatecat:
Hmm, site routing has always been a performance weak spot but never this slow. It's quite possible that carries are simply hitting a case that it was not thought of, unfortunately I also don't fully understand the site routing code and its various paths so I can't think of a fix straight away.

gatecat:
I'm trying to test this out locally, using the `antmicro/bels-chains-support` branch of `python-fpga-interchange`, but `counter_basys3` is failing with:

```
Info: Cluster '$auto$alumacc.cc:485:replace_alu$1419.genblk1.slice[0].genblk1.carry4' [CARRY4] Placement:
Info: Cell '$auto$alumacc.cc:485:replace_alu$1419.genblk1.slice[0].genblk1.carry4' placed on (13, 129, 16)
Info: Cell '$auto$alumacc.cc:485:replace_alu$1419.genblk1.slice[1].genblk1.carry4' placed on (13, 128, 16)
Info: Cell '$auto$alumacc.cc:485:replace_alu$1419.genblk1.slice[2].genblk1.carry4' placed on (13, 127, 16)
Info: Cell '$auto$alumacc.cc:485:replace_alu$1419.genblk1.slice[3].genblk1.carry4' placed on (13, 126, 16)
Info: Cell '$auto$alumacc.cc:485:replace_alu$1419.genblk1.slice[4].genblk1.carry4' placed on (13, 125, 16)
Info: Cell '$auto$alumacc.cc:485:replace_alu$1419.genblk1.slice[5].genblk1.carry4' placed on (13, 124, 16)
Info: Cell '$auto$alumacc.cc:485:replace_alu$1419.genblk1.slice[6].genblk1.carry4' placed on (13, 123, 16)
terminate called after throwing an instance of 'nextpnr_fpga_interchange::assertion_failure'
  what():  Assertion failure: site_info->is_wire_in_site(site_wire) (/home/david/nextpnr/fpga_interchange/site_arch.impl.h:73)
```

any idea what this could be?

gatecat:
I think a lot of the conflicts that are slowing down the site router are down to the LUT->carry connections, stuff like

`Net conflict binding wire SLICE_X1Y54.SLICEL/C5LUT_O5 to net GLOBAL_LOGIC0, conflicts with net $auto$alumacc.cc:485:replace_alu$1419.Y[26]`

perhaps the LUTs driving the carry chain should also be included in the cluster

kowalewskijan:
> terminate called after throwing an instance of 'nextpnr_fpga_interchange::assertion_failure'
> what():  Assertion failure: site_info->is_wire_in_site(site_wire) (/home/david/nextpnr/fpga_interchange/site_arch.impl.h:73

This one refers to missing CARRY4 ports. I pushed the [hack](https://github.com/SymbiFlow/python-fpga-interchange/pull/70/commits/c3273def479d41a6ace8f907b66747b7e01aa821) to the python-fpga-interchange as a temporary solution for that.

> I think a lot of the conflicts that are slowing down the site router are down to the LUT->carry connections

Thanks, I'll look into that.

gatecat:
Hmm, I think this is probably hitting a bit of a weak spot in nextpnr at the moment.

One option would be to explicitly create and constrain route through LUTs when the driver is not a LUT, or not a LUT that can be packed. This is how nextpnr-xilinx handled this case, and makes sure the LUT bels are explicitly blocked.

Otherwise this is something that will need more long term thought and probably a rewrite of the placer legalisation, which might be a way off.

kowalewskijan:
I was able to generate a physical netlist for the counter example with some workarounds and I think they might be discussed:

Regarding constraining LUTs: I came up with a techmap solution, so basically I mapped a `CARRY4` to `LUT1`s + `CARRY4` ([link](https://github.com/YosysHQ/nextpnr/pull/703/files#diff-3e9a3ce17267cc5152837659f02affbfe357cea40a55302f07ad800497678cd1R35-R59)). This improved the placement time consumption - it dropped down from ~10min to ~30sec, however the placement refinement still takes some time. There are also some routing issues like e.g. 
* LUTs can be placed outsite the CARRY site
* LUTs equations are not checked in the process

Another issue is associated with handling the `GND->CIN` connection. Yosys generates a carry cell with a constant connected to the `CIN` port (root cell of a carry-chain in this case), this resulted in the unroutable design. Currently I solved this again with a techmap which drops the `CIN` port when the constant is connected to the port. In the Nextpnr the disconnected port can cause some problems, so for now I assumed that we may skip the disconnected driver cell if it is a part of the cluster ([related code](https://github.com/YosysHQ/nextpnr/pull/703/files#diff-b81a8cd7a33e6001bab027dd942d9f8ae73b2e1666c4922d28c0169597564d21R149-R276)) (we might need to indicate the skip better)

gatecat:
> Another issue is associated with handling the GND->CIN connection. Yosys generates a carry cell with a constant connected to the CIN port (root cell of a carry-chain in this case), this resulted in the unroutable design. Currently I solved this again with a techmap which drops the CIN port when the constant is connected to the port. In the Nextpnr the disconnected port can cause some problems, so for now I assumed that we may skip the disconnected driver cell if it is a part of the cluster (related code) (we might need to indicate the skip better)

The correct way to solve this is probably to have a synthesis-time pass that moves CIN to CYINIT. Then ground should be routeable by virtue of the site pip:
Screenshot from 2021-05-20 18-53-02

fpga_interchange/dedicated_interconnect.h
```diff
@@ -31,6 +31,14 @@
 
 NEXTPNR_NAMESPACE_BEGIN
 
+enum SinkOnNetStatus
```
gatecat, line 34: I'd recommend adding a prefix to these enums, to avoid conflicts with anything else (`VALID` in particular being a common name).

gatecat:
Closing, as apparently superseded by #720. Will reopen if that isn't the case.

PR #710, opened 2021-05-16
mistral: Use MLABs as if they're LABs (for now)

Ravenslofty:
Much of this is just mechanically replacing hardcoded `CycloneV::LAB`.

This reduces routing time for attosoc by quite a bit, presumably due to the ALMs being closer together.


Ravenslofty:
I don't think anybody other than gatecat is watching this, but as a node for the future, the big blocker on this is that the LUT bits of the MLAB are permuted, and attempts to discover this are presently not going so well.

PR #720, opened 2021-06-02
interchange: enable clusters support

acomodi:
This PR is based on the ideas of https://github.com/YosysHQ/nextpnr/pull/703, and aims at providing an implementation of the clustering support, so that the placer can be guided by information defined in the chip info.

The current state of developement, combined with https://github.com/SymbiFlow/python-fpga-interchange/pull/91, is able to get a working counter test that uses carry-chains.

TODO:
- [x] Test murax design: currently takes quite a while to place if carry-chains are enabled.
- [x] Add LUT-to-FF clustering support to speed up placement
- [x] Iterate and refine the cluster chipinfo defintions to be more compact and cleaner
- [ ] Add comments

NOTE:
- The murax test design is not meant to be checked in, but is only useful in the various developement iterations.

### DIFF
--- a/fpga_interchange/arch.h
+++ b/fpga_interchange/arch.h
@@ -94,6 +94,15 @@ struct TileStatus
     PseudoPipModel pseudo_pip_model;
 };
 
+struct Cluster
+{
+    uint32_t index;
+    CellInfo *root;
+    std::vector<CellInfo *> cluster_nodes;
+    dict<IdString, IdString> cell_cluster_node_map;
+    dict<IdString, std::vector<std::pair<IdString, CellInfo *>>> cluster_node_cells;
+};
+
 struct Arch : ArchAPI<ArchRanges>
 {
     boost::iostreams::mapped_file_source blob_file;
@@ -258,6 +267,20 @@ struct Arch : ArchAPI<ArchRanges>
                 map_cell_pins(cell, mapping, /*bind_constants=*/false);
             }
             constraints.bindBel(tile_status.tags.data(), get_cell_constraints(bel, cell->type));
+
+            // Clean previous cell placement in tile
+            if (cell->bel != BelId()) {
+                TileStatus &prev_tile_status = get_tile_status(cell->bel.tile);
+                NPNR_ASSERT(prev_tile_status.boundcells[cell->bel.index] != nullptr);
+
+                const auto &prev_bel_data = bel_info(chip_info, cell->bel);
+                NPNR_ASSERT(prev_bel_data.category == BEL_CATEGORY_LOGIC);
+
+                get_site_status(prev_tile_status, prev_bel_data).unbindBel(cell);
+                prev_tile_status.boundcells[cell->bel.index] = nullptr;
+
+                constraints.unbindBel(prev_tile_status.tags.data(), get_cell_constraints(cell->bel, cell->type));
+            }
         } else {
             map_port_pins(bel, cell);
             // FIXME: Probably need to actually constraint io port cell/bel,
@@ -687,7 +710,14 @@ struct Arch : ArchAPI<ArchRanges>
 
     void place_iobufs(WireId pad_wire, NetInfo *net, const pool<CellInfo *, hash_ptr_ops> &tightly_attached_bels,
                       pool<CellInfo *, hash_ptr_ops> *placed_cells);
+
     void pack_ports();
+
+    // Clusters
+    void pack_cluster();
+    void prepare_cluster(const ClusterPOD *cluster, uint32_t index);
+    dict<ClusterId, Cluster> clusters;
+
     void decode_lut_cells();
 
     const GlobalCellPOD *global_cell_info(IdString cell_type) const;
@@ -821,10 +851,10 @@ struct Arch : ArchAPI<ArchRanges>
         }
         const TileStatus &tile_status = iter->second;
         const CellInfo *cell = tile_status.boundcells[bel.index];
+
         if (cell != nullptr) {
-            if (!dedicated_interconnect.isBelLocationValid(bel, cell)) {
+            if(cell->cluster == ClusterId() && !dedicated_interconnect.isBelLocationValid(bel, cell))
                 return false;
-            }
 
             if (io_port_types.count(cell->type)) {
                 // FIXME: Probably need to actually constraint io port cell/bel,
@@ -837,24 +867,21 @@ struct Arch : ArchAPI<ArchRanges>
                 return false;
             }
         }
+
         // Still check site status if cell is nullptr; as other bels in the site could be illegal (for example when
         // dedicated paths can no longer be used after ripping up a cell)
         auto &bel_data = bel_info(chip_info, bel);
-        return get_site_status(tile_status, bel_data).checkSiteRouting(getCtx(), tile_status);
+        bool site_status = get_site_status(tile_status, bel_data).checkSiteRouting(getCtx(), tile_status);
+
+        return site_status;
     }
 
-    // -------------------------------------------------
-
-    // TODO
-    CellInfo *getClusterRootCell(ClusterId cluster) const override { NPNR_ASSERT_FALSE("unimplemented"); }
-    ArcBounds getClusterBounds(ClusterId cluster) const override { NPNR_ASSERT_FALSE("unimplemented"); }
-    Loc getClusterOffset(const CellInfo *cell) const override { NPNR_ASSERT_FALSE("unimplemented"); }
-    bool isClusterStrict(const CellInfo *cell) const override { NPNR_ASSERT_FALSE("unimplemented"); }
+    CellInfo *getClusterRootCell(ClusterId cluster) const override;
+    ArcBounds getClusterBounds(ClusterId cluster) const override;
+    Loc getClusterOffset(const CellInfo *cell) const override;
+    bool isClusterStrict(const CellInfo *cell) const override;
     bool getClusterPlacement(ClusterId cluster, BelId root_bel,
-                             std::vector<std::pair<CellInfo *, BelId>> &placement) const override
-    {
-        NPNR_ASSERT_FALSE("unimplemented");
-    }
+                             std::vector<std::pair<CellInfo *, BelId>> &placement) const override;
 
     IdString get_bel_tiletype(BelId bel) const { return IdString(loc_info(chip_info, bel).name); }
 
--- a/fpga_interchange/arch.h
+++ b/fpga_interchange/arch.h
@@ -853,7 +853,7 @@ struct Arch : ArchAPI<ArchRanges>
         const CellInfo *cell = tile_status.boundcells[bel.index];
 
         if (cell != nullptr) {
-            if(cell->cluster == ClusterId() && !dedicated_interconnect.isBelLocationValid(bel, cell))
+            if (cell->cluster == ClusterId() && !dedicated_interconnect.isBelLocationValid(bel, cell))
                 return false;
 
             if (io_port_types.count(cell->type)) {
--- a/fpga_interchange/arch_pack_clusters.cc
+++ b/fpga_interchange/arch_pack_clusters.cc
@@ -193,9 +193,6 @@ bool Arch::getClusterPlacement(ClusterId cluster, BelId root_bel,
     const Context *ctx = getCtx();
     const Cluster &packed_cluster = clusters.at(cluster);
 
-    IdString GND = id("GND");
-    IdString VCC = id("VCC");
-
     CellInfo *root_cell = getClusterRootCell(cluster);
     if (!ctx->isValidBelForCellType(root_cell->type, root_bel))
         return false;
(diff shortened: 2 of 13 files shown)


fpga_interchange/arch.h
```diff
@@ -688,6 +702,16 @@ struct Arch : ArchAPI<ArchRanges>
     void place_iobufs(WireId pad_wire, NetInfo *net, const std::unordered_set<CellInfo *> &tightly_attached_bels,
                       std::unordered_set<CellInfo *> *placed_cells);
     void pack_ports();
+
+    // Chain packer
+    void pack_chains();
+    std::unordered_map<IdString, std::vector<IdString>> cluster_driver_cells;
```
gatecat, line 708: as a note, now that #718 has just been merged, this will need to be updated to use `dict` instead of `std::unordered_map`

fpga_interchange/arch_pack_clusters.cc
```diff
@@ -0,0 +1,496 @@
...
+
+                for (auto &cell : cluster_driver_in_out_cells.at(cluster)) {
+                    if (IdString(cell) == driver_cell->type && driver_cell_driver == next_cell) {
+                        cell_ok = true;
+                        break;
+                    }
+                }
+            }
+
+            if (!cell_ok or driver_cell == nullptr)
```
gatecat, line 206: please use `||` not `or`

fpga_interchange/arch_pack_clusters.cc
```diff
@@ -0,0 +1,496 @@
...
+        return false;
+
+    // Place cluster
+    // Prepare first loop iteration
+    CellInfo *next_cell = root_cell;
+    NetInfo *next_net = nullptr;
+    BelId next_bel = root_bel;
+    Loc next_loc = getBelLocation(next_bel);
+
+    std::vector<IdString> placed_bels;
```
gatecat, line 167: I think using `pool` (i.e.  the new `unordered_set` replacement) would make more sense than `std::vector`, as far as I can tell the only thing you do with this structure is check membership (which can be done significantly faster with a hash set than `O(n)` in a vector) with nothing that cares about ordering

fpga_interchange/arch_pack_clusters.cc
```diff
@@ -0,0 +1,496 @@
...
+
+    do {
+        //log_info("Cell name: %s - Cell type: %s - Bel: %s\n", next_cell->name.c_str(ctx), next_cell->type.c_str(ctx), ctx->nameOfBel(next_bel));
+        placement.push_back(std::make_pair(next_cell, next_bel));
```
gatecat, line 171: I think `emplace_back` should work and is cleaner than `push_back(std::make_pair...`

fpga_interchange/arch_pack_clusters.cc
```diff
@@ -0,0 +1,496 @@
...
+        next_net = next_cell->ports[port].net;
+
+        // Check if we have reached the end of the chain
+        if (next_net == nullptr || next_net->users.size() == 0) {
+            break;
+        }
+
+        next_cell = next_net->users.at(0).cell;
+
+        next_bel = find_next_chained_bel(ctx, next_bel, ctx->id("CO3"));
```
gatecat, line 250: this hardcoded `CO3` needs either to be fixed, or a `FIXME` comment left explaining why it's there

fpga_interchange/arch_pack_clusters.cc
```diff
@@ -0,0 +1,496 @@
...
+        }
+
+        next_net = next_cell->ports[port].net;
+
+        // Check if we have reached the end of the chain
+        if (next_net == nullptr || next_net->users.size() == 0) {
+            break;
+        }
+
+        next_cell = next_net->users.at(0).cell;
```
gatecat, line 248: what if `next_net` has more than one user?

fpga_interchange/arch_pack_clusters.cc
```diff
@@ -0,0 +1,496 @@
...
+
+    return true;
+}
+
+ArcBounds Arch::getClusterBounds(ClusterId cluster) const
+{
+    NPNR_ASSERT(cluster != ClusterId());
+    CellInfo *root_cell = getClusterRootCell(cluster);
+    CellInfo *last_cell = packed_clusters.at(cluster).back();
+    Loc loc_root = getBelLocation(root_cell->bel);
```
gatecat, line 265: these functions (this and `getClusterOffset`) are only intended to return approximations, but it's important that they function **before** the chain has been placed (that is in fact their main use, after the chain has been placed anything downstream that wants to know approximate bounds/offsets of the chain can just query actual positions). So this current implementation isn't really the intended purpose of these functions.

fpga_interchange/arch_pack_clusters.cc
```diff
@@ -0,0 +1,496 @@
...
+}
+
+void Arch::prepare_cluster(const BelChainPOD *chain)
+{
+    Context *ctx = getCtx();
+    IdString chain_name(chain->name);
+
+    // Get chainable cells
+    std::vector<CellInfo *> chainable_cells;
+    for (auto cell : sorted(ctx->cells)) {
```
gatecat, line 348: another note post #718, this will need to be changed to `for (auto &cell : ctx->cells) {` `CellInfo *ci = cell.second.get();`

fpga_interchange/arch_pack_clusters.cc
```diff
@@ -0,0 +1,496 @@
...
+        while (next_cell != nullptr) {
+            // Find possible source type/port to follow in cluster building
+            std::pair<IdString, IdString> config;
+            IdString src_cell_port(chain->source->cell_port);
+            IdString src_bel_port(chain->source->bel_port);
+            IdString sink_cell_port(chain->sink->cell_port);
+            IdString sink_bel_port(chain->sink->bel_port);
+
+            if (next_cell == root) {
+                disconnect_port(ctx, next_cell, sink_cell_port);
```
gatecat, line 387: can you give a bit more context behind this `disconnect_port` ?

acomodi: Sure, comment added

fpga_interchange/arch_pack_clusters.cc
acomodi: Fixed

fpga_interchange/arch_pack_clusters.cc
acomodi: I might have missed something, but I have noticed that the `getClusterBounds` is not called at the moment in any of the algos. At the moment I left it blank, with a  `TODO`, to fill it as soon as it will be necessary.

I have also implemented a "fixed" heuristic for the clusters that span multiple tiles, hacked for the moment to return a valid offset for the carry-chains. I think that this kind of information could be provided as a device configuration in the cluster info.

fpga_interchange/arch_pack_clusters.cc
acomodi: This has been reworked and the discovery of the chained cells is only done when preparing clusters and all the users are checked when doing so.

fpga_interchange/arch_pack_clusters.cc
acomodi: this was actually a leftover, this information now comes from the chip info.

fpga_interchange/arch_pack_clusters.cc
acomodi: Fixed

fpga_interchange/arch_pack_clusters.cc
acomodi: Sure, I guess I just came back from python when writing this :)

fpga_interchange/arch.h
acomodi: Changed all instances into `pool/dict`

acomodi:
@gatecat I have reworked quite a bit the clustering approach for the interchange. With the current implementation I can get to a working murax design placed and routed with carry chains in ~8-9 minutes, where most of the time is still spent during placement.

For what I understand, there is still quite some time spent by the placer to place FFs which are not legal due to different S/R/CE nets w.r.t. already placed FFs in the same site.
One possible solution I though of, that might still require an extension of the clustering approach, is to have a sort of attraction function that gathers and organizes sets of FFs (or other) to be placed at the same site, thus making the whole legalization step quite faster, as all the compatible cells can be placed at once in the same site. Can this be a valid solution?

gatecat:
>  One possible solution I though of, that might still require an extension of the clustering approach, is to have a sort of attraction function that gathers and organizes sets of FFs (or other) to be placed at the same site, thus making the whole legalization step quite faster, as all the compatible cells can be placed at once in the same site. Can this be a valid solution?

Yes and this seems reasonable in terms of the long-term direction for nextpnr, however it will need some discussion design wise to make sure it's flexible enough.

While I'm sure it will help a lot, I still think the biggest root-cause here is simply the validity check being 10-100x slower in interchange than other arches and speeding that up will be the easiest way to significantly improve placer runtime.

fpga_interchange/arch_pack_clusters.cc
gatecat: yeah, there are some possible future placer ideas I had that meant it could be useful, but a stub is alright for now `getClusterOffset` is the more important one for HeAP.

acomodi:
Ok, I'd say I may start looking into how to get some more performance out of the validity checks once we land the first clustering handling implementation.

Apart from the `getClusterOffset` implementation that is only effective for CARRY4, I believe this PR is now ready for another review.

fpga_interchange/arch_pack_clusters.cc
```diff
@@ -0,0 +1,585 @@
...
+            IdString cell_pin(pin_map.cell_pin);
+            IdString bel_pin(pin_map.bel_pin);
+
+            cell_bel_pins[cell_pin].push_back(bel_pin);
+        }
+
+        placement.emplace_back(cluster_node, next_bel);
+
+        for (auto port : cluster_node->ports) {
+            log_info("      - %s\n", port.first.c_str(ctx));
```
gatecat, line 242: this should be removed, or gated behind one of the debug flags and/or a macro

fpga_interchange/arch_pack_clusters.cc
```diff
@@ -0,0 +1,585 @@
...
+        offset.x = chainable_port.avg_x_offset * distance;
+        offset.y = chainable_port.avg_y_offset * distance;
+    }
+
+    return offset;
+}
+
+bool Arch::isClusterStrict(const CellInfo *cell) const { return true; }
+
+void dump_clusters(const ChipInfoPOD *chip_info, Context *ctx)
```
gatecat, line 341: I think this should be `static`

fpga_interchange/arch_pack_clusters.cc
```diff
@@ -0,0 +1,585 @@
...
+            ci->cluster.set(ctx, ci->name.c_str(ctx));
+            roots.push_back(ci);
+
+            // Chained cells use dedicated connections, usually not exposed to the
+            // general interconnect resources. The port disconnection is required for
+            // sink ports which are connected to GND or VCC by default, which are not
+            // reachable due to the fixed dedicated interconnect.
+            // E.g.: The CI input of carry chains in 7series corresponds to the CIN bel port,
+            //       which can only be connected to the COUT output of the tile below.
+            disconnect_port(ctx, ci, sink_port);
```
gatecat, line 429: it's notable that "CIN connected to 0" and "CIN connected to 1" are cases that need to be told apart, both being supported but with distinct FASM features: https://github.com/SymbiFlow/prjxray-db/blob/c28433840752ff0da5b8353bfdeafed24148aa38/artix7/segbits_clbll_l.db#L674-L675

I guess the  plan is to determine this from the logical netlist during FASM generation, as the connectivity in the physical netlist will be lost?

acomodi: I think this is an interesting case where the CIN and CYINIT BEL inputs are actually physically the same, but their connectivity is expressed differently for P&R purposes. 

If the `CYINIT` cell port is emitted by yosys, than there would be no problems and the port will not get disconnected, because the `PRECYINIT` routing BEL is connected to the general routing. The problem is that the `GND` net is assigned to the `CI` cell port as well, and this cell port is mapped to the `CIN` bel port, which does not have connections to `GND` (at least from a P&R perspective).

The root of a chain, all the `C0`, `C1`, `AX` connections can be present for `CYINIT`, and the information preserved in the physical netlist as well.

gatecat: Yeah, that seems reasonable. We might need something somewhere to move constants from `CIN` to `CYINIT` if this proves to be a problem.

fpga_interchange/arch_pack_clusters.cc
acomodi: This was actually some debug prints as CI is currently failing and I wanted to get some more data on why. Removed this log

fpga_interchange/arch_pack_clusters.cc
acomodi: Right, fixed

acomodi:
After some debugging, the CI failure is caused by [this issue](https://github.com/Xilinx/RapidWright/issues/183). The pin mappings for the CARRY4 are incomplete and should get added in the next RW release.

For the time being I disabled carry inference, and the only cluster that will be applied is going to be the `LUT+FF` one.

gatecat:
We could always modify CI to download that patch

acomodi:
> We could always modify CI to download that patch

Yes, this is actually better so we make sure nothing is broken with BELs chaining

gatecat (approved):
LGTM, I'll merge once the CI has gone green

PR #721, opened 2021-06-04
Updates for latest libmistral

gatecat:
See Ravenslofty/mistral#13

We also need to look at better versioning long term, probably managing the mistral clone in CMake, but that's for another PR.


gatecat:
note to self - DNM until the mistral patch is merged, and the temporary hack of pointing at my fork to pass CI, is changed back to upstream.

PR #739, opened 2021-06-30
interchange: Place entire IO macro based on routeability

gatecat:
This is needed to support UltraScale+ input buffers which are split between an `INBUF` and a `IBUFCTRL`, both in the same site with dedicated routing between the two that must be used.

Screenshot from 2021-06-30 11-49-17

### DIFF
--- a/fpga_interchange/arch.h
+++ b/fpga_interchange/arch.h
@@ -1152,6 +1152,7 @@ struct Arch : ArchAPI<ArchRanges>
     const DefaultCellConnsPOD *get_default_conns(IdString cell_type) const;
     void pack_default_conns();
 
+    dict<IdString, std::vector<CellInfo *>> macro_to_cells;
     void expand_macros();
 };
 
--- a/fpga_interchange/arch_pack_io.cc
+++ b/fpga_interchange/arch_pack_io.cc
@@ -22,7 +22,48 @@
 #include "nextpnr.h"
 #include "util.h"
 
+#include <queue>
+
 NEXTPNR_NAMESPACE_BEGIN
+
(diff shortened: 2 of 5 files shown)


gatecat:
cc @acomodi, this might be relevant to the IO stuff you've been looking at too.

acomodi:
@gatecat Indeed, I've hit a few issues that might get solved with this.

Specifically:

1. some macros were expanded in the same macro names indefinetely. E.g. OBUFTDS gets expanded in two OBUFTDS again and so on.
2. For the OBUFDS case, I hit the issue where an INV cell is necessary as well, but the INV cell can get placed in a wide number of BELs and the inverter BEL just before the OUTBUF is never used. As I see there is a change here to expand downstream and pre-place the additional BUFs, but I wonder whether we might do the same upstream.

gatecat:
(1) is probably a separate issue, although still something I can investigate if you want.

(2) would definitely be fixed by following downstream routing as well, and something I can do as a follow-up once this PR is in.

acomodi:
> (1) is probably a separate issue, although still something I can investigate if you want.

I have a local fix for this actually, which I need to refine a bit and I may open a PR soon

> (2) would definitely be fixed by following downstream routing as well, and something I can do as a follow-up once this PR is in.

Sounds great

gatecat:
> I have a local fix for this actually, which I need to refine a bit and I may open a PR soon

It'd be useful to have this, as it'll give me a test case to work on the second issue.

acomodi:
Sure, opened a PR here: https://github.com/YosysHQ/nextpnr/pull/749

PR #741, opened 2021-06-30
interchange: fix dedicated interconnect exploration

acomodi:

This relaxes the detailed routing check when verifying that there is a dedicated connection. Instead of returning `false` whenever the max depth is reached, the loop is continued instead to check other expansion branches.

For instance, the ISERDESes `D` input pin could never be found, as a set of routing nodes was traversed, making the dedicated inteconnect from IOB to D impossible, by prematurely returning `false`.

In addition, the max depth is calculated only for routing nodes.

### DIFF
--- a/fpga_interchange/dedicated_interconnect.cc
+++ b/fpga_interchange/dedicated_interconnect.cc
@@ -50,7 +50,7 @@ struct WireNode
 //
 // Routing networks with depth <= kMaxDepth is considers a dedicated
 // interconnect.
-constexpr int kMaxDepth = 20;
+constexpr int kMaxDepth = 6;
 
 void DedicatedInterconnect::init(const Context *ctx)
 {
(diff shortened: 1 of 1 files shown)


fpga_interchange/dedicated_interconnect.cc
```diff
@@ -50,7 +50,7 @@ struct WireNode
 //
 // Routing networks with depth <= kMaxDepth is considers a dedicated
 // interconnect.
-constexpr int kMaxDepth = 20;
+constexpr int kMaxDepth = 4;
```
gatecat, line 53: is there any reason why this is being reduced so dramatically? 20 does seem a big high, but 4 also seems low

acomodi: Basically, kMaxDepth was increased also for expansions within a site. It now keeps track only of expansions out of site

gatecat: would increasing it to around 6 still perform decently? that seems more reasonable to be

acomodi: I am not sure on how many routing nodes can be traversed for a dedicated interconnect, where a wire might be divided in multiple nodes. I might as well increase this a bit more though

acomodi: > would increasing it to around 6 still perform decently? that seems more reasonable to be

Sure, I'll test it right away

acomodi: It seems that there is a small but I think negligible hit on performance with 6

PR #742, opened 2021-07-01
interchange: phys: do not output nets which have no users

acomodi:

This solves https://github.com/Xilinx/RapidWright/issues/196.

Nets which do not have any users should not be emitted, or otherwise the DCP generated out of a physical netlist with such nets will not be readable, producing the errors described in https://github.com/Xilinx/RapidWright/issues/196

### DIFF
--- a/fpga_interchange/fpga_interchange.cpp
+++ b/fpga_interchange/fpga_interchange.cpp
@@ -535,10 +535,21 @@ void FpgaInterchange::write_physical_netlist(const Context * ctx, const std::str
         phys_cell.setPhysType(PhysicalNetlist::PhysNetlist::PhysCellType::PORT);
     }
 
-    auto nets = phys_netlist.initPhysNets(ctx->nets.size());
+    int nets_to_remove = 0;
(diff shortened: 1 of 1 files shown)


fpga_interchange/fpga_interchange.cpp
```diff
@@ -535,11 +535,22 @@ void FpgaInterchange::write_physical_netlist(const Context * ctx, const std::str
         phys_cell.setPhysType(PhysicalNetlist::PhysNetlist::PhysCellType::PORT);
     }
 
-    auto nets = phys_netlist.initPhysNets(ctx->nets.size());
+    int nets_to_remove = 0;
+    for(auto & net_pair : ctx->nets) {
+        auto &net = *net_pair.second;
+
+        if (net.users.size() == 0)
```
gatecat, line 542: nitpick `net.users.empty()` would be neater

acomodi: Right, fixed. Thanks

gatecat (approved):

PR #743, opened 2021-07-01
interchange: Reserve site ports only reachable from dedicated routing

### LINKED ISSUE #740: interchange: Pre-binding site routing is fundamentally broken in more complex cases
Right now, the site router is used to pre-bind the routing on the inputs to (and, to some extent, outputs from) sites, as part of the various attempts to prevent the router from routing through sites using site routing (which isn't legal in the fpga-interchange format).

However, this means the site router is essentially picking input/output pins at random, on the assumption that all site pins are equally routeable. But there is no guarantee that this is the case - consider how the UltraScale+ implements its IOB sites:

Screenshot from 2021-06-30 14-05-45

`OP` _must_ be used as the input site pin to the output buffer unless this is the negative side of a pseudo-differential IOB, in which case `OUTB_B_IN` _must_ be used.

By coincidence, for the simple wire case, the site router wrongly picks `OUTB_B_IN` and the design is hence unrouteable.

I can see three ways to fix this:
 - Modify the site router to check routeability and only pick site pins that appear to be routeable (this will have to be on a heuristic that won't pick up every case, but should pick up most). This will probably be done for binding site routing only, at least to start with, to avoid the performance cost in other cases.
 - Disable input pre-binding for certain sites/site pins in the YAML config only.
 - Remove the input pre-binding altgother, doing site routing in the general router, and investigate other ways to prevent illegal site route throughs.

gatecat:
Fixes #740, with this patch the "wire" test is now working on ZCU104 hardware

cc @acomodi

### DIFF
--- a/fpga_interchange/site_router.cc
+++ b/fpga_interchange/site_router.cc
@@ -27,6 +27,8 @@
 #include "site_arch.h"
 #include "site_arch.impl.h"
 
+#include <queue>
+
 NEXTPNR_NAMESPACE_BEGIN
 
 bool verbose_site_router(const Context *ctx) { return ctx->debug; }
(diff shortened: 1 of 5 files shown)


fpga_interchange/site_router.cc
```diff
@@ -1133,6 +1140,57 @@ static void block_cluster_wires(SiteArch *site_arch)
...
 }
 
+// Reserve site ports that are on dedicated rather than general interconnect
+static void reserve_site_ports(SiteArch *site_arch)
+{
+    const Context *ctx = site_arch->site_info->ctx;
+    site_arch->blocked_site_ports.clear();
+    for (PipId in_pip : site_arch->input_site_ports) {
+        pool<NetInfo*, hash_ptr_ops> dedicated_nets;
+        const int max_iters = 1000;
```
acomodi, line 1150: Might this be too high? To my understanding, the majority of the cases end up in reaching `max_iters`, unless there is a dedicated net and `max_iters` won't be reached.

Given that dedicated nets are quite short and do not require too many hops, would it make sense to reduce `max_iters`?

gatecat: sure, I'll reduce it it to 100, that'll be fine for now. if we hit problems with that we can look at better heuristics

PR #747, opened 2021-07-01
MachXO2 Checkpoint 1

cr1901:
I've made enough bug fixes that didn't get caught during the initial PR that it's worth making another PR to fix them. Fixes include initial VHDL support, generating global routing connections (global routing still unsupported), and fixing packing FFs that do not have LUTs attached to their inputs.

### DIFF
--- a/machxo2/examples/tinyfpga.vhd
+++ b/machxo2/examples/tinyfpga.vhd
@@ -1,0 +1,38 @@
+library ieee ;
+context ieee.ieee_std_context;
+
+use work.components.all;
+
+entity top is
+  port (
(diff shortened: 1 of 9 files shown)


machxo2/examples/tinyfpga.vhd
```diff
@@ -0,0 +1,38 @@
+library ieee ;
+context ieee.ieee_std_context;
```
gatecat, line 2: If you want to follow what prjtrellis did for ECP5, this kind of example is better off in prjtrellis/examples than in nextpnr

PR #749, opened 2021-07-05
interchange: tests: add obuftds test

acomodi:

This is to address comments in https://github.com/YosysHQ/nextpnr/pull/739#issuecomment-874024318.

This test shows an infinite macro expansion loop being generated for particular macros, and a possible fix is proposed.


fpga_interchange/arch_pack_io.cc
```diff
@@ -65,32 +65,24 @@ bool search_routing_for_placement(Arch *arch, WireId start_wire, CellInfo *cell,
...
                         pool<CellInfo *, hash_ptr_ops> *placed_cells)
 {
-    for (BelPin bel_pin : getWireBelPins(pad_wire)) {
-        BelId bel = bel_pin.bel;
-        for (CellInfo *cell : tightly_attached_bels) {
+    const auto &package_bel_data = bel_info(chip_info, package_bel);
+    auto bel_loc = getBelLocation(package_bel);
+    for (CellInfo *cell : tightly_attached_bels) {
+        for (BelId bel : getBelsByTile(bel_loc.x, bel_loc.y)) {
+            if (bel_info(chip_info, bel).site != package_bel_data.site)
```
gatecat, line 76: this approach means that only one top level pad per site is now permissible, which even though no arches currently hit, is not a constraint that I think we want to enforce?

I was looking at a solution that searches pips up/downhill from the pad wire, although didn't quite finish it today. what do you think about that approach instead?

acomodi: Yeah, the solution in this PR might also be rather fragile as well, and yeah I think that an exploration approach from the pad wire is more robust. I think that the working principles will be similar to the clustering code that searches for BELs compatible at the same site connected to a given port (pad in this case).

gatecat: This'll also be needed for the UltraScale+ differential inputs, where the differential input buffer is in another site to the IO pad altogether.

gatecat:
Closing as the test & fix got merged in #750.

PR #750, opened 2021-07-06
IO improvements for OBUFTDS

gatecat:
This supersedes #749 with, in theory, a more robust approach to finding pad-attached bels by looking at routing.

Unfortunately it's currently failing to route the negative side tristate signal, probably due to site pin issues similar to #743


gatecat:
The last commit seems to fix the tristate issue. cc @acomodi

acomodi:
> The last commit seems to fix the tristate issue

@gatecat Yes, thanks for that. This produces valid DCPs, working on HW as well.

PR #755, opened 2021-07-07
Pin modes parser

yrabbit:
A simple parser of pin mode flags. Only understands one flag per line.


pepijndevos:
(This is part of https://github.com/YosysHQ/apicula/pull/35 )

gowin/arch.cc
```diff
@@ -483,37 +484,44 @@ DelayQuad Arch::getWireTypeDelay(IdString wire)
...
+			const PairPOD *belname = pairLookup(package->pins.get(), package->num_pins, pinname.index);
+			if (belname == nullptr)
+				log_error("Pin %s not found\n", pinname.c_str(this));
+			std::string bel = IdString(belname->src_id).str(this);
+			it->second->attrs[IdString(ID_BEL)] = bel;
+		} else { // IO_PORT attr=value
+			std::string attr = "&";
+			attr += match[2];
+			boost::algorithm::to_upper(attr);
+			it->second->attrs[id(attr)] = 0;
```
gatecat, line 523: if you are setting a boolean attribute, setting it to 1 is more appropriate than 0

gowin/arch.cc
```diff
@@ -483,37 +484,44 @@ DelayQuad Arch::getWireTypeDelay(IdString wire)
...
-        //     std::cout << cell.first.str(this) << std::endl;
-        // }
         auto it = cells.find(net);
         if (it == cells.end()) {
             log_info("Cell %s not found\n", net.c_str(this));
             continue;
         }
-        std::string bel = IdString(belname->src_id).str(this);
-        it->second->attrs[IdString(ID_BEL)] = bel;
+		if (io_loc) { // IO_LOC name pin
```
gatecat, line 512: this could do with a `make clangformat` run to normalise the formatting

gowin/arch.cc
yrabbit: I'm only interested in having the attribute itself, so I'll correct it to 1.

gowin/arch.cc
yrabbit: I found clang-format10.

gowin/arch.cc
gatecat: seems like your last commit fixing the formatting accidentally set this back to 0

yrabbit: Damn. I need to be more careful.

gatecat (approved):
LGTM. @pepijndevos, would you like this to wait until the apicula side is merged?

pepijndevos:
I don't think it depends on Apicula, just that it will not work if you use these flags. So probably fine to merge.
I'll probably review the Apicula side this weekend.

PR #756, opened 2021-07-07
interchange: reduce run-time to check dedicated interconnect

acomodi:
This adds the possibility to cluster cells belonging to different sites (e.g. IDELAY -> IDELAYCTRL). To avoid spending useless time discovering valid BELs for only in-site clusters, I have added an additional piece of data to the [clustersPOD](https://github.com/SymbiFlow/python-fpga-interchange/pull/107).

The PR is marked as WIP as I have removed a detailed check for the dedicated interconnects which I think should be done, but might require a restructuring due to being quite expensive. Upon building the context, all the dedicated interconnects are pre-calculated. We can potentially enhance the pre-computed structures with additional data, avoiding us to perform a detailed check once again. I'll need to investigate a bit more what can be a better way of performing this check.

### DIFF
--- a/fpga_interchange/dedicated_interconnect.cc
+++ b/fpga_interchange/dedicated_interconnect.cc
@@ -39,6 +39,12 @@ enum WireNodeState
     IN_SOURCE_SITE = 2
 };
 
+enum ExpansionDirection
+{
+    EXPAND_DOWNHILL = 0,
+    EXPAND_UPHILL = 1
+};
+
 struct WireNode
 {
     WireId wire;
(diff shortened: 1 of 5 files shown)


fpga_interchange/site_router.cc
```diff
@@ -1315,6 +1315,7 @@ bool SiteRouter::checkSiteRouting(const Context *ctx, const TileStatus &tile_sta
 
     block_lut_outputs(&site_arch, blocked_wires);
     block_cluster_wires(&site_arch);
+    reserve_site_ports(&site_arch);
```
gatecat, line 1318: we had previously discussed not putting this in the `checkSiteRouting` path for performance reasons. we might need to benchmark and see if it slows things down significantly, if it does then some optimisation (caching etc) will be needed.

acomodi: Right, this was actually a leftover. The issue was that, given that the dedicated placement check was not running for clusters, the IDELAYs in the test-case were placed in invalid locations, therefore triggering an issue. The addition of this to the site routing check was a temporary check.

gatecat: Sure, thanks for explaining!

fpga_interchange/dedicated_interconnect.cc
```diff
@@ -52,6 +58,44 @@ struct WireNode
 // interconnect.
 constexpr int kMaxDepth = 6;
 
+static uint32_t get_num_pips(const Context *ctx, WireId wire, ExpansionDirection direction)
```
gatecat, line 61: I think this should also exclude pseudo-pips - there may be a few cases of dedicated interconnect where multiple pseudo-pips are downstream.

1 should be alright for basic Xilinx stuff but we might need to play with the pip count threshold to get good results on other arches.

PR #757, opened 2021-07-09
interchange: Add caching of site LUT mapping solution

mkurc-ant:
During placement site LUT mapping verification is an expensive step in terms of execution time. This PR adds caching of site LUT mapping solution to speed up the whole flow.

Cached data includes:
 - LUT cell pin to LUT BEL pin assignments (rotation)
 - LUT unused pin assignment
 - Solution validity flag

Cache key includes:
 - Tile type
 - Site in tile instance
 - LUT Cells bound to LUT BELs
 - Mapping of LUT cell ports to nets
   (note that those are not global nets but "normalized" / "localized" net ids just to know which ports are used and which ones are shared).


fpga_interchange/site_lut_mapping_cache.h
```diff
@@ -0,0 +1,110 @@
...
+NEXTPNR_NAMESPACE_BEGIN
+
+// Key structure used in site LUT mapping cache
+struct SiteLutMappingKey {
+
+    // LUT Cell data
+    struct Cell {
+        IdString                    type;       // Cell type
+        int32_t                     belIndex;   // Bound BEL index
+        dict<IdString, IdString>    conns;      // Port net assignments
```
gatecat, line 35: as the port names should always be known, I wonder if an array here with indexing corresponding to the list of inputs would make more sense? A `SSOArray` with a reserved size of 6 would avoid any heap allocations at all for almost all LUTs.

fpga_interchange/site_lut_mapping_cache.h
```diff
@@ -0,0 +1,110 @@
...
+               (cells == other.cells);
+    }
+
+    bool operator != (const SiteLutMappingKey &other) const {
+        return (tileType != other.tileType) ||
+               (site != other.site) ||
+               (cells != other.cells);
+    }
+
+    unsigned int hash () const {
```
gatecat, line 66: if the key is immutable, it might be more efficient to compute the hash once when creating the key, then store it and only have to return it when `hash()` is called

fpga_interchange/site_lut_mapping_cache.h
```diff
@@ -0,0 +1,110 @@
...
+};
+
+// Site LUT mapping result data
+struct SiteLutMappingResult {
+
+    // LUT cell data
+    struct Cell {
+    };
+
+    std::vector<std::pair<IdString, IdString>>  blockedWires;
```
gatecat, line 89: this might be another case where using a `SSOArray` makes sense to reduce the number of heap allocations

gatecat:
I like the direction this is going in and am interested to see what the results will be like! Left a few comments as to possible efficiency improvements.

mkurc-ant:
@gatecat Thanks for the insights! I'll be continuing work on that later this week.

fpga_interchange/site_lut_mapping_cache.h
```diff
@@ -0,0 +1,135 @@
...
+
+    // LUT Cell data
+    struct Cell {
+        IdString    type;       // Cell type
+        int32_t     belIndex;   // Bound BEL index
+  
+        // Port to net assignments. These are local net ids generated during
+        // key creation. This is to abstract connections from actual design
+        // net names. the Id 0 means unconnected.
+        int32_t     conns [MAX_LUT_INPUTS];
```
gatecat, line 44: I'd prefer `std::array` over C-style arrays

fpga_interchange/site_lut_mapping_cache.cc
```diff
@@ -0,0 +1,187 @@
...
+    dict<IdString, int32_t> netMap;
+    for (CellInfo* cellInfo : lutCells) {
+
+        NPNR_ASSERT(key.numCells < SiteLutMappingKey::MAX_LUT_CELLS);
+        auto& cell = key.cells[key.numCells++];
+
+        cell.type       = cellInfo->type;
+        cell.belIndex   = cellInfo->bel.index;
+
+        memset((void*)cell.conns, 0,
```
gatecat, line 78: once this becomes `std::array`, this could use `std::fill` instead

fpga_interchange/site_lut_mapping_cache.h
mkurc-ant: ok

fpga_interchange/site_lut_mapping_cache.cc
mkurc-ant: allright, I'll change those to `std:array`.

gatecat:
Mostly looking good now!

Two things to consider:

 - I think this could also do with a `make clangformat` to normalise the formatting? 
 - it's probably worth checking how big the cache gets with murax to see if we might need an eviction strategy above say 500MB-1GB

mkurc-ant:
Thanks. I can add reporting of the peak cache size as well as run code formatting.

There is a typo in one of the assertions ('=' instead of '==') so I expect that CI will fail for LIFCL chips (surprisingly it didn't for xc7)

mkurc-ant:
A quick test on the Murax design targeted to Basys 3 (xc7a35t) shows without the LUT caching:
```
Info: HeAP Placer Time: 73.95s
Info: SA placement time 178.65s
Info: Router2 time 37.53s
```

and with the LUT caching:
```
Info: HeAP Placer Time: 58.05s
Info: SA placement time 162.26s
Info: Router2 time 37.66s
```
Unfortunately the size of all the cached results is quite significant with the current implementation:
```
Info: Site LUT mapping cache stats:
Info:     miss ratio: 1.4%
Info:     peak size : 545MB (464029 items)
```

mithro:
I would say that we shouldn't worry about a <1Gb increase in memory usage -- especially if it is for an optional cache.

mkurc-ant:
I've formatted the code and added an option to disable the LUT cache (it is enabled by default). Is that how it should be?

gatecat (approved):
LGTM, thanks!

PR #761, opened 2021-07-12
interchange: add user placement constraints handling

acomodi:
This PR enables to LOC cells on BELs both from:
- attributes: `(* LOC="<site>", BEL="<bel>" *)`
- xdc: 
  - `set_property LOC <site> [get_cells <cell>]`
  - `set_property BEL <bel> [get_cells <cell>]`

Constraining a cell works only if both the LOC and BEL attributes are provided.

### DIFF
--- a/common/placer_heap.cc
+++ b/common/placer_heap.cc
@@ -402,13 +402,6 @@ class HeAPPlacer
         for (auto &cell_entry : ctx->cells) {
             CellInfo *cell = cell_entry.second.get();
 
-            if (cell->bel != BelId()) {
-                if (ctx->verbose)
-                    log_error("Cell \'%s\' (%s) has already been constrianed by arch implementation",
-                              cell->name.c_str(ctx), cell->type.c_str(ctx));
-                continue;
-            }
-
             auto loc = cell->attrs.find(ctx->id("BEL"));
             if (loc != cell->attrs.end()) {
                 std::string loc_name = loc->second.as_string();
--- a/common/placer_heap.cc
+++ b/common/placer_heap.cc
@@ -401,6 +401,14 @@ class HeAPPlacer
         // Initial constraints placer
         for (auto &cell_entry : ctx->cells) {
             CellInfo *cell = cell_entry.second.get();
+
+            if (cell->bel != BelId()) {
+                if (ctx->verbose)
+                    log_error("Cell \'%s\' (%s) has already been constrianed by arch implementation",
+                              cell->name.c_str(ctx), cell->type.c_str(ctx));
+                continue;
+            }
+
             auto loc = cell->attrs.find(ctx->id("BEL"));
             if (loc != cell->attrs.end()) {
                 std::string loc_name = loc->second.as_string();
--- a/fpga_interchange/arch_place_constr.cc
+++ b/fpga_interchange/arch_place_constr.cc
@@ -1,0 +1,102 @@
+/*
+ *  nextpnr -- Next Generation Place and Route
+ *
+ *  Copyright (C) 2021  Symbiflow Authors
+ *
+ *
+ *  Permission to use, copy, modify, and/or distribute this software for any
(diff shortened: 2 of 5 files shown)


common/placer_heap.cc
```diff
@@ -401,6 +401,14 @@ class HeAPPlacer
         // Initial constraints placer
         for (auto &cell_entry : ctx->cells) {
             CellInfo *cell = cell_entry.second.get();
+
+            if (cell->bel != BelId()) {
+                if (ctx->verbose)
+                    log_error("Cell \'%s\' (%s) has already been constrianed by arch implementation",
```
gatecat, line 407: an error should not be gated by `ctx->verbose` (also, s/constrianed/constrained)

fpga_interchange/arch_place_constr.cc
```diff
@@ -0,0 +1,102 @@
...
+                      "since it is already bound to cell \'%s\'\n",
+                      cell->name.c_str(getCtx()), nameOfBel(bel), bound_cell->name.c_str(getCtx()));
+
+        bindBel(bel, cell, STRENGTH_USER);
+        if (!isBelLocationValid(bel))
+            log_error("Bel \'%s\' is not valid for cell "
+                      "\'%s\' of type \'%s\'\n",
+                      nameOfBel(bel), cell->name.c_str(getCtx()), cell->type.c_str(getCtx()));
+
+        constrained_cells.emplace_back(cell->name, bel);
```
gatecat, line 91: at this point you should also remove the `BEL` attribute as nextpnr uses it internally for something a bit different (a single attribute containing both site and bel) - I think this is what you have been hitting in the placer but it's best to deal with it just by removing the attribute during packing after it's been used.

fpga_interchange/xdc.cc
```diff
@@ -106,6 +119,45 @@ static int get_ports(ClientData data, Tcl_Interp *interp, int objc, Tcl_Obj *CON
...
+    if (objc == 1) {
+        // Return list of all ports.
+        Tcl_SetStringResult(interp, "Unimplemented");
+        return TCL_ERROR;
+    } else if (objc == 2) {
+        const char *arg0 = Tcl_GetString(objv[1]);
+        IdString cell_name = ctx->id(arg0);
+
+        CellInfo *cell = nullptr;
+        for (auto &cell_pair : ctx->cells) {
```
gatecat, line 134: I think you could use `ctx->cells.find` rather than iterating over all the cells?

fpga_interchange/arch_place_constr.cc
acomodi: Indeed, thanks for the hint. I have erased the `BEL` attr as suggested, which is way better than dealing with that in the heap placer.

fpga_interchange/xdc.cc
acomodi: Sure, fixed

fpga_interchange/arch_place_constr.cc
```diff
@@ -0,0 +1,103 @@
...
+                      cell->type.c_str(getCtx()));
+
+        auto bound_cell = getBoundBelCell(bel);
+        if (bound_cell)
+            log_error("Cell \'%s\' cannot be bound to bel \'%s\' "
+                      "since it is already bound to cell \'%s\'\n",
+                      cell->name.c_str(getCtx()), nameOfBel(bel), bound_cell->name.c_str(getCtx()));
+
+        bindBel(bel, cell, STRENGTH_USER);
+        if (!isBelLocationValid(bel))
```
gatecat, line 86: I'd suggest checking the location of all constrained bels at the end (like in the IO placer), in the interchange arch sometimes there are cases where multiple bels must be placed (due to dedicated interconnect, site routeability etc), before they become valid.

acomodi: Right, makes sense, fixed

gatecat (approved):

PR #765, opened 2021-07-17
gowin: remove unused find_package(PythonInterp). NFC

whitequark:


pepijndevos:
#766 also updates the readme but is otherwise exactly the same

PR #789, opened 2021-08-02
ecp5: Copy REGMODE in PDP mode to both A and B ports

gatecat:
Signed-off-by: gatecat <gatecat@ds0.me>


kittennbfive:
Thank you!

PR #790, opened 2021-08-05
interchange: place only cells belonging to the same clusters in the same site

acomodi:
This PR is to improve placement run-time, sacrificing placement density.

It appears in fact that quite some time is still spent when finding routing solutions within a site, and, with the presence of clusters (and specifically carry-chains), congestion within the sites has increased.

Clusters have a flag to indicate whether they are to be the only ones occupying a site.

Note: the site routing cache clear step is not required for this patch to work, but the problem I saw was that, in some runs when trying to place and route a LiteX design, the site router is not able to find a solution after clearing the cache. This seems quite odd, but I still need to investigate why the failure is happening.

### DIFF
--- a/.github/workflows/interchange_ci.yml
+++ b/.github/workflows/interchange_ci.yml
@@ -114,7 +114,7 @@ jobs:
       env:
         RAPIDWRIGHT_PATH: ${{ github.workspace }}/RapidWright
         PYTHON_INTERCHANGE_PATH: ${{ github.workspace }}/python-fpga-interchange
-        PYTHON_INTERCHANGE_TAG: v0.0.17
+        PYTHON_INTERCHANGE_TAG: v0.0.18
         PRJOXIDE_REVISION: 1bf30dee9c023c4c66cfc44fd0bc28addd229c89
         DEVICE: ${{ matrix.device }}
       run: |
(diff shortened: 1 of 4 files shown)


fpga_interchange/arch.cc
```diff
@@ -841,10 +841,6 @@ static void prepare_sites_for_routing(Context *ctx)
         }
     }
 
-    // Clear the site routing cache. This is because routing at this stage is done with the extra constraint of blocked
-    // pins to ensure a routeable pin choice.
-    ctx->site_routing_cache.clear();
```
gatecat, line 846: I'm concerned this is going to break UltraScale+ IO (which is where this was added, due to the blocked pins situation), ideally this should be investigated rather than worked around.

acomodi: Indeed, I'll try to get a smaller and reproducible case of the failure when the cache clearing is enabled.

Thinking out loud, it might as well be a bug in the cache, such that some placements are allowed based on a cache that does not exactly represent the real site status.

acomodi: @gatecat After some debugging iterations, I have discovered that there is a possible issue with the routing cache which still needs to be further investigated.

Basically, it seems that during the site routing binding, there could be some cache entries belonging to other sites that produce the effect of having fewer solutions when trying to solve routing for these other sites, making it impossible, in some cases, to find a feasible site routing.

I need to experiment a bit more on the choice of the routing cache key, which may solve the problem, but for now, a short term solution is to disable the usage of the site routing cache when binding the various site routings prior to global routing.

gatecat (approved):
LGTM, disabling the cache for strict binding of routes seems fair

PR #798, opened 2021-08-19
Make EXTREFB handling more robust

kleinai:
Avoids a segfault if an EXTREFB does not connect directly to its associated DCUA.
Also adds location constraints specifically for EXTREFB.

### DIFF
--- a/ecp5/pack.cc
+++ b/ecp5/pack.cc
@@ -1766,6 +1766,26 @@ class Ecp5Packer
             if (ci->type == id_EXTREFB) {
                 const NetInfo *refo = net_or_nullptr(ci, id_REFCLKO);
                 CellInfo *dcu = nullptr;
+                std::string loc_bel = std::string("NONE");
+                std::string dcu_bel = std::string("NONE");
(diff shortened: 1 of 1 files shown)


kleinai:
The change will need future updates to support routing through the refclk to fpga core path. I have a theory as to how this might work but can't seem to find the wires to use.

gatecat:
Looks good, I've merged this part and happy to look into the routing if you can point me to the suspected missing wires?

kleinai:
That would be a huge help. I'm way out of my depth with the trellis database stuff.

The wires might already exist I'm just not seeing them. I think what I'm looking for is `JSERDESREFCLK` on `BMID`. I _think_ that's the input to the DCCs from the DCUs. What I don't see is a wire or arc from the DCUs named anything remotely close to refclk or extref.

My plan for packing was to move any non-direct non-DCU users of the EXTREF's `refclko` pin over to a new (or merged) net from the DCU's extref-to-core pin and silently enable that output. I believe this is how the vendor tool handles it as it's valid to route `refclko` to say a general PLL.

PR #810, opened 2021-09-10
Timing report in JSON format

mkurc-ant:
In this PR I'm adding writing the final timing analysis report in the machine-readable JSON format. The report is attached to the current report JSON.

The included changes are:

- [x] decouple critical path report generation from its printout
- [ ] ~add an commandline option to specify report file name~
- [x] write critical paths to the JSON report
- [x] write all sequential net delays to the JSON report
- [x] commandline option to control detailed per-net timing writeout
- [x] decouple fmax met/not met calculation from its printout
- [ ] ~write fmax and met/not met information to the JSON report~ Already there
- [x] add some documentation for the JSON structure

### DIFF
--- a/common/command.cc
+++ b/common/command.cc
@@ -183,6 +183,9 @@ po::options_description CommandHandler::getGeneralOptions()
     general.add_options()("placed-svg", po::value<std::string>(), "write render of placement to SVG file");
     general.add_options()("routed-svg", po::value<std::string>(), "write render of routing to SVG file");
 
+    general.add_options()("timing-report", po::value<std::string>(),
+                          "write timing analysis report in CSV format to file");
+
     return general;
 }
 
@@ -247,6 +250,11 @@ void CommandHandler::setupContext(Context *ctx)
 
     if (vm.count("timing-allow-fail")) {
         ctx->settings[ctx->id("timing/allowFail")] = true;
+    }
+
+    if (vm.count("timing-report")) {
+        std::string filename = vm["timing-report"].as<std::string>();
+        ctx->settings[ctx->id("timing/reportFile")] = filename;
     }
 
     if (vm.count("placer")) {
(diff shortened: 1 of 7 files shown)


common/command.cc
```diff
@@ -181,6 +181,9 @@ po::options_description CommandHandler::getGeneralOptions()
     general.add_options()("placed-svg", po::value<std::string>(), "write render of placement to SVG file");
     general.add_options()("routed-svg", po::value<std::string>(), "write render of routing to SVG file");
 
+    general.add_options()("timing-report", po::value<std::string>(),
+                          "write timing analysis report in CSV format to file");
```
gatecat, line 185: looks like this is doing JSON, not CSV

mkurc-ant: Right, I forgot to change that in the help string.

mkurc-ant:
@gatecat Yes, I saw what happens in `report.cc`. I've been thinking how to include timing data there. I've noticed that `timing_analysis` is called from `router1`. To do reporting from `report.cc` I would need to store the timing report within the context. I can re-think my approach but that would certainly require more code refactoring in nextpnr.

gatecat:
If you are up for it, storing the timing report in the context as per the current `report.cc` approach would be preferable. This would also enable the Python API to discover the critical path, for example.

mkurc-ant:
@gatecat Alright, I'll be working on refactoring that so it works as the current `report.cc`. But in the meantime could we merge this implementation as it is? We would like to start using these reports.

gatecat:
I'd really prefer to have some stable to review. Is there any particular part of this that's most important for your workflow?

mkurc-ant:
@gatecat So the code is pretty stable at this point. It only requires storing the report within the context. As for the workflow: we need a machine-readable timing report for a CI dashboard. It would be great if we could land this PR and then polish the code in subsequent ones. Could you agree on that ?

mkurc-ant:
@gatecat I finally found some time to rework this PR. The timing reports are now appended to the existing report JSON.

gatecat:
Thanks, almost there :)

Please can you run a `make clangformat` to get this ready to be merged?

mkurc-ant:
Alright, done.

gatecat (approved):
LGTM

mkurc-ant:
Thanks! :)

PR #812, opened 2021-09-14
Convert macros to clusters for better placement 

mtdudek:
This code adds cluster detection and placement for clusters created from macros.
So far only in-site macros are being placed, as out-site macros may could use global routing resources leading to visiting whole FPGA fabric.
This issue can be rectified be checking routing outside site if it's dedicated routing, like in BRAM64, carry chains or DIFF IO.
and only using them, and fail on routing that uses global routing.

Done so far:
- [x] Cluster from macro detection
    - [x] Cluster cells to idx mapping
    - [x] Idx solver 
- [x]  Cluster placement 
    - [x] root_bel check
    - [x] compatible bell checker
    - [x] logical netlist checker

### DIFF
--- a/fpga_interchange/arch_pack_clusters.cc
+++ b/fpga_interchange/arch_pack_clusters.cc
@@ -49,6 +49,7 @@ struct ClusterWireNode
     WireId wire;
     ClusterWireNodeState state;
     int depth;
+    bool only_down;
 };
 
 static void handle_expansion_node(const Context *ctx, WireId prev_wire, PipId pip, ClusterWireNode curr_node,
@@ -188,18 +189,10 @@ CellInfo *Arch::getClusterRootCell(ClusterId cluster) const
     return clusters.at(cluster).root;
 }
 
-bool Arch::getClusterPlacement(ClusterId cluster, BelId root_bel,
-                               std::vector<std::pair<CellInfo *, BelId>> &placement) const
+bool Arch::normal_cluster_placement(
+    const Context *ctx, const Cluster &packed_cluster, const ClusterPOD &cluster_data,
+    CellInfo *root_cell, BelId root_bel, std::vector<std::pair<CellInfo *, BelId>> &placement) const
 {
-    const Context *ctx = getCtx();
-    const Cluster &packed_cluster = clusters.at(cluster);
-
-    auto &cluster_data = cluster_info(chip_info, packed_cluster.index);
-
-    CellInfo *root_cell = getClusterRootCell(cluster);
-    if (!ctx->isValidBelForCellType(root_cell->type, root_bel))
-        return false;
-
     BelId next_bel;
 
     // Place cluster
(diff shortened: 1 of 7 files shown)


fpga_interchange/arch_pack_clusters.cc
```diff
@@ -282,6 +276,312 @@ bool Arch::getClusterPlacement(ClusterId cluster, BelId root_bel,
...
+            }
+
+            if (bel_data.synthetic)
+                continue;
+
+            if (!ctx->isValidBelForCellType(cell->type, bel))
+                continue;
+
+            log_info("\t\t\tfound: %s\n", bel_pin.pin.c_str(ctx));
+            bels[std::pair<BelId, BelId>(root_pin.bel, bel_pin.bel)].insert(
```
gatecat, line 311: `emplace` is neater than `insert(std::pair...`

mtdudek:
Doing this implementation I found issue when mapping from logical netlist to physical bels and wires, as some connections that are in logical level cannot be directly translated to physical ones. Also resolving idx(cell)-bel mapping every time cluster is placed will use loads of computation time. So I'll move idx<->bel mapping to device DB as it will not increase size significantly and it will reduce time by large amount

mtdudek:
Waiting for python-fpga-interchange to get merge, then I'll update dependencies and this should get green

mtdudek:
This PR need #825 to land first

fpga_interchange/arch.cc
```diff
@@ -1767,6 +1767,9 @@ bool Arch::checkPipAvailForNet(PipId pip, NetInfo *net) const
         NPNR_ASSERT(src != wire);
         NPNR_ASSERT(dst != wire);
 
+        if (tile_type.lut_elements.size() > 0)
```
gatecat, line 1770: looks like this is disabling all pseudo-pips in tiles with LUTs, are you sure you want to do this?

mtdudek: I;m testing it right now, if it's really required. I had issues before, but maybe they are fixed now.

mtdudek: So with out this lines Nextpnr doesn't work for xilinx 7-series, as it uses pseudo-PIPs going through (A|B|C|D)6LUTs, but it doesn't set A6 pin high making this pseudo-PIP broken

mtdudek: So I've disabled offending pseudoPIPs in python-fpga-interchange. This code is no longer used

tmichalak (changes requested):

fpga_interchange/macros.cc
```diff
@@ -78,6 +79,7 @@ void Arch::expand_macros()
 
             // Get the ultimate root of this macro expansion
             IdString parent = (cell->macro_parent == IdString()) ? cell->name : cell->macro_parent;
+            log_info("%s %s\n", cell->name.c_str(ctx), parent.c_str(ctx));
```
gatecat, line 82: this looks like debugging that should either be removed or have some descriptive text added and a `ctx->debug` gate

fpga_interchange/arch_pack_clusters.cc
```diff
@@ -370,6 +506,344 @@ static bool check_cluster_cells_compatibility(CellInfo *old_cell, CellInfo *new_
...
+                log_info("Possible idx %s:\n", arc.first->name.c_str(ctx));
+                for (const auto idx : arc.second)
+                    log_info("    - %d\n", idx);
+            }
+        }
+        // Solve for binary constraints
+        std::queue<std::pair<uint32_t, uint32_t>> workqueue;
+        for (const auto &arc : idx_to_cells)
+            for (const auto &connection : cluster->connection_graph[arc.first].connections)
+                workqueue.push(std::pair<uint32_t, uint32_t>(arc.first, connection.target_idx));
```
gatecat, line 717: I think this can be `emplace` instead of `push`

fpga_interchange/arch_pack_clusters.cc
```diff
@@ -370,6 +506,344 @@ static bool check_cluster_cells_compatibility(CellInfo *old_cell, CellInfo *new_
...
+            cells_in_macro[IdString(cell_type.name)] = cell_type.count;
+            cell_types.insert(IdString(cell_type.name));
+        }
+
+        for (auto &node_cell : macro_to_cells[ci->macro_parent]){
+            auto cell_type = node_cell->type;
+            if(!counter.count(cell_type))
+                counter[cell_type] = 0;
+            counter[cell_type]++;
+            cell_types.insert(cell_type);
```
gatecat, line 646: is the separate `cell_types` pool needed, or could you just use they keys of `counter`? some comments describing what exactly these are storing would be useful

fpga_interchange/arch_pack_clusters.cc
```diff
@@ -370,6 +506,344 @@ static bool check_cluster_cells_compatibility(CellInfo *old_cell, CellInfo *new_
...
+            if(ctx->verbose && cells_in_macro.count(cell_type))
+                log_info("Have: %s %d\n", cell_type.c_str(ctx), counter[cell_type]);
+            if(!cells_in_macro.count(cell_type) || !counter.count(cell_type) || cells_in_macro[cell_type] != counter[cell_type])
+                failed = true;
+            if(failed && ctx->verbose)
+                log_info("Cell count stage failed, for sure not this cluster\n");
+            if(failed)
+                break;
+        }
+        if(failed){
```
gatecat, line 661: I think this needs a `make clangformat` run on it

fpga_interchange/arch_pack_clusters.cc
```diff
@@ -282,6 +276,148 @@ bool Arch::getClusterPlacement(ClusterId cluster, BelId root_bel,
...
+    if(ctx->debug)
+        log_info("Used bels:\n");
+    for(const auto &bel : cluster.physical_placements[idx].places[placement_idx].bels){
+        IdString s_bel(bel);
+        BelId t = check_and_return(root_bel.tile, s_bel);
+        IdStringList cpy(root_bel_full_name.size());
+        if (t == BelId()){
+            for(uint32_t j = 0; j < root_bel_full_name.size(); j++)
+                cpy.ids[j] = root_bel_full_name[j];
+            cpy.ids[1] = s_bel;
```
gatecat, line 358: index `size()-1` is probably better than index 1 if a name had more than 2 components ever

fpga_interchange/arch_pack_clusters.cc
```diff
@@ -370,6 +506,344 @@ static bool check_cluster_cells_compatibility(CellInfo *old_cell, CellInfo *new_
     return true;
 }
 
+bool reduce(uint32_t x, uint32_t y, const ClusterPOD *cluster, dict<uint32_t, pool<CellInfo *, hash_ptr_ops>> &domain, Context *ctx){
```
gatecat, line 509: this could do with a more descriptive name, or a comment describing what it's doing

fpga_interchange/arch_pack_clusters.cc
```diff
@@ -370,6 +506,344 @@ static bool check_cluster_cells_compatibility(CellInfo *old_cell, CellInfo *new_
...
+                log_info("Possible idx %s:\n", arc.first->name.c_str(ctx));
+                for (const auto idx : arc.second)
+                    log_info("    - %d\n", idx);
+            }
+        }
+        Cluster cluster_info;
+        cluster_info.root = ci;
+        cluster_info.index = index;
+        cluster_info.cluster_nodes.resize(idx_to_cells.size());
+        ci->cluster.set(ctx, ci->name.str(ctx));
```
gatecat, line 835: `ci->cluster = ci->name` should be fine here?

fpga_interchange/arch_pack_clusters.cc
mtdudek: In my opinion it's needed to store cell_types that were seen in macro_to_cells and cell_types that are used to create cluster.

fpga_interchange/arch_pack_clusters.cc
mtdudek: Fixed in next commit

fpga_interchange/arch_pack_clusters.cc
mtdudek: Added comment in new commit

fpga_interchange/arch_pack_clusters.cc
mtdudek: Fixed in new commit

fpga_interchange/macros.cc
mtdudek: Removed

fpga_interchange/arch_pack_clusters.cc
mtdudek: Changed to emplace

fpga_interchange/arch_pack_clusters.cc
```diff
@@ -370,6 +506,344 @@ static bool check_cluster_cells_compatibility(CellInfo *old_cell, CellInfo *new_
...
+        dict<IdString, uint32_t> cells_in_macro, counter;
+        pool<IdString> cell_types;
+        for (auto &cell_type : cluster->required_cells){
+            cells_in_macro[IdString(cell_type.name)] = cell_type.count;
+            cell_types.insert(IdString(cell_type.name));
+        }
+
+        for (auto &node_cell : macro_to_cells[ci->macro_parent]){
+            auto cell_type = node_cell->type;
+            if(!counter.count(cell_type))
```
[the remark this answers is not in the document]
mtdudek: Removed if statement, you are right

fpga_interchange/arch_pack_clusters.cc
mtdudek: Run make clangformat on project

gatecat (approved):
LGTM, thanks for your quick response to my review comments!

PR #815, opened 2021-09-17
nexus: Fixed an improved SIOLOGIC handling

mkurc-ant:
This PR includes two major changes required for `IDDRX1` and `ODDRX1` plus tri-state control to work together on `LIFCL-17`:
 - Fixed the case when `SIOLOGIC` is used ant `SEIOxx_CORE.T` connection is bypassing it
 - Added inference and integration of flip-flops which drive `SEIOxx_CORE.T` into `SIOLOGIC` whenever possible

The first issue was causing a routing conflict as there are no separate routes for `SEIOxx_CORE.T` and `SIOLOGIC.TSDATA0` outside the tile. This is solved by detecting that `SEIOxx_CORE.T` bypasses `SIOLOGIC` and connecting `SIOLOGIC.TSDATA0`  to the same net. This does not affect the design operation as the mux on `SEIOxx_CORE.T` still choses the correct input making `SIOLOGIC.TOUT` ignored.

The second thing is an improvement which allows nextpnr to take advantage of "TREG" located aside the ODDR part of `SIOLOGIC`. Whenever it is detected that a flip-flop drives `SEIOxx_CORE.T` bypassing `SIOLOGIC` appropriate transformation is applied which removes the filp-flop and enables the one in `SIOLOGIC`.

There is currently a workaround in place for unknown bits emission which seem to control the `SEIOxx_CORE.T` input mux.

### DIFF
--- a/nexus/pack.cc
+++ b/nexus/pack.cc
@@ -1197,10 +1197,10 @@ struct NexusPacker
             NetInfo *dout = get_net_or_empty(ci, id_DOUT);
             if (dout != nullptr && dout->users.size() == 1)
                 iob = dout->users.at(0).cell;
-            NetInfo *tout = get_net_or_empty(ci, id_DOUT);
+            NetInfo *tout = get_net_or_empty(ci, id_TOUT);
             if (tout != nullptr && tout->users.size() == 1)
                 iob = tout->users.at(0).cell;
-            if (iob->type != id_SEIO18_CORE && iob->type != id_SEIO33_CORE && iob->type != id_DIFFIO18_CORE)
+            if (iob == nullptr || (iob->type != id_SEIO18_CORE && iob->type != id_SEIO33_CORE && iob->type != id_DIFFIO18_CORE))
                 log_error("Failed to find associated IOB for IOLOGIC %s\n", ctx->nameOf(ci));
             io_to_iol[iob->name].push_back(ci);
         }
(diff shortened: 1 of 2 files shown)


nexus/pack.cc
```diff
@@ -2075,6 +2075,239 @@ struct NexusPacker
...
+            return nullptr;
+        }
+
+        // Get the flip-flop that drives it
+        CellInfo* ff = net->driver.cell;
+        if (ff->type != id_OXIDE_FF) {
+            return nullptr;
+        }
+
+        // Get clock nets of IOLOGIC and the flip-flop
```
gatecat, line 2095: I think this is going to need checks for LSR compatability, and I'm also not sure about how CE works here. Safest is to gate merging them on the `syn_useioff` attribute like ECP5 so it only happens when you explicitly want it and not otherwise.

nexus/pack.cc
```diff
@@ -2075,6 +2075,239 @@ struct NexusPacker
...
+                // Propagate parameters
+                iol->params[id_SRMODE] = ff->params.at(id_SRMODE);
+                iol->params[id_REGSET] = ff->params.at(id_REGSET);
+
+                iol->params[ctx->id("TSREG.REGSET")] = std::string("SET");
+
+                // CEOUTMUX.1
+                iol->params[ctx->id("CEOUTMUX")] = std::string("1");
+
+                // FIXME: Workaround for an unknown bit
```
gatecat, line 2284: I'll try and have a poke at the fuzzer later today.

nexus/pack.cc
```diff
@@ -2075,6 +2075,239 @@ struct NexusPacker
...
+            }
+            NPNR_ASSERT(iob != nullptr);
+
+            // SIOLOGIC handling
+            if (iol->type == id_SIOLOGIC) {
+
+                // We have IDDR+ODDR
+                if (isODDR && isIDDR) {
+                    if (!iob->attrs.count(ctx->id("GLITCHFILTER"))) {
+                        iob->attrs[ctx->id("GLITCHFILTER")] = std::string("ON");
```
gatecat, line 2174: to double check, are you sure this defaults to `ON` in the vendor tools?

nexus/pack.cc
gatecat: unknown bit should be added in https://github.com/gatecat/prjoxide/commit/7d56d6392d859e18155e43818a62ea3f5789946a

nexus/pack.cc
mkurc-ant: Alright, I can add an attribute controlling that. I couldn't find any mention of such an attribute used by the Vendor tools for crosslink-nx. 

And as I wrote in the description: with this PR It merges only a flip-flop driving T and only if there is an active OLIGIC there. I've noticed the behavior only when using Synplify as the synthesis engine. It didn't happen with LSE.

nexus/pack.cc
mkurc-ant: Great, thanks. This unknown bits I use seem to control whether T is driven by `IOLOGIC` or externally.

nexus/pack.cc
mkurc-ant: So I did a minitest considering various `IDDRX1` and `ODDRX1` configurations:
 - regular bidirectional IO
 - IDDRX1 only
 - ODDRX1 only
 - ODDRX1 plus tri-state control (external to the ODDR)
 - IDDRX1 + ODDRX1 plus external tri-state control
 
And I noticed the glitch filter enabled only for the bidirectional DDR case by default. Those attributes assigned around this place in the code are the result of my findings.

nexus/pack.cc
mkurc-ant: I'll update nextpnr to use correct FASM features now.

nexus/pack.cc
gatecat: Great, would you be willing to submit a PR with these tests into https://github.com/gatecat/prjoxide/tree/master/minitests for future reference?

mkurc-ant: Sure. I just need to clean them a little bit.

nexus/pack.cc
mkurc-ant: @gatecat I added the requirement for `syn_useioff` to be set to `1` on a net driving `T` for the flip-flop integration to be considered. I also added a check whether the flip flop can be removed.

PR #816, opened 2021-09-17
[EARLY WIP] Proof of concept Xilinx arch with interchange backend

gatecat:
This is currently an attempt at a much cleaner rewrite of the old [nextpnr-xilinx](https://github.com/gatecat/nextpnr-xilinx). It's not yet remotely usable for any kind of actual design; but the fundamental database and validity checking rules are in place and it's another example of the new direction for nextpnr arches.

At the moment it requires RapidWright to build databases; and outputs logical and physical netlists that RapidWright can turn into DCPs for Vivado. However, it's not designed to be tied to these long term and may well even change name before merge to reflect this, as the hope is that it can be used with and open source database, and moreover bitstream generation.

Currently it has only been tested on a 'blinky' type design and much work remains to implement relative constraint rules for carries; distributed RAM; complex IO; etc as well as add timing data. There is also work to be done on tuning delay estimates, placement rules, etc, for performance

However, the plan is to merge it upstream sooner rather than later (i.e. in the next few weeks) to avoid the bitrot that afflicted nextpnr-xilinx, even if it's less feature-complete than that or even the current FPGA interchange arch (which despite the benefits of being generic offers neither the performance nor 'creature comforts' of a true nextpnr arch).

It's very important to note that at this stage this is not so much for end user purposes (not least due to the Vivado requirement), but as a platform for experimenting with place and route algorithms on large very FPGAs and provide a starting point for integration with future open source Xilinx work - even if some database changes are needed, I would be happy to accommodate these and add bitstream generation outputs in whatever format such a project required.

The current deduplication approach has database sizes of 112MB for the xcvu13p (3.8M LEs) and 131MB for the xcvc1902 (1.9M LEs). It's unclear how much lower one could go without significantly impacting on access complexity. For perspective the iCE40HX8K's non-deduplicated database is 89MB in size.


gatecat:
This is now planned to become a himbaechel uarch instead.

PR #825, opened 2021-09-23
Fix chain swap

mtdudek:
Issue was due to dest_bels being not cleared between clusters unbindes, causing
newly bind bels to be unbinded and having their old bel value changed to new bel value.
Then when swap failed 2 cells were being bind to a single bel.

I tested leaving dest_bels in the function scope and moving it to the loop scope.
Code with dest_bels in the loop scope was faster than leaving it in the function scope,
and checking if the cell is in the processed cluster.

Signed-off-by: Maciej Dudek <mdudek@antmicro.com>

### DIFF
--- a/common/placer1.cc
+++ b/common/placer1.cc
@@ -619,7 +619,6 @@ class SAPlacer
     {
         std::vector<std::pair<CellInfo *, Loc>> cell_rel;
         dict<IdString, BelId> moved_cells;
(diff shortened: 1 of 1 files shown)


mtdudek:
@gatecat could you take a look?

gatecat (approved):
LGTM

PR #836, opened 2021-10-03
mistral: Add bel pins for MLAB write port

gatecat:
This is the first step towards MLAB-as-memory support.

Next steps will be:
 - [ ] adding a packer to combine 1-bit mlab cells with the same read/write ports into larger clusters for easier placement
 - [ ] adding a cell-bel pin map rule for mlab cells, which appears to follow, for a single ALM:
```
logical     top half   bottom half
RADDR[0]    A          A
RADDR[1]    B          B
RADDR[2]    C          C
RADDR[3]    D          D
RADDR[4]    F0         F0
WDATA       E0         E1
RDATA       COMBOUT    COMBOUT
```
 - [ ] working out the specifics of clock/enable routing and legal mlab+ff pairings and updating the validity checker accordingly
 - [ ] adding bitstream generation support, hardcoding the pulse width value as Quartus doesn't seem to vary them at least in a basic test design


Ravenslofty:
`Acked-by: Lofty <dan.ravensloft@gmail.com>`

PR #843, opened 2021-10-09
mistral: very basic timing info

Ravenslofty:
Featuring hardcoded timing information, yay :P

### DIFF
--- a/mistral/arch.cc
+++ b/mistral/arch.cc
@@ -400,15 +400,6 @@ void Arch::assignArchInfo()
     }
 }
 
-delay_t Arch::estimateDelay(WireId src, WireId dst) const
-{
-    int x0 = CycloneV::rn2x(src.node);
-    int y0 = CycloneV::rn2y(src.node);
-    int x1 = CycloneV::rn2x(dst.node);
-    int y1 = CycloneV::rn2y(dst.node);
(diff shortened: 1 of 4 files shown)


mistral/arch.cc
```diff
@@ -423,6 +423,99 @@ ArcBounds Arch::getRouteBoundingBox(WireId src, WireId dst) const
     return bounds;
 }
 
+TimingPortClass Arch::getPortTimingClass(const CellInfo *cell, IdString port, int &clockInfoCount) const
+{
+    clockInfoCount = 0;
+    if (cell->type.in(id_MISTRAL_NOT, id_MISTRAL_BUF, id_MISTRAL_ALUT2, id_MISTRAL_ALUT3, id_MISTRAL_ALUT4, id_MISTRAL_ALUT5, id_MISTRAL_ALUT6)) {
```
gatecat, line 429: worth dealing with `MISTRAL_ALUT_ARITH` too ?

Ravenslofty: Yeah, I will. But since most of the cells are in one of these categories I figured it'd be a good start. LUTRAM timing also needs to be in there, I think.

gatecat: Yeah, makes sense

PR #858, opened 2021-11-05
MachXO2 Checkpoint 2

cr1901:
This PR corrects the following bugs I found in the MachXO2 backend:

* Bitstream gen has more accurate logic to work around MachXO2's "wonderful" net naming scheme.
* `nextpnr-machxo2` will now error if the detected top-level I/O ports are not attached to a `FACADE_IO` `PAD` port.
* `--no-iobs` option was removed. It was always enabled, and it was a mistake to expose it in the first place.
* `mitertest.sh` can now handle JSON containing `FACADE_IO`. The script uses a flatten pass to instantiate tristate buffer modules and remove their `inout`s, which `write_smt2` doesn't support. 
  * `uart.v` example now passes BMC w/ an SMT solver. I'm holding off on making it pass k-induction for now because I ran into a [snag](https://github.com/cr1901/yosys-experiments/tree/master/sat/dffhold) I'm not sure how to handle rn. Additionally, `uart.v` still doesn't pass w/ `yosys's` built-in SAT solver, but this is more than likely because I don't understand how the built-in SAT solver works :).
* Add MachXO2 to the list of supported archs :).

### DIFF
--- a/machxo2/bitstream.cc
+++ b/machxo2/bitstream.cc
@@ -62,6 +62,7 @@ static std::string get_trellis_wirename(Context *ctx, Location loc, WireId wire)
     // relative coords push them outside the bounds of the chip.
     // Indents are based on wires proximity/purpose.
     auto is_pio_wire = [](std::string name) {
+        // clang-format off
         return (name.find("DI") != std::string::npos || name.find("JDI") != std::string::npos ||
                     name.find("PADD") != std::string::npos || name.find("INDD") != std::string::npos ||
                     name.find("IOLDO") != std::string::npos || name.find("IOLTO") != std::string::npos ||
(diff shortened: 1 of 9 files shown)


machxo2/bitstream.cc
```diff
@@ -59,13 +60,20 @@ static std::string get_trellis_wirename(Context *ctx, Location loc, WireId wire)
 
     // Handle MachXO2's wonderful naming quirks for wires in left/right tiles, whose
     // relative coords push them outside the bounds of the chip.
+    // Indents are based on wires proximity/purpose.
```
gatecat, line 63: I think clangformat is going to destroy this? (there is  hot comment you can use to disable it though iirc)

cr1901: Fixed with `// clang-format [on, off]`.

PR #859, opened 2021-11-06
gowin: Add partnumbers and packages to the chipdb

yrabbit:
Instead of parsing the partnumber with a regular expression, a simple table is used. This is done because the structure of the partnumber changes as new features appear (e.g., ES instead of C6/I5)

### DIFF
--- a/.cirrus/Dockerfile.ubuntu20.04
+++ b/.cirrus/Dockerfile.ubuntu20.04
@@ -65,4 +65,4 @@ RUN set -e -x ;\
     PATH=$PATH:$HOME/.cargo/bin cargo install --path prjoxide
 
 RUN set -e -x ;\
-    pip3 install apycula==0.0.1a9
+    pip3 install apycula==0.0.1a12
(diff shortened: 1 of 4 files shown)


pepijndevos:
In order for CI to pass you'll need to update to the latest Apycula release
https://github.com/YosysHQ/nextpnr/blob/1615b0a50044b2cef2fa5d572c7010040c98fe19/.cirrus/Dockerfile.ubuntu20.04#L68

And since we bumped the chipdb version number, we also need to update... or add a check, so that Nextpnr gives an error if it's used with the wrong version database.

https://github.com/YosysHQ/nextpnr/blob/1615b0a50044b2cef2fa5d572c7010040c98fe19/gowin/arch.h#L148

pepijndevos:
Yea probably needs a version check here
https://github.com/YosysHQ/nextpnr/blob/1615b0a50044b2cef2fa5d572c7010040c98fe19/gowin/arch.cc#L677

PR #862, opened 2021-11-19
common: Improved the random seed initialisation for the context

DX-MON:
The method for randomisation initialisation used in nextpnr at present has been deprecated because of several extreme issues with it for some time now (for example, some implementations of rand() have near unchanging low bits, resulting in very predicable values).

This patch replaces that bad C code with modern C++11 `<random>` header based code.

### DIFF
--- a/common/command.cc
+++ b/common/command.cc
@@ -33,6 +33,7 @@
 #include <boost/program_options.hpp>
 #include <fstream>
 #include <iostream>
+#include <random>
 #include "command.h"
 #include "design_utils.h"
 #include "json_frontend.h"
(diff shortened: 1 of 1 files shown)


gatecat:
Unfortunately this appears to be failing on CI. While it would be nice to improve this a bit, given its not a particularly well used feature it's also not worth breaking compatability here.

DX-MON:
We managed to forget the `#include` for the types used. `rand()` and `srand()` produce warnings on MSVC, and are notably marked as being non thread-safe.

We decided to patch this in response to a user hitting issues with invoking nextpnr too quickly using this feature over on the 1bitsquared Discord where because the PRNG was being seeded based on `time()` it was not providing any inter-run randomness for a full second which was screwing up their searching for usable random seeds in parallel

gatecat:
Thanks!

PR #863, opened 2021-11-22
nexus: LUT and FF clustering

mkurc-ant:
This PR adds to the nexus architecture inference of directly connected LUT+FF pairs followed by clustering them together.

For each directly connected LUT and FF nextpnr now creates a new cluster. Moreover it glues FFs to existing clusters comprised of LUTs like wide LUTs and carry-chains.

The new behavior is enabled by default but can be disabled by the `--no-pack-lutff` option. Another option `--carry-lutff-ratio` allows to specify a ratio of clustering together FFs and LUTs which are part of carry-chains.

The added clustering approach has been analyzed on a single design (the SoC from https://chromium.googlesource.com/chromiumos/platform/hps-firmware) targeting LIFCL-17 FPGA using 500 chosen seeds. A timeout of 5min was set for each run so that when it expired a run was considered a failure. The conclusions are the following:

- median runtime was reduced to ~78% of the original,
- 99% of the runs succeeded instead of 88.2% for without clustering,
- 16.0% runs met timing instead of 0.6% for without clustering,
- median router iteration count dropped to 67 from 103.

The following two charts show router2 progression in terms of overused wire count vs. iteration count. Each trace is terminated either with a green circle (indicating a success) or a red "x" (indicating a failure). Plots are clipped to 200 iterations for better clarity.

No clustering:
no-clustering

With clustering:
with-clustering

I'm also attaching CSV files with the gathered statistical results:
[no-clustering.csv](https://github.com/YosysHQ/nextpnr/files/7582536/no-clustering.csv)
[with-clustering.csv](https://github.com/YosysHQ/nextpnr/files/7582537/with-clustering.csv)

As it can be seen in the pictures with the LUT and FF clustering enabled the router is able to converge much quicker and gets stuck more rarely.

The option to specify clustering ratio for FFs and existing carry-chain clusters is meant to mitigate situations when placement of such large clusters of fully populated slices cannot be legalized.

### DIFF
--- a/nexus/pack.cc
+++ b/nexus/pack.cc
@@ -2395,7 +2395,11 @@ struct NexusPacker
         pack_luts();
         pack_ip();
         handle_iologic();
-        pack_lutffs();
+
+        if (!bool_or_default(ctx->settings, ctx->id("no_pack_lutff"))) {
+            pack_lutffs();
+        }
+
         promote_globals();
         place_globals();
         generate_constraints();
--- a/nexus/pack.cc
+++ b/nexus/pack.cc
@@ -2304,6 +2304,78 @@ struct NexusPacker
         }
     }
 
+    void pack_lutffs () {
+        log_info("Inferring LUT+FF pairs...\n");
+
+        size_t num_comb = 0;
+        size_t num_ff   = 0;
+        size_t num_pair = 0;
+
+        for (auto &cell : ctx->cells) {
+            CellInfo *ff = cell.second.get();
+            if (ff->type != id_OXIDE_FF) {
+                continue;
+            }
+
+            num_ff++;
+
+            // Get input net
+            NetInfo *di = get_net_or_empty(ff, id_M); // At the packing stage all inputs go to M
+            if (di == nullptr || di->driver.cell == nullptr) {
+                continue;
+            }
+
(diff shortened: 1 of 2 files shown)


nexus/pack.cc
```diff
@@ -2304,6 +2304,128 @@ struct NexusPacker
...
+                // Get the cluster root
+                CellInfo* root = ctx->cells.at(lut->cluster).get();
+
+                // Constrain the FF relative to the LUT
+                ff->cluster = root->cluster;
+                ff->constr_x = lut->constr_x;
+                ff->constr_y = lut->constr_y;
+                ff->constr_z = lut->constr_z + 2;
+                ff->constr_abs_z = lut->constr_abs_z;
+                root->constr_children.push_back(ff);
```
gatecat, line 2402: this case also needs to be checking that the FF can legally be added to the cluster (i.e. CLK/LSR/CE sharing etc)

mkurc-ant: Oh, you are right. I missed that... I'll push a fix soon.

mkurc-ant: Alright, should be fixed by now.

gatecat (approved):

gatecat:
Thanks!

PR #868, opened 2021-12-12
ecp5: Reduce some chipdb fields from 32 to 16 bit

mkj:
This reduces the final binary size by ~9 MB for 85k

### DIFF
--- a/ecp5/arch.h
+++ b/ecp5/arch.h
@@ -54,11 +54,10 @@ NPNR_PACKED_STRUCT(struct BelPortPOD {
 
 NPNR_PACKED_STRUCT(struct PipInfoPOD {
     LocationPOD rel_src_loc, rel_dst_loc;
-    int32_t src_idx, dst_idx;
-    int32_t timing_class;
-    int16_t tile_type;
+    int16_t src_idx, dst_idx;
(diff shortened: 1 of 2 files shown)


gatecat:
Some of these look alright, others are causing things to no longer be suitably aligned hence the failure in CI.

mkj:
Not sure why building locally works OK but CI fails, I've reordered some fields now.

ecp5/arch.h
```diff
@@ -54,24 +54,24 @@ NPNR_PACKED_STRUCT(struct BelPortPOD {
...
+    int16_t timing_class;
     int16_t tile_type;
     int8_t pip_type;
     int8_t padding_0;
 });
 
 NPNR_PACKED_STRUCT(struct PipLocatorPOD {
     LocationPOD rel_loc;
-    int32_t index;
+    int16_t index;
```
gatecat, line 66: I'd rather keep `PipLocatorPOD` as being 32-bit aligned

mkj:
The structs all keep 32 bit alignment now. Turns out I was missing `-DNDEBUG` in my local test build here.

gatecat (approved):
LGTM now, thanks!

PR #873, opened 2021-12-14
ice40: Pack LUT at start of carry chain if there is 1 candidate

### LINKED ISSUE #864: Packer fails to pack SB_CARRY and SB_LUT4 for the LSB of adders.
See reproducer in :  [carry.zip](https://github.com/YosysHQ/nextpnr/files/7637687/carry.zip)

Simple 8 bit adder : 

```verilog
module test(
    input wire [7:0] a,
    input wire [7:0] b,
    output reg [7:0] x,
    input wire       clk
);

    always @(posedge clk)
        x <= a + b;

endmodule
```

And the result :

For bit 0 the `SB_CARRY` and the `SB_LUT4` were not packed in the same LCs
despite having compatible inputs. So you see above it used 9 LCs instead of 8.
And the bottom one is the `SB_CARRY` alone and the one on the right is the `SB_LUT4` alone for `x[0]`.

gatecat:
Fixes #864

### DIFF
--- a/ice40/pack.cc
+++ b/ice40/pack.cc
@@ -203,6 +203,8 @@ static void pack_carries(Context *ctx)
             CellInfo *carry_lc = nullptr;
             if (carry_ci_lc && carry_lcs.find(carry_ci_lc->name) != carry_lcs.end()) {
                 carry_lc = carry_ci_lc;
+            } else if (ci_const && carry_lcs.size() == 1) {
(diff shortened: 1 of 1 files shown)


smunaut:
Argh, I realized I forgot to come back to you about this, but it indeed seems to work just fine both in the simple reproducer I attached to the issue and in several real designs.

PR #867, opened 2021-12-12
Improvements to ecp5 router speed

mkj:
With these two changes I see a roughly 3x speedup in router1 time, ~550 seconds to 180 secs (on a i5-1145G7 laptop). The vector change is the main improvement.

I've been testing with a build of [microwatt for orangecrab](https://github.com/mkj/microwatt).
`nextpnr-ecp5 --json microwatt-placed.json --no-pack --no-place --lpf constraints/orange-crab-0.2.lpf --textcfg microwatt_out.config.tmp --85k --freq 48 --speed 8 --write microwatt-routed.json --timing-allow-fail --ignore-loops --package CSFBGA285 `
Input files
 [microwatt-placed-bench.json](https://matt.ucc.asn.au/p/microwatt-placed-bench.json) 
[orange-crab-0.2.lpf](https://matt.ucc.asn.au/p/orange-crab-0.2.lpf)
([microwatt-bench.json](https://matt.ucc.asn.au/p/microwatt-bench.json) as input for pack and place)

PR #879, opened 2021-12-17
nexus: router1 speedup based on #867

gatecat:
This uses a set of data structures based on #867 to speed up the nexus router1 runtime

cc @tcal-x; as you are using router1 for congested designs the runtime speedup from this might be of interest to you


tcal-x:
Thanks @gatecat 
fyi @alanvgreen @kgugala

PR #884, opened 2021-12-20
gowin: Add simplified IO cells processing

yrabbit:
Some models have I/O cells that are IOBUFs, and other types (IBUFs and
OBUFs) are obtained by feeding 1 or 0 to the OEN input.  This is done
with general-purpose routing so it's best to do it here to avoid
conflicts.

For this purpose, in the new bases, these special cells are of type IOBS
(IOB Simplified).

The proposed changes are compatible with bases of previous versions of
Apycula and do not require changing .CST constraint files.


pepijndevos:
For the record, this is part of https://github.com/YosysHQ/apicula/pull/70

It appears we're currently silently including routing bits into the IOB modes, which is not great.

Looks good but I'll test the combined set of changes on hardware later.

pepijndevos (approved):
Confirmed working on hardwar

PR #886, opened 2021-12-21
mistral: M10K support

Ravenslofty:
Presently we have bels added, at least.


mistral/m10k.cc
```diff
@@ -66,4 +66,75 @@ void Arch::create_m10k(int x, int y)
     }
 }
 
+static void assign_lab_pins(Context *ctx, CellInfo *cell)
```
gatecat, line 69: should be `assign_m10k_pins`

gatecat (approved):
LGTM

gatecat:
🎊

PR #895, opened 2022-01-18
Sync with the current state of mistral

galibert:
Mistral now handles the defaults fully, so that can be removed from nextpnr.  Also a lot of bugs were fixed so a resync is a good idea.


mistral/bitstream.cc
```diff
@@ -29,108 +29,13 @@ struct MistralBitgen
...
-            case CycloneV::HPS_TEST: {
-                if (pt == CycloneV::CFG_DFX_BYPASS_ENABLE)
-                    cv->inv_set(pn2r.second, true);
-                break;
-            };
-            case CycloneV::GPIO: {
-                // Ignore GPIO used by the design
-                BelId bel = ctx->bel_by_block_idx(CycloneV::pn2x(pn), CycloneV::pn2y(pn), id_MISTRAL_IO,
-                                                  CycloneV::pn2bi(pn));
-                if (bel != BelId() && ctx->getBoundBelCell(bel) != nullptr)
```
gatecat, line 112: I need to check this PR on hardware, but note that this only sets the IO default inverters when they are unused. Do we need to add code to clear the default inverters when they are used instead?

galibert: Possibly so yes.  At this point mistral (I think) correctly sets the state for an empry fpga for the given package when calling clear().  So you only need to delta from that.

mistral/arch.h
```diff
@@ -461,9 +461,41 @@ struct Arch : BaseArch<ArchRanges>
...
+    }
+
     WireId get_port(CycloneV::block_type_t bt, int x, int y, int bi, CycloneV::port_type_t port, int pi = -1) const
     {
-        return WireId(cyclonev->pnode_to_rnode(CycloneV::pnode(bt, x, y, port, bi, pi)));
+        auto rn = find_rnode(bt, x, y, port, bi, pi);
+        if(rn)
+            return WireId(rn);
+
+        fprintf(stderr, "Trying to connect unknown node %s\n", CycloneV::pn2s(CycloneV::pnode(bt, x, y, port, bi, pi)).c_str());
```
gatecat, line 492: use `log_error` rather than `fprintf` and `exit`

gatecat: this might need to move into `arch.cc` to avoid some horrid circular header dependencies accessing the log functions, too

galibert: Is log_error aborting?

gatecat: Yeah, it is

PR #897, opened 2022-01-21
nexus: arch: add option to adjust the estimation delay multiplication factor

acomodi:
(edited)

This PR adds the a runtime parameter to control the multiplication factor of the estimate delay for nexus devices, without changing the default behaviour.

@gatecat cc

### DIFF
--- a/nexus/arch.cc
+++ b/nexus/arch.cc
@@ -601,7 +601,8 @@ delay_t Arch::estimateDelay(WireId src, WireId dst) const
     int dst_x = dst.tile % chip_info->width, dst_y = dst.tile / chip_info->width;
     int dist_x = std::abs(src_x - dst_x);
     int dist_y = std::abs(src_y - dst_y);
-    return 75 * dist_x + 75 * dist_y + 250;
+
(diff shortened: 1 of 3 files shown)


nexus/arch.cc
```diff
@@ -601,7 +601,12 @@ delay_t Arch::estimateDelay(WireId src, WireId dst) const
     int dst_x = dst.tile % chip_info->width, dst_y = dst.tile / chip_info->width;
     int dist_x = std::abs(src_x - dst_x);
     int dist_y = std::abs(src_y - dst_y);
-    return 75 * dist_x + 75 * dist_y + 250;
+
+    int mult = 75;
+    if (getCtx()->settings.count(getCtx()->id("estimate-delay-mult")))
```
gatecat, line 606: a call to ctx->id here is going to be adding an unnecessary string hash and compare to a hot path, can you make mult a field of arch that is loaded from settings before place/route instead?

acomodi: Right, thanks. I have added `mult` as a member of arch as suggested and the settings reading step just before placement.

gatecat:
Thanks!

PR #900, opened 2022-01-28
Honor nexus OSCA frequency tolerance

mkurc-ant:
This PR adds honoring of the +/- 15% tolerance of `OSCA` output frequency for auto-generated clock constraints.

### DIFF
--- a/nexus/pack.cc
+++ b/nexus/pack.cc
@@ -1987,8 +1987,9 @@ struct NexusPacker
                     copy_constraint(ci, id_CLKI, id_CLKO, 1);
                 } else if (ci->type == id_OSC_CORE) {
                     int div = int_or_default(ci->params, ctx->id("HF_CLK_DIV"), 128);
-                    set_period(ci, id_HFCLKOUT, delay_t((1.0e6 / 450) * (div + 1)));
-                    set_period(ci, id_LFCLKOUT, delay_t((1.0e3 / 10)));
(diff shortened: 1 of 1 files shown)


danc86:
Are you sure this change is the right away around?

It looks like `set_period()` takes a delay (measured in time, not frequency) so I think this change should have made the period 15% *shorter* to account for the worst case. Instead it made the period 15% longer which seems to be the wrong direction.

I noticed because in our Litex-based design, we have HFOSC divisor 9 (50MHz) and an explicit period constraint of 17.391ns (57.5MHz) allowing for 15% variation. In the nextpnr output, after this change, I see the following:

```
13:16:10 Info: constraining clock net 'clkout' to 57.50 MHz
[...]
13:16:11 Info: Generating derived timing constraints...
13:16:11 Warning:     Overriding derived constraint of 57.5 MHz on net clkout with user-specified constraint of 43.5 MHz.
13:16:11 Info:     Derived frequency constraint of 57.5 MHz for net clkout$glb_clk
```

The warning message is a bit confusingly worded. I think it's saying the "user-specified" constraint of 43.5MHz is the one added by this PR. But I expected it should be 57.5MHz (i.e. assume worst case faster oscillator) not 43.5MHz.

mkurc-ant:
@danc86 You're right... I have created another PR which shortens the period instead of lengthening it https://github.com/YosysHQ/nextpnr/pull/902

PR #901, opened 2022-01-29
gowin: Add GUI.

yrabbit:
* Items such as LUT, DFF, MUX, ALU, IOB are displayed;
* Local wires, 1-2-4-8 wires are displayed;
* The clock spines, taps and branches are displayed with some caveats.

For now, you can not create a project in the GUI because of possible
conflict with another PR (about GW1NR-9C support), but you can specify
the board in the command line and load .JSON and .CST in the GUI.

Although ALUs are displayed, but the CIN and COUT wires are not. This is
still an unsolved problem.


gowin/gfx.h
```diff
@@ -0,0 +1,4935 @@
...
+
+#ifndef GFX_H
+#define GFX_H
+
+#include "nextpnr.h"
+
+NEXTPNR_NAMESPACE_BEGIN
+
+// LUTs
+const float lut_w = 0.6732 - 0.6386;
```
gatecat: Ideally these constants should be in the `.cc` file (or, if not possible at all, marked `static const`). Otherwise if the header is included in two different `.cc` files then I suspect there will be multiple definition failures at link time.

gowin/gfx.h
```diff
@@ -0,0 +1,4935 @@
...
+const float ios_x = io_x;
+const float ios_y = ios_scl * io_y;
+
+const dict<IdString, float> portPoint = {
+        {id_O, 3. * io_h / 4.},
+        {id_I, 2. * io_h / 4.},
+        {id_OEN, 1. * io_h / 4.},
+};
+
+const dict<IdString, std::vector<std::tuple<float, float, float, float>>> portSign = {
```
gatecat: out of curiosity, were these autogenerated by something? would it be possible to include the generator script for reference, too?

yrabbit: Unfortunately, there is no generator, I write this way by hand:(

gowin/gfx.h
yrabbit: Done.

PR #904, opened 2022-02-03
gowin: Add a DS location recognition

yrabbit:
For differential signals it is necessary to set the position of two pins
at once: P and N.
This commit adds that capability and also adds another style of location
setting --- with the pin letter in square brackets used in vendor tools.


gowin/arch.cc
```diff
@@ -551,17 +554,20 @@ void Arch::read_cst(std::istream &in)
...
+                    // may be it's IOx#[AB] style?
+                    Loc loc = getLoc(match_pinloc, getGridDimX(), getGridDimY());
+                    BelId bel = getBelByLocation(loc);
+                    if (bel == BelId()) {
+                        log_error("Pin %s not found (TRBL style). \n", pinline.c_str());
+                    }
+                    std::string belname = getCtx()->nameOfBel(bel);
+                    it->second->setAttr(IdString(ID_BEL), belname);
+                } else {
+                    std::cout << "plain error:[" << pinline << "]" << std::endl;
```
gatecat, line 568: this looks like debugging that can be removed?

yrabbit: Oops, how did I miss that?

yrabbit:
What does error 503 mean? Has the tar run out of space on the HDD?
room

gatecat:
Probably spurious, certainly not because of this PR.

PR #922, opened 2022-02-23
gowin: Add support for true differential output

yrabbit:
The new primitive appears as an amalgamation of two existing OBUF
primitives.  Compatible with older versions of apicula, although, of
course, using TLVDS_OBUF with old databases will not bring the desired
result, but no crash.


gatecat:
CI failures interchange related and not a problem in this PR afaics, merged.

PR #936, opened 2022-03-08
Add missing part of PR #933

whitequark:
Not sure how I missed this before, sorry.


gatecat:
Do you think there'd be value in checking that it builds for wasm as part of the nextpnr CI?

whitequark:
That would be nice!

gatecat:
I need to look into tidying up the current split between GH actions and Cirrus, and I'll see if I can add that at the same time.

whitequark:
Thank you~

PR #939, opened 2022-03-11
[fpga_interchange] FIX: Don't hold reference to current visit in route_global_arc

kboronski-ant:
This reference can become invalid after modifying the dict at line 109, because the dict implementation may reallocate it. Copy by value instead to ensure the validity of visit data.


gatecat:
Thanks!

PR #940, opened 2022-03-11
ecp5: accept lowercase characters in hex strings

VioletEternity:
Currently, if you specify a `DCUA` parameter as a hex string, e.g. `.CH0_UDF_COMMA_B("0x17c")`, any lowercase characters in it will be silently treated the same as `F`.


gatecat:
Thanks!

PR #943, opened 2022-03-14
gowin: support for locales other than en_US and C

yrabbit:
Specifically, those locales where the fractional part separator in
floating point numbers is not a period.

### DIFF
--- a/common/command.cc
+++ b/common/command.cc
@@ -45,7 +45,15 @@
 
 NEXTPNR_NAMESPACE_BEGIN
 
-CommandHandler::CommandHandler(int argc, char **argv) : argc(argc), argv(argv) { log_streams.clear(); }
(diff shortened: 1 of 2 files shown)


gatecat:
Looking at CI this is failing with `what():  locale::facet::_S_create_c_locale name not valid`. If you can fix that, this should then be applied to all arches - probably in `CommandHandler`.

yrabbit:
Apparently this happens because the ubuntu test image does not have locale support installed. I think I can detect such a case and do something reasonable, but I would have to use this CI a couple of times to debug it (it would be faster than if I downloaded and installed an ubuntu distribution somewhere). 
Is this acceptable?

gatecat:
Yeah, that sounds reasonable to me

yrabbit:
Great. I think it works.
I don't know about the other tests, but ubuntu no longer swears at the locale:)

PR #944, opened 2022-03-14
ice40: fix crash when packing LUTs with no output

Riesi:
nextpnr-ice40 was crashing when packing forcefully placed LUTs with no output

The same error is probably in the generic packer at https://github.com/YosysHQ/nextpnr/blob/master/generic/pack.cc#L50
But I didn't have the setup to test it.

MWE:
run the following command to checkout the bug: 
`nextpnr-ice40 --up5k --package sg48 -v --json test.json.txt --pcf icoboard.pcf.txt`
[test.json.txt](https://github.com/YosysHQ/nextpnr/files/8244420/test.json.txt)
[icoboard.pcf.txt](https://github.com/YosysHQ/nextpnr/files/8244426/icoboard.pcf.txt)


gatecat:
Thanks!

PR #948, opened 2022-03-15
nexus: DCS support

mkurc-ant:
This PR adds support for the `DCS` primitive to the `LIFCL` architecture. Requires `prjoxide-db` with timing data for `DCS` which can be obtained with https://github.com/gatecat/prjoxide/pull/20.

The PR also fixes incorrect frequency constraint imposed on the LF output of OSCA.

`DCS` operation has been tested in hardware (LIFCL-40) and it worked correctly (switching clock between OSCA LF and HF output).


nexus/pack.cc
```diff
@@ -1981,19 +1984,37 @@ struct NexusPacker
...
+                }
             }
             changed_nets.clear();
             for (auto cell : changed_cells) {
                 CellInfo *ci = ctx->cells.at(cell).get();
                 if (ci->type == id_DCC) {
                     copy_constraint(ci, id_CLKI, id_CLKO, 1);
+                } else if (ci->type == id_DCS) {
+                    // For DCC copy the worst case ("fastest") constraint
+                    delay_t period_clk0, period_clk1;
```
gatecat, line 2003: it's a false positive I think but the compiler gives a maybe uninitialized warning for these that is failing CI, adding a default initialiser to zero or something is probably the easiest solution...

mkurc-ant: Alright, fixed.

PR #952, opened 2022-03-16
Nexus: Fixed OSCA parameters, add pll default parameters

mtdudek:
This PR adds default parameters to PLL cell in nexus.
Without this change bitstream generated using prjoxide does not work.

Signed-off-by: Maciej Dudek <mdudek@antmicro.com>

### DIFF
--- a/nexus/arch.h
+++ b/nexus/arch.h
@@ -1408,7 +1408,8 @@ struct Arch : BaseArch<ArchRanges>
     // -------------------------------------------------
 
     // Parse a possibly-Lattice-style (C literal in Verilog string) style parameter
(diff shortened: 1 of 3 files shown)


gatecat:
I must admit I'd rather not go this route given the possibilities of subtle changes in behaviour, so you'll have to define any defaults you care about in `bitstream.cc`.

mtdudek:
@gatecat I hope this is better

gatecat:
Sorry, had a bit of a backlog to deal with, I'll try and get a review of this & the prjoxide stuff done over the next few days...

mtdudek:
@gatecat  I've removed temporary cell. i hope now it's good to go.

PR #966, opened 2022-03-28
ice40: Merge driving LUT<=2s into carry-only LCs

gatecat:
This is an optimisation that relies on that fact that carry-only LCs have a LUT, 2 LUT inputs and the LUT output spare for a LUT2 to be packed into them (looking at LUT2s driving them is a simple heuristic to find these ahead of placement).

Unfortunately this sometimes creates what appear to be combinational loops, if you don't realise the LUT output only depends on the two inputs that aren't "don't care" in its function; so icetime will need to be fixed to handle this (although the need for icetime with nextpnr's timing analysis being more capable in general is marginal at best...). Handling of this is already implemented for nextpnr's own timing data.

### DIFF
--- a/ice40/arch.cc
+++ b/ice40/arch.cc
@@ -923,9 +923,20 @@ std::vector<GraphicElement> Arch::getDecalGraphics(DecalId decal) const
 
 bool Arch::getCellDelay(const CellInfo *cell, IdString fromPort, IdString toPort, DelayQuad &delay) const
 {
-    if (cell->type == id_ICESTORM_LC && cell->lcInfo.dffEnable) {
-        if (toPort == id_O)
-            return false;
+    if (cell->type == id_ICESTORM_LC) {
+        if (toPort == id_O) {
+            if (cell->lcInfo.dffEnable)
+                return false;
+            // "false paths"
+            if (fromPort == id_I0 && ((cell->lcInfo.lutInputMask & 0x1U) == 0))
+                return false;
+            if (fromPort == id_I1 && ((cell->lcInfo.lutInputMask & 0x2U) == 0))
+                return false;
+            if (fromPort == id_I2 && ((cell->lcInfo.lutInputMask & 0x4U) == 0))
+                return false;
+            if (fromPort == id_I3 && ((cell->lcInfo.lutInputMask & 0x8U) == 0))
+                return false;
+        }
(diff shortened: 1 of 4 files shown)


ice40/pack.cc
```diff
@@ -266,6 +266,75 @@ static void pack_carries(Context *ctx)
...
+            // Check cardinality
+            bool lut2_or_less = true;
+            for (auto lut_port : {id_I1, id_I0}) { // NOTE: LUT MSBs used first
+                if (drv->getPort(lut_port)) {
+                    lut2_or_less = false;
+                    break;
+                }
+            }
+            if (!lut2_or_less)
+                continue;
```
Ravenslofty, line 322: Having played around with the patch a bit, it seems like you might as well rewrite it as 
```suggestion
            bool uses_i1 = drv->getPort(id_I1);
            bool uses_i0 = drv->getPort(id_I0);
            if (uses_i1 || uses_i0)
                continue;
```
Or maybe even just inline those conditions into the if.

ice40/arch.cc
```diff
@@ -1231,6 +1242,21 @@ void Arch::assignCellInfo(CellInfo *cell)
...
+        cell->lcInfo.lutInputMask = 0x0;
+        unsigned init = int_or_default(cell->params, id_LUT_INIT);
+        for (unsigned k = 0; k < 4; k++) {
+            bool care = false;
+            for (unsigned i = 0; i < 16; i++) {
+                // If toggling the LUT input makes a difference it's not a don't care
+                if (((init >> i) & 0x1U) != ((init >> (i ^ (1U << k))) & 0x1U)) {
+                    care = true;
+                    break;
+                }
```
Ravenslofty, line 1255: Hmm. How about 
```suggestion
                // If toggling the LUT input makes a difference it's not a don't care
                care = ((init >> i) & 0x1U) != ((init >> (i ^ (1U << k))) & 0x1U));
                if (care)
                    break;
```

Ravenslofty: Actually, if you have an `if (care)` condition, you might as well also insert the contents of the `if (care)` block below this loop.

gatecat: not sure what you mean by below this loop? and I must admit the original is more in line with my personal style, although the suggestion is pretty reasonable too.

gatecat: oh I see, I can get rid of `care` altogether, that makes sense

Ravenslofty (approved):

PR #972, opened 2022-04-07
ecp5: Split the SLICE bel into separate LUT/FF/RAMW bels

gatecat:
This is one of those big scary PRs as far as the size of the change goes! :D

There are still some possible small QoR and runtime regressions with this that are being worked on, although most designs should not be significantly worse; and importantly some high utilisations designs that were previously failing to fit or route now build successfully.

This also helps align things better with the future direction of nextpnr and clears up some barnacles.


Ravenslofty (approved):
This is a touch big for me to give it the attention it deserves, but, looks good to me.

PR #982, opened 2022-05-08
ice40: Fix propagation of constraints through SB_GB

gatecat:
Signed-off-by: gatecat <gatecat@ds0.me>


smunaut:
Confirmed it works for me !

PR #985, opened 2022-05-11
[interchange] Tying unused LUT inputs according to architecture

mkurc-ant:
This PR makes nextpnr use the preferred constant net for unused LUT inputs as specified in FPGA interchange device resources. In code the `VCC` pin mask is now replaced by a vector of `PinConnection` which stores how a particular pin should be tied.


mkurc-ant:
@gatecat Ok, I understand. Anyways, I've fixed the last issue by using VCC as the preferred constant if the arch does not specify one. Could you please review?

mkurc-ant:
Thanks!

PR #993, opened 2022-05-27
gowin: Add support for long wires

yrabbit:
Gowin chips have a highly sophisticated system of long wires that are
wired to each cell and allow the clock or logic to spread quickly.

This commit implements some of the capabilities of the long wire system
for quadrants, leaving out the fine-tuning of them for each column.

To make use of the long wire system, the specified wire is cut at the
driver and a special cell is placed between the driver and the rest of
the wire.

* VCC and GND can not use long wires because they are in every cell and
  there is no point in using a net
* Long wire numbers can be specified manually or assigned automatically.
* The route from the driver to the port of the new cell can be quite
  long, this will have to be solved somehow.
* It might make sense to add a mechanism for automatically finding
  candidates for long wires.


yrabbit:
Compatible with older Apicula chip bases

PR #995, opened 2022-06-05
Gowin: WIP shadowram

pepijndevos:
In https://github.com/YosysHQ/apicula/pull/80/files shadow ram was added to Apicula, this is a WIP PR for adding it to Nextpnr

So far it compiled but I haven't gotten around to do any testing yet and am out of time for now.

I'm particularly concerned about all the wire juggling. This is as usual pretty much a copy of ECP5, but ECP5 seems to have a quite complicated `dram_to_comb` for reasons that are not entirely clear to me, and a few other stuff I'm not sure we need. I fully expect this to segfault as soon as I run it because there is some dangling pointer after packing.


gatecat:
> I'm particularly concerned about all the wire juggling. This is as usual pretty much a copy of ECP5, but ECP5 seems to have a quite complicated dram_to_comb for reasons that are not entirely clear to me, and a few other stuff I'm not sure we need. I fully expect this to segfault as soon as I run it because there is some dangling pointer after packing.

This was just the nicest way to implement it in the end. Do bear in mind that I believe ECP5 and Gowin may have now diverged, as ECP5 is now a finer-grained implementation in nextpnr with four bels per slice ('_COMB' containing LUT/carry/RAM and '_FF' containing the FF)

pepijndevos:
This seems to work now https://twitter.com/pepijndevos/status/1543291220084068357

Only thing I'm not sure about is... if the DFF can be used while using lutram, because the current implementation can't, which is a safe default, but potentially not optimal.

PR #998, opened 2022-06-23
gowin: add a separate router for the clocks

yrabbit:
A simple router that takes advantage of the fact that in each cell with
DFFs their CLK inputs can directly connect to the global clock network.

Networks with a large number of such sinks are sought and then each
network is assigned to the available independent global clock networks.

There are limited possibilities for routing mixed networks, that is,
when the sinks are not only CLKs: in this case an attempt is made to use
wires such as SN10/20 and EW10/20, that is, one short transition can be
added between the global clock network and the sink.

* At this time, networks with a source other than the I/O pin are not
  supported. This is typical for Tangnano4k and runber boards.

* Router is disabled by default, you need to specify option
  --enable-globals to activate

* No new chip bases are required. This may change in the distant future.


gowin/globals.cc
```diff
@@ -0,0 +1,335 @@
+/*
+ *  nextpnr -- Next Generation Place and Route
+ *
+ *  Copyright (C) 2018  gatecat <gatecat@ds0.me>
```
gatecat, line 4: please feel free to add your own copyright here!

gowin/globals.cc
```diff
@@ -0,0 +1,335 @@
...
+                }
+                continue;
+            }
+            used_pips.insert(spine_pip_id);
+
+            // >>> SPINExx <- IO
+            dstWire = ctx->getPipSrcWire(spine_pip_id);
+            dstWireInfo = ctx->wire_info(dstWire);
+            PipId io_pip_id = PipId();
+            for (auto const &uphill_pip : ctx->getPipsUphill(dstWire)) {
```
gatecat, line 278: nitpick but because `PipId` is a small register-size type you don't need the `&` here

yrabbit: What was I thinking of? Fixed.

gowin/arch.cc
```diff
@@ -543,6 +553,28 @@ void Arch::setDelayScaling(double scale, double offset)
     args.delayOffset = offset;
 }
 
+void Arch::addCellTimingCombIn(IdString cell, IdString port) { cellTiming[cell].portClasses[port] = TMG_COMB_INPUT; }
```
gatecat, line 556: are all these helper functions really needed? wouldn't one function with an extra class argument make more sense?

PR #1000, opened 2022-06-25
ice40: Fix accidental creation of empty ports

gatecat:
Also, add some `const`s so we can't make the same mistake again


dragonmux:
We can confirm this fixes the bug we encountered, and without affecting f(max)

PR #1006, opened 2022-07-29
gowin: auto allocation of long wires

yrabbit:
A simple long line router is added. Only the CE ports are handled so
far.

The router searches for all networks with such ports and assigns the
available long lines to the networks with the most ports.

The router is disabled by default, enabled by the command line key
--enable-auto-longwires


yrabbit:
And these innovations are compatible with the Apicula master branch, but are not friendly with the release so will be ready to merge when the new Apicula release comes out.

yrabbit:
Due to the discovery of a more complex mechanism of interaction between long wires and local buses, this PR has become obsolete.

PR #1015, opened 2022-08-17
fabulous: Add a viaduct uarch

gatecat:
This is WIP viaduct support for https://github.com/FPGA-Research-Manchester/FABulous

Current status:
 - a basic fabric example can be loaded (bels, wires, pips) and a tiny test design successfully built for it
 - there's data structures developed to enable much greater fabric configurability than at present (LUT structure, CLB routing, etc etc) although much of this isn't actually implemented this. the big required change will be fully supporting LUT+FF as separate bels (in nextpnr, they can be combined in fabulous) for fabric where it's viable to use LUTs and FFs separately, like Xilinx, ECP5, Intel, etc all are - and are implemented that way in nextpnr
 - we should also be splitting the MUX8 into separate bels, similar to Xilinx, or including it in the COMB/LC bels (like we do for ECP5) so we can do things like create two 2LUT4+MUX2 clusters using relative constraints and have the placer put them at different offsets in the same slice (again, this is a pure nextpnr bel tweak, no fabulous work required)
 - it would be nice to have a better way of specifying custom IP bels, but I believe this is a WIP on the fabulous side
 - other things missing but I'd like to work on:
   - LUT route-throughs and permutation
   - carry chains of various types, including the ones currently used
   - the aforementioned MUXs


gatecat:
Merging this initial work for now, some better ways to test are still needed but the basics seem to be there.

PR #1016, opened 2022-08-21
Use CMake's Python3 rather than PythonInterp in subdirs

atsampson:
ae8966040bc7f148958d4ac0b989e04059b6ba06 converted the top-level CMakeLists.txt to use the Python3 module rather than PythonInterp, but there are also uses of it - and the variables it defines - in subdirectories.

I haven't tested the Windows command in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ sudo make install
 On Windows, you may specify paths explicitly:
 
 ```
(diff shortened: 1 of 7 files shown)


gatecat:
Thanks!

mmicko:
Thanks for update, oss-cad-suite build scripts are now updated and working well. https://github.com/YosysHQ/oss-cad-suite-build/commit/e64a6bf29ca4c9f6448710ba25adc835086f92af

PR #1017, opened 2022-08-22
Router fix

mmicko:
get_togo_cost is used with parameters in different order than defined. Noticed error when using clang as compiler.


gatecat:
Looks good, just want to run a couple of quick tests out of curiosity which will take a few mins

gatecat:
Yea, all seems fine :)

PR #1019, opened 2022-08-31
[WIP] Support cross-domain clock relations in timing analyser

mkurc-ant:
This PR is a proof of concept of doing cross-domain timing checks. In general case two clock domains are independent. However, if both clocks come from the same source their frequencies are equal and they differ just in phase (or delay). This can be used to do timing checks for cross-domain paths whenever they are related.

The implemented algorithm looks for a common driver for each clock pair. If one is found then those clocks are considered related and a delay difference from that driver to their nets is stored. The delay is then taken into account when doing timing checks. A check is performed for cross-domain path only if its clocks are related.

### DIFF
--- a/common/kernel/timing.cc
+++ b/common/kernel/timing.cc
@@ -29,12 +29,17 @@
 
 NEXTPNR_NAMESPACE_BEGIN
 
+namespace {
+const char *edge_name(ClockEdge edge) { return (edge == FALLING_EDGE) ? "negedge" : "posedge"; }
+} // namespace
+
 void TimingAnalyser::setup()
 {
     init_ports();
     get_cell_delays();
     topo_sort();
     setup_port_domains();
+    identify_related_domains();
     run();
 }
 
@@ -249,6 +254,7 @@ void TimingAnalyser::setup_port_domains()
                     copy_domains(port, CellPortKey(pi.net->driver), true);
             }
         }
+
         // Iterate over ports and find domain paris
         for (auto port : topological_order) {
             auto &pd = ports.at(port);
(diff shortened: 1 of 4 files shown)


rowanG077:
This could probably be extended for related clocks even if their frequencies differ. Assuming you have two clocks A and B from the same source. if you cross from A to B the max delay between the crossing register should be the min (period of A) (period of B).

gatecat:
My one concern here would be in a more complex architecture where the routing delays for the two clocks also differed. This would require bigger changes to the timing code to care about clock routing (and to reduce runtime I'd rather not check this other than during signoff).

What do you think about that?

mkurc-ant:
@rowanG077 You're right. Your approach will hold provided that the ratio of the frequencies is a rational number. Otherwise the edge-to-edge delay between the clocks would continuously drift.

@gatecat That is my concern as well. I think we could start with analyzing clock route delays in post-routing analysis only and introduce leveraging it during the routing later on.

rowanG077:
@mkurc-ant

Yes you'd indeed require integer multiples. But even if not I think you can calculate it.

- Find the least common multiple of the clock periods for clocks A and B. This is the length of periods we need to analyze as the pattern will repeat afterwards
- Go through every subsequent pair of rising clock edges from A to B, and calculate the delay between them.
- The minimum of the delay is the "worst effective clock period".

This way you could handle any fixed frequency and phase difference.

gatecat:
@rowanG077 this is drifting a bit off topic, but I would recommend some research and experimentation into how existing open and commercial tools are dealing with the multiple-related-but-not-the-same clocks problem. It would be best for nextpnr to behave as close as possible as to what's normally expected.

common/kernel/timing.cc
```diff
@@ -1216,6 +1392,10 @@ void timing_analysis(Context *ctx, bool print_histogram, bool print_fmax, bool p
                   (update_results && ctx->detailed_timing_report) ? &detailed_net_timings : nullptr);
     timing.walk_paths();
 
+    // Use TimingAnalyser to determine clock-to-clock relations
+    TimingAnalyser timingAnalyser (ctx);
+    timingAnalyser.setup();
```
gatecat, line 1397: another thought not for this PR but if you have some capacity for general nextpnr development (cc @rowanG077 too) - it would be nice to finally replace all the old timing check code with the new `timingAnalyser` that's already used during placement

mkurc-ant:
> * Find the least common multiple of the clock periods for clocks A and B. This is the length of time we need to analyze as the pattern will repeat afterwards
> * Go through every subsequent pair of rising clock edges from A to B, and calculate the delay between them.
> * The minimum of the delay is the "worst effective clock period".
> 
> This way you could handle any fixed frequency and phase difference. Do you agree?

@rowanG077 Indeed you are right, this should be possible.

mkurc-ant:
@gatecat Could you tell what's the minimal set of changes required to get this PR merged? There are certainly multiple ways those timing checks can be improved but IMO we should begin doing that incrementally.

gatecat:
If you think that some degree of routing analysis is possible quickly then I'd suggest adding that too.

Otherwise for now I would also recommend adding an `--ignore-rel-clk` option or similar as a get-out if this breaks for someone.

mkurc-ant:
@gatecat Ok. For now I added the commandline option you suggested.

gatecat:
I think the WIP can be removed now if you are OK with this being merged...

The one CI failure looks unrelated, I'm investigating now.

gatecat (approved):

gatecat:
edit: the CI failure isn't unrelated, looks like you accidentally committed a `tests` submodule change (I know this is easy to do accidentally and not you're fault). hopefully reverting that fixes the PR.

mkurc-ant:
@gatecat My bad. Should be fixed now.

gatecat:
Thanks!

uis246:
Nextpnr now has more features. Good.

PR #1033, opened 2022-10-09
ice40: improved chipdb generation performance

lethalbit:
This PR replaces the normal python `re.match` and `re.find` calls in python with AOT compiled versions of the same calls, as most of these are in hot areas of the code, like in the `pipdelay` and `cmp_wire_names`, this improves performance.

For example, when generating the UP5K chipdb with the following command:
```shell
$ python -m cProfile -o up5k.prof ../ice40/chipdb.py -p ../ice40/constids.inc -g ../ice40/gfx.h --slow $ICEBOX_INSTALL/timings_up5k.txt $ICEBOX_INSTALL/chipdb-5k.txt > chipdb-up5k.new
```

Prior to changes:
```
Welcome to the profile statistics browser.
up5k-pre.prof% stats 0
Sun Oct  9 09:15:49 2022    up5k-pre.prof

         133904934 function calls (133904766 primitive calls) in 74.772 seconds
```

Post changes:
```
Welcome to the profile statistics browser.
up5k-post.prof% stats 0
Sun Oct  9 09:12:57 2022    up5k-post.prof

         131131017 function calls (131130849 primitive calls) in 74.060 seconds
```

Currently the other main bottleneck is the BBA emission, where each line is printed to stdout as it's called. There is a possible improvement there by possibly exchanging memory usage for speed. And also a lot of the string-ops are quite expensive as well.

While it is a small perf improvement, things like this stack, so I feel it's worth it.

### DIFF
--- a/ice40/chipdb.py
+++ b/ice40/chipdb.py
@@ -311,6 +311,7 @@ def norm_wire_xy(x, y, name):
     return None # FIXME
     return (x, y)
 
+wire_name = re.compile(r".*_(\d+)$")
 def cmp_wire_names(newname, oldname):
     if maj_wire_name(newname):
         return True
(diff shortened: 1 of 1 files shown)


whitequark:
> This PR replaces the normal python `re.match` and `re.find` calls in python with AOT compiled versions of the same calls

Python does the same thing internally; it has an LRU cache in the `re` module (see the Note in https://docs.python.org/3/library/re.html#re.compile). All this PR does is avoids querying this cache.

PR #1035, opened 2022-10-17
Correct Not Equal operator implementation in ice40

tyler274:
I noticed this during my work reimplementing nextpnr, and it seems to be dead and wrong, or at least dead. Either way I think this is what was intended unless anyone can correct me.


gatecat:
Thanks! Seems like this somehow never hit a case where it mattered

PR #1040, opened 2022-11-10
gowin: add initial PLL support

yrabbit:
The rPLL primitive for the simplest chip (GW1N-1) in the family is processed. All parameters of the primitive are passed on to gowin_pack, and general-purpose wires are used for routing outputs of the primitive.

Compatible with older versions of apicula, but in this case will refuse to place the new primitive.


gowin/cells.cc
```diff
@@ -176,6 +193,53 @@ void gwio_to_iob(Context *ctx, CellInfo *nxio, CellInfo *iob, pool<IdString> &to
     }
 }
 
+void reconnect_rplla(Context *ctx, CellInfo *pll, CellInfo *plla)
+{
+    pll->movePortTo(id_CLKIN, plla, id_CLKIN);
+    pll->movePortTo(id_CLKFB, plla, id_CLKFB);
+    pll->movePortTo(ctx->id("FBDSEL[0]"), plla, id_FBDSEL0);
```
gatecat, line 200: you could use for loops (and `ctx->idf`) here to reduce the code length a bit

gowin/arch.cc
```diff
@@ -1071,6 +1071,26 @@ void Arch::addMuxBels(const DatabasePOD *db, int row, int col)
...
+{
+    IdString portname;
+    char buf[40];
+
+    for (int pid : {ID_CLKIN,   ID_CLKFB,   ID_FBDSEL0, ID_FBDSEL1, ID_FBDSEL2, ID_FBDSEL3, ID_FBDSEL4, ID_FBDSEL5,
+                    ID_IDSEL0,  ID_IDSEL1,  ID_IDSEL2,  ID_IDSEL3,  ID_IDSEL4,  ID_IDSEL5,  ID_ODSEL0,  ID_ODSEL1,
+                    ID_ODSEL2,  ID_ODSEL3,  ID_ODSEL4,  ID_PSDA0,   ID_PSDA1,   ID_PSDA2,   ID_PSDA3,   ID_DUTYDA0,
+                    ID_DUTYDA1, ID_DUTYDA2, ID_DUTYDA3, ID_FDLY0,   ID_FDLY1,   ID_FDLY2,   ID_FDLY3}) {
+        portname = IdString(pairLookup(bel->ports.get(), bel->num_ports, pid)->src_id);
+        snprintf(buf, 32, "R%dC%d_%s", row + 1, col + 1, portname.c_str(this));
```
gatecat, line 1084: as a general code comment it might be worth refactoring to use `ctx->idf(format, ...)` to build formatted idstrings rather than the snprintf into a buf approach to tidy things up a bit - `idf` won't be quite as fast but for all the non performance critical places like here it should be fine enough...

yrabbit: I myself am not thrilled with the constant snprintf(buf...) and extensive refactoring is prescribed in the plan. I'm more inclined to total transition to IdStringList if manipulations with it to assemble identifiers from several ones will be more effective than now to replace everything with ctx->idf().

One way or another, but gowin-part will be modified, another thing is that this stage is the sixth or seventh in the plan:)

gatecat: moving to `IdStringList` sounds like a great long-term plan, yeah

PR #1046, opened 2022-11-25
Adding initial placer_star

gatecat:
This is an initial attempt to investigate whether the Star wirelength heuristic with timing additions as per [this thesis](https://atrium.lib.uoguelph.ca/xmlui/bitstream/handle/10214/21275/Martin_Timothy_202009_MSc.pdf) could give better results than the current bound2bound based HeAP placer.

Currently, the answer is no (although it's not really worse either on the small test so far), but there's plenty of further investigation remaining. And there might be more scope for parallelism than HeAP and other ideas to explore here too...

### DIFF
--- a/common/kernel/array2d.h
+++ b/common/kernel/array2d.h
@@ -1,0 +1,134 @@
+/*
+ *  nextpnr -- Next Generation Place and Route
+ *
+ *  Copyright (C) 2021-22  gatecat <gatecat@ds0.me>
+ *
+ *  Permission to use, copy, modify, and/or distribute this software for any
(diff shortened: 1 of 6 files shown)


cbalint13:
@gatecat

On similar topic here, not sure if you are aware of this recent work:
  * https://arxiv.org/abs/2210.08682
  * https://github.com/zslwyuan/AMF-Placer

gatecat:
Unfortunately didn't make much progress in getting any meanigful benefits from this, but I'm still going to look into other ways of improving placer QoR.

Issue #1062, opened 2022-12-15
Main analytical placer runs does not (seem to) make progress after modest increase in logic


arjenroodselaar:
I am working on a design using a LFE5U45-CABGA554 which is approaching ~60% of the available LUT resources. It seems like the logic is getting harder to place as small changes to the design may cause the main analytical phase of the HeAP placer to seemingly not make progress. It may complete one or two iterations, which on a successful run take about 5 seconds each, but more often than not run until I stop the process (I let it run overnight once without reaching a solution, for a total of almost 7 hours). With some patience I can find a seed value where the placer will run to completion and the design closes timing, but which seeds works changes each time I make changes to the design.

The device utilization is currently as follows:

```
Info: Device utilisation:
Info:             TRELLIS_IO:   142/  245    57%
Info:                   DCCA:     1/   56     1%
Info:                 DP16KD:     2/  108     1%
Info:             MULT18X18D:     0/   72     0%
Info:                 ALU54B:     0/   36     0%
Info:                EHXPLLL:     0/    4     0%
Info:                EXTREFB:     0/    2     0%
Info:                   DCUA:     0/    2     0%
Info:              PCSCLKDIV:     0/    2     0%
Info:                IOLOGIC:     0/  160     0%
Info:               SIOLOGIC:     0/   85     0%
Info:                    GSR:     0/    1     0%
Info:                  JTAGG:     0/    1     0%
Info:                   OSCG:     0/    1     0%
Info:                  SEDGA:     0/    1     0%
Info:                    DTR:     0/    1     0%
Info:                USRMCLK:     0/    1     0%
Info:                CLKDIVF:     0/    4     0%
Info:              ECLKSYNCB:     0/   10     0%
Info:                DLLDELD:     0/    8     0%
Info:                 DDRDLL:     0/    4     0%
Info:                DQSBUFM:     0/   10     0%
Info:        TRELLIS_ECLKBUF:     0/    8     0%
Info:           ECLKBRIDGECS:     0/    2     0%
Info:                   DCSC:     0/    2     0%
Info:             TRELLIS_FF: 14703/43848    33%
Info:           TRELLIS_COMB: 28016/43848    63%
Info:           TRELLIS_RAMW:     0/ 5481     0%
```

Before I go off and read both the HeAP paper and try and instrument the code to determine what happens, a couple of questions:

[...]

PR #1066, opened 2022-12-18
Timeout when legal placement can't be found for cell

arjenroodselaar:
The `iter` variable used to generate a timeout if no valid placement can be found for a cell may get reset in https://github.com/YosysHQ/nextpnr/blob/78926b31db6dbfb465e2585ce0050bf93cb3e35c/common/place/placer_heap.cc#L930 The design in https://github.com/YosysHQ/nextpnr/issues/1062 contains a cell where the radius and ripup radius increase to 90 and after that it seems to never find a bel which satisfies the contraints. But the `notempty` label is consistently hit every 909 iterations while searching, causing `iter` to become 0 before it will ever trigger the timeout. The process will happily run for hours without going anywhere.

I think there is a reason this variable gets reset so a separate total is added to track the total number iterations allowing the timeout to be triggered. Note that this counter is only incremented when an actual coordinate is visited, allowing it to try up to 8 times the number of cells in the design.

With this patch applied the design in https://github.com/YosysHQ/nextpnr/issues/1062 properly errors providing the user with a clue that utilization may be too high.
```
Clock '$glbnet$clk_50m_fpga_refclk$TRELLIS_IO_IN' can be driven by:
 clk_50m_fpga_refclk$tr_io.O delay 0.000ns
Info: Placed 142 cells based on constraints.
Info: Creating initial analytic placement for 29349 cells, random placement wirelen = 2284884.
Info:     at initial placer iter 0, wirelen = 20720
Info:     at initial placer iter 1, wirelen = 13471
Info:     at initial placer iter 2, wirelen = 11056
Info:     at initial placer iter 3, wirelen = 11050
Info:     at iteration #1, type ALL: wirelen solved = 11382, spread = 318213, legal = 423187; time = 4.06s
ERROR: Unable to find legal placement for cell 'MUX_controller_ignition_controllers_27_hello_expired_count_r$write_1__VAL_2_LUT4_Z_2' after 128593 attempts, check constraints and utilisation.
0 warnings, 1 error
```

### DIFF
--- a/common/kernel/command.cc
+++ b/common/kernel/command.cc
@@ -189,6 +189,7 @@ po::options_description CommandHandler::getGeneralOptions()
     general.add_options()("placer-heap-critexp", po::value<int>(),
                           "placer heap criticality exponent (int, default: 2)");
     general.add_options()("placer-heap-timingweight", po::value<int>(), "placer heap timing weight (int, default: 10)");
+    general.add_options()("no-placer-timeout", "allow the placer to attempt placement without timeout");
 
 #if !defined(__wasm)
     general.add_options()("parallel-refine", "use new experimental parallelised engine for placement refinement");
@@ -327,6 +328,9 @@ void CommandHandler::setupContext(Context *ctx)
 
     if (vm.count("placer-heap-timingweight"))
         ctx->settings[ctx->id("placerHeap/timingWeight")] = std::to_string(vm["placer-heap-timingweight"].as<int>());
+
+    if (vm.count("no-placer-timeout"))
+        ctx->settings[ctx->id("placerHeap/noTimeout")] = true;
 
     if (vm.count("parallel-refine"))
         ctx->settings[ctx->id("placerHeap/parallelRefine")] = true;
--- a/common/kernel/command.cc
+++ b/common/kernel/command.cc
@@ -189,7 +189,8 @@ po::options_description CommandHandler::getGeneralOptions()
     general.add_options()("placer-heap-critexp", po::value<int>(),
                           "placer heap criticality exponent (int, default: 2)");
     general.add_options()("placer-heap-timingweight", po::value<int>(), "placer heap timing weight (int, default: 10)");
-    general.add_options()("no-placer-timeout", "allow the placer to attempt placement without timeout");
+    general.add_options()("placer-heap-cell-placement-timeout", po::value<int>(),
+            "allow placer to attempt up to the given number of iterations to place the cell (int, default: design size^2 / 8, 0 for no timeout)");
 
 #if !defined(__wasm)
     general.add_options()("parallel-refine", "use new experimental parallelised engine for placement refinement");
@@ -329,8 +330,9 @@ void CommandHandler::setupContext(Context *ctx)
     if (vm.count("placer-heap-timingweight"))
         ctx->settings[ctx->id("placerHeap/timingWeight")] = std::to_string(vm["placer-heap-timingweight"].as<int>());
 
-    if (vm.count("no-placer-timeout"))
-        ctx->settings[ctx->id("placerHeap/noTimeout")] = true;
+    if (vm.count("placer-heap-cell-placement-timeout"))
+        ctx->settings[ctx->id("placerHeap/cellPlacementTimeout")] =
+            std::to_string(std::max(0, vm["placer-heap-cell-placement-timeout"].as<int>()));
 
     if (vm.count("parallel-refine"))
         ctx->settings[ctx->id("placerHeap/parallelRefine")] = true;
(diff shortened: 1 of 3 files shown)


gatecat:
Unfortunately, one of the regression tests for ice40 is failing here, probably with a global buffer. The radius approach might need to be changed for seldom cells like this - I can look into this but it might be a few days.

arjenroodselaar:
Hmm.. I actually happen to have a iCE40 LP1K design which is at 78% utilization and also fails with this change applied. The cell it can't place (`txr_rx_channels_0_expect_idle_SB_DFFE_Q_DFFLC`) is not a global though (does not show up in the list of promoted signals), so finding an appropriate upper limit here may be less straightforward than it looks.

arjenroodselaar:
I took the naive approach and simply made the timeout much larger. The design in https://github.com/YosysHQ/nextpnr/issues/1062 tests about 1M iterations every 1-2 seconds on my machine and thus times out after about 5 minutes. I am sure we can bring this down, but given that I am not looking to create more work for you in the short term this seems like an improvement over never timing out. With this timeout both the attasoc smoketest and my LP1K design place as expected.

cbalint13:
> I took the naive approach and simply made the timeout much larger.

@arjenroodselaar 

Can a CLI switch be added to control this new timeout by users who might want override ? 
Also, a value of e.g -1 to preserve its infiniteness behaviour (with a small comment mentioning this).

arjenroodselaar:
@cbalint13 please disregard my last commit, I did not interpret your request correct. I'll look at this again with fresh eyes tomorrow morning, like I should have done in the first place.

common/kernel/command.cc
```diff
@@ -189,6 +189,7 @@ po::options_description CommandHandler::getGeneralOptions()
     general.add_options()("placer-heap-critexp", po::value<int>(),
                           "placer heap criticality exponent (int, default: 2)");
     general.add_options()("placer-heap-timingweight", po::value<int>(), "placer heap timing weight (int, default: 10)");
+    general.add_options()("no-placer-timeout", "allow the placer to attempt placement without timeout");
```
cbalint13, line 192: I would suggest to reflect the usage options text to users more clearly:
add_options()("placer-timeout", "....placer timeout.....,  XXXXX seconds default, -1 means disabled")

Feel free to do other default/no-default/counter/seconds variations.

cbalint13:
> @cbalint13 please disregard my last commit, I did not interpret your request correct. I'll look at this again with fresh eyes tomorrow morning, like I should have done in the first place.

@arjenroodselaar 
No problem.
Thanks for the commit, on my side I also found this quite useful !
I found it useful in context of [tuning parameters](https://github.com/cbalint13/nextpnr-anneal) (where some variations might run endlessly).

arjenroodselaar:
> > @cbalint13 please disregard my last commit, I did not interpret your request correct. I'll look at this again with fresh eyes tomorrow morning, like I should have done in the first place.
> 
> @arjenroodselaar No problem. Thanks for the commit, on my side I also found this quite useful ! I found it useful in context of [tuning parameters](https://github.com/cbalint13/nextpnr-anneal) (where some variations might run endlessly).

Oh neat! I've been thinking about a tuning script like this. We have one design in particular which will be "done" at some point and effectively put in ROM. Before doing so I was hoping to find some subset of parameters that would give us the most design margin (to account for device/temperature variation etc.) given the intended frequency of the design.

I think I understand what you're after. Let me make some changes.

common/kernel/command.cc
```diff
@@ -189,6 +189,8 @@ po::options_description CommandHandler::getGeneralOptions()
     general.add_options()("placer-heap-critexp", po::value<int>(),
                           "placer heap criticality exponent (int, default: 2)");
     general.add_options()("placer-heap-timingweight", po::value<int>(), "placer heap timing weight (int, default: 10)");
+    general.add_options()("placer-heap-cell-placement-timeout", po::value<int>(),
+            "allow placer to attempt up to the given number of iterations to place the cell (int, default: design size^2 / 8, 0 for no timeout)");
```
cbalint13, line 193: I believe your idea to control in terms of array (squared) is nice, but why not allow user control the divisor ?

```
  general.add_options()("placer-heap-cell-placement-timeout", po::value<int>(),
     "allow placer to attempt up to (design_size^2 / N) of iterations to place the cell 
      (int N, default: 8, 0 for no timeout)");
```

* So in this way we could end up: 
      N = 8 default, otherwise what user input,
   *  if N > 0 : (design_size^2) / N 
   *  if N = 0 : no-timeout

arjenroodselaar: Oh hmm.. sure, I hadn't considered that.

arjenroodselaar: Third time is a charm! This is much better than an absolute value, allowing a user to clamp down or run unlimited. Apologies for being slow to catch on here.

cbalint13 (approved):
Thanks for accounting my suggestion !
Still have to wait @gatecat final approval.

PR #1074, opened 2023-01-01
CXXify awooter (almost)

auramarua:
Unfortunately, even with setting the libraries to PUBLIC (not a desirable situation!) this branch does not yet fully link and execute, and getting the code generated into the right locations is a bit of a struggle to begin with. I am just quite stuck on the last few steps of making this build. I don't rightly understand CMake, which is part of the problem, and the other half of the problem is that Corrosion, the CMake helper in use, does not fully understand [cxx](https://cxx.rs).

I intend to revisit what cxx and the Corrosion helper is doing soon and assemble the final missing piece to make this actually build, link, and run, but you might better know the magick incantations, @Ravenslofty.


CMakeLists.txt
```diff
@@ -94,9 +94,10 @@ endif()
 
 if (AWOOTER)
     include(FetchContent)
-    FetchContent_Declare(Corrosion GIT_REPOSITORY https://github.com/corrosion-rs/corrosion.git GIT_TAG v0.3.0)
+    FetchContent_Declare(Corrosion GIT_REPOSITORY https://github.com/corrosion-rs/corrosion.git GIT_TAG master)
```
Ravenslofty, line 97: I think it'd be kind of inconvenient if some upstream change broke the build, which is why I specified a version rather than simply a branch.

CMakeLists.txt
```diff
@@ -312,7 +313,11 @@ foreach (family ${ARCH})
 
     # Link in awooter, if enabled.
     if (AWOOTER)
-        target_link_libraries(${PROGRAM_PREFIX}nextpnr-${family} PRIVATE awooter)
+        include_directories(common/route/awooter 
+            common/kernel/ common/place/ common/route/ "${family}/")
```
Ravenslofty, line 317: I'm not entirely sure about this; doesn't this duplicate the `aux_source_directories`? https://github.com/YosysHQ/nextpnr/blob/fc01f13cfde4e128c3a4c651636ade328d8d5eae/CMakeLists.txt#L257-L259

common/route/awooter/awooter.h
```diff
@@ -16,15 +16,26 @@
  *  OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  *
  */
-
+#pragma once
```
Ravenslofty, line 19: This just duplicates the include guards, you don't need it.

common/route/awooter/awooter.h
```diff
@@ -16,15 +16,26 @@
...
 extern bool router_awooter(Context* ctx);
 
+uint64_t npnr_netinfo_users_leak(NetInfo &net, rust::Vec<PortRef> &vec) {
+    for (auto& item : net.users) {
+        vec.push_back(item);
+    }
+    vec.shrink_to_fit();
+    return vec.size();
+}
+
```
Ravenslofty, line 38: This kind of makes me uncomfortable; could you not just provide a prototype and keep the definition inside `awooter.cc`?

common/route/awooter/src/lib.rs
```diff
@@ -102,49 +135,46 @@ fn route(ctx: &mut npnr::Context, pressure: f32, history: f32) -> bool {
...
     let nets_str = nets.len().to_string();
     log_info!("Found {} nets\n", nets_str.bold());
 
     let mut count = 0;
-    for (&name, net) in nets.to_vec().iter() {
-        let _src = ctx.source_wire(**net);
-        let net = unsafe { net.as_mut().unwrap() };
+    for (&name, net) in &nets.nets {
+        // let _src = ctx.source_wire(net);
+        // let net = unsafe { net.as_mut().unwrap() };
```
Ravenslofty, line 144: Just delete them.

common/route/awooter/src/npnr.rs
```diff
@@ -0,0 +1,371 @@
...
+}
+
+unsafe impl Send for Nets {}
+unsafe impl Sync for Nets {}
+
+impl Nets {
+    /// Create a new store for the nets of a context.
+    ///
+    /// Note that this leaks memory created by nextpnr; the intention is this is called once.
+
```
Ravenslofty, line 265: Just delete it, since you inlined the code.

common/route/awooter/src/partition.rs
```diff
@@ -1376,14 +1373,16 @@ impl PipSelector {
             .flat_map(|pos| pips.get(&pos))
             .flat_map(|vec| vec.iter())
             .filter_map(|pip| {
-                if !ctx.pip_avail_for_net(*pip, raw_net) {
+                // checkPipAvailForNet should really have its NetInfo arg annotated as const,
+                // thus this dubious-looking cast and nonsense unsafety
+                if unsafe { !ctx.pip_avail_for_net(*pip, raw_net as *const _ as *mut _) } {
```
Ravenslofty, line 1378: You mean like in c62a947a2804c3027471b20815c7544e0fd698d0?

Ravenslofty (commented):
Better than I was expecting given the inevitable pain of FFI; I'm a bit confused about how some things now require mutability however.

CMakeLists.txt
auramarua: Most likely! I truly have no idea what I am doing with CMake.

CMakeLists.txt
auramarua: The Corrosion maintainer held back `corrosion_add_cxxbridge` from the v0.3.0 release for reasons I am not clear on (though given the number of issues that I have encountered and that have been opened against the repo, I can understand the reluctance). Supposedly it will land in a proper release by v0.4.0 which seems destined to arrive within a month or two.

In fixing the final cxx-build problems I may take a route that makes `corrosion_add_cxxbridge` temporarily irrelevant, which would be one way of getting rid of tagging master.

common/route/awooter/src/partition.rs
auramarua: ...as it turns out, yes!

common/route/awooter/awooter.h
auramarua: It apparently did not need to be exposed at all? Unsure. May revisit this after the rebase.

common/route/awooter/src/partition.rs
auramarua: Since it's not `const T& const`, that only gets us to matching `*const NetInfo` according to cxx. I still removed the lamentation from the comments since a cast to `*const _` is accordingly less peculiar.

common/route/awooter/awooter.cc
```diff
@@ -183,63 +171,46 @@ extern "C" {
...
+        return wires.size();
     }
 
+    size_t npnr_netinfo_users_leak(NetInfo &net, rust::Vec<PortRef> &vec) {
+        for (auto& item : net.users) {
+            vec.push_back(item);
+        }
+        vec.shrink_to_fit();
+        return vec.size();
+    }
```
SpaceCat-Chan, line 196: these no longer leak memory, i would recommend renaming them here and in rust accordingly

auramarua:
The current linking problem was resolved by rearranging directories somewhat (awooter.cc was no longer in a CMake's `aux_source_directory` and thus awooter.cc had started to elude being built). It now fails at a different stage of linking because of undefined references to the bridge code instead of to awooter's own router function! This is most likely due to a failure to instantiate templates or an incorrect inclusion direction, but that is more easily solved by twiddling the arrangement of things now that the problem is obvious and the details of CMake are no longer the primary concern.

common/route/awooter/awooter.cc
```diff
@@ -126,30 +127,17 @@ extern "C" {
...
-    uint64_t npnr_context_get_pip_src_wire(const Context *const ctx, uint64_t pip) { return wrap(ctx->getPipSrcWire(unwrap_pip(pip))); }
-    uint64_t npnr_context_get_pip_dst_wire(const Context *const ctx, uint64_t pip) { return wrap(ctx->getPipDstWire(unwrap_pip(pip))); }
-    float npnr_context_estimate_delay(const Context *const ctx, uint64_t src, uint64_t dst) { return ctx->getDelayNS(ctx->estimateDelay(unwrap_wire(src), unwrap_wire(dst))); }
-    float npnr_context_get_pip_delay(const Context *const ctx, uint64_t pip) { return ctx->getDelayNS(ctx->getPipDelay(unwrap_pip(pip)).maxDelay()); }
-    float npnr_context_get_wire_delay(const Context *const ctx, uint64_t wire) { return ctx->getDelayNS(ctx->getWireDelay(unwrap_wire(wire)).maxDelay()); }
+    BelId npnr_belid_null() { return BelId(); }
+    WireId npnr_wireid_null() { return WireId(); }
+    PipId npnr_pipid_null() { return PipId(); }
+
+    float npnr_context_estimate_delay(const Context *const ctx, const WireId &const src,  const WireId &const dst) { return ctx->getDelayNS(ctx->estimateDelay(src, dst)); }
```
gatecat, line 134: the first `const` in `const Context *` is definitely needed to signify that the object pointed to can't be modified (i.e. the pointer is an immutable reference) but is the `*const` - after the `*` really needed too? the pointer itself is passed by value

Ravenslofty:
I think given that the Rust bindings got merged upstream this can be closed, especially since the manual approach is at least much easier to integrate into the nextpnr build system.

PR #1076, opened 2023-01-04
ECP5: Add DSP signal remapping

adamgreig:
Apologies if the below is already well known to people, but a bit of ECP5 background for this PR:

In ECP5, the DSP tiles are grouped into blocks of 9 tiles on the same row, each block containing two slices, and each slice containing two MULT18X18D and one ALU54B. Each register in those multipliers and ALU units can have their clock selected from one of four clock ports CLK0..3, and similarly for CE and RST. Those ports are then connected to external signals on a per-unit basis, which means you can write HDL that has one MULT18X18D's CLK0 connected to one clock signal, and a second MULT18X18D with CLK0 connected to a different signal. Those two MULT18X18Ds may end up in the same block, if manually placed, or because the packer wanted to, or because they feed the same ALU.

However, each DSP block only has one actual external connection per CLK/CE/RST signal, so all CLK0s within a block must be the same external signal. Currently this isn't checked, and instead the router will either run forever or fail with an unconnectable signal error. If all the DSPs are being placed manually it's possible for the end user to rejig signals to avoid this problem, but because there are two slices per block, I found it was common to have e.g. two independent MACs (two MULT and an ALU each) packed into the same block by the packer. If those MACs had independent reset signals using the same name, routing would fail.

It turns out that Diamond works around this problem by automatically (and silently) remapping these signals. For example, if ALU1 and ALU2 in the same block both use RST0 for some registers, but both connect their RST0 port to different external signals, Diamond will change one of them to use RST1 for those registers and connect the signal to RST1 instead. If more than four distinct RST signals are used in a block, it errors out instead, since it's not possible to remap them all.

This PR implements equivalent logic. `Arch::isBelLocationValid` is updated to check that no more than four distinct CLK/CE/RST signals are used per DSP block, and `Arch::place` is updated to perform the remapping.

I've tested this on a variety of synthetic and real designs, including the one that led me to find the problem in the first place, and it seems to work as expected, and gives the same results as Diamond. I wrote this back in November so it's not as fresh in my mind any more but I only just got around to putting together a PR for it. Hopefully the code is reasonably self explanatory but of course I'm happy to discuss and update it as you see fit!


ecp5/arch_place.cc
```diff
@@ -190,20 +190,273 @@ bool Arch::isBelLocationValid(BelId bel, bool explain_invalid) const
...
+            size_t j = std::distance(
+                ports.cbegin(),
+                std::find(ports.cbegin(), ports.cend(), remap_ports[i]));
+            if (j != ports.size()) {
+                new_clk_div_props[j] = param->second;
+            }
+        }
+
+        for (size_t i = 0; i < 4; i++) {
+            if (new_clk_div_props[i].is_string) {
```
gatecat, line 443: would just like to double check what this `is_string` is about? I think it's just checking if `new_clk_div_props[i]` has been set (it will be false if it hasn't), if I understand this correctly maybe `if (new_clk_div_props[i] == Property())` would be cleaner?

gatecat:
Thanks, this is useful to have and really well written, much appreciated! Just one small comment left.

ecp5/arch_place.cc
adamgreig: Yep, it's just checking whether `new_clk_div_props[i]` was set in the previous loop. I think I just had `is_string` in my head from the previous chunk where I do actually access properties as strings. The only valid settings for this property are `"ENABLED"` and `"DISABLED"` so I guess it would always be a string, but nevertheless your suggestion is neater.

I think you mean `if (new_clk_div_props[i] != Property())` (`!=` instead of `==`)?

gatecat: Thanks for confirming - and yes, I did mean `!=`...

adamgreig: I've updated the PR with this change, thanks!

gatecat (approved):

PR #1081, opened 2023-01-24
use eigen as an IMPORTED target in CMake

danc86:
Eigen considers the EIGEN3_INCLUDE_DIRS and EIGEN3_DEFINITIONS variables to be deprecated and they will no longer be exported in the next release after 3.4.0:
https://gitlab.com/libeigen/eigen/-/commit/f2984cd0778dd0a1d7e74216d826eaff2bc6bfab

Use the IMPORTED target instead, which seems to be the preferred way of consuming third-party CMake libraries.


gatecat:
Thanks!

PR #1086, opened 2023-01-29
ice40: Improve `output` handling vs pull-ups and undriven

smunaut:
Previously if a port was declared as `output`, there would be two issues :

 - In all cases, the pull up settings from the PCF wouldn't be applied, even if there was a `TBUF`
 - If the port was bound to a constant net `x` or `z`, this would end up as PIN_TYPE `0110_01` ( `PIN_OUTPUT` ) with `D_OUT_0` no connection, which would result in the pin being actively driven.  ( And AFAICT it could be driven to either 0 or 1 depending on its position in the IO tile ? Although I have yet to reconfirm this ).

This fixes both of theses.

I'm marking this as Draft because I don't have access to hardware ATM to double check everything ... will do during the week, but still open for code review if anyone has comments.


gatecat (approved):
LGTM, unfortunately I'm also traveling so you'll probably get a chance to test it on hw before me.

smunaut:
Just confirmed it works on hardware as I'd expect it to.

PR #1087, opened 2023-01-30
gowin: Add PLL support for the GW1NR-9 chip

yrabbit:
And also unified the fixing of PLL to bels: the point is that PLL being at a certain location has the possibility to use a direct implicit wire to the clock source, but once we decide to use this direct wire, the PLL can no longer be moved.


gowin/arch.h
```diff
@@ -530,6 +531,14 @@ enum
 };
 }
 
+namespace PLL { // fixed PLL locations
+enum
```
gatecat, line 535: why not use an `enum class`?

yrabbit: Yes, it makes sense.

PR #1089, opened 2023-02-01
ice40: Add support for PLL ICEGATE function

smunaut:
Technically you can enable it independently on CORE and GLOBAL output, but this is not exposed in the classic primitive, so we do the same as icecube2 and enable/disable it for both output path depending on the argument


smunaut:
AFAICT the CI issues are because you need the matching icestorm update to add the new bits in the icebox DB.

gatecat:
Do you think you could make it not fail with older icebox unless the bits are actually used, to save people figuring out why things have broken?

smunaut:
Would this be ok ? 

Basically if we try to set any unknown ec bit to zero, then instead of asserting, we just print a warning.

```diff
diff --git a/ice40/bitstream.cc b/ice40/bitstream.cc
index 6794ecca..c150b9a7 100644
--- a/ice40/bitstream.cc
+++ b/ice40/bitstream.cc
@@ -153,7 +153,10 @@ static void set_ec_cbit(chipconfig_t &config, const Context *ctx, const BelConfi
             return;
         }
     }
-    NPNR_ASSERT_FALSE_STR("failed to config extra cell config bit " + name);
+    if (value)
+        NPNR_ASSERT_FALSE_STR("failed to config extra cell config bit " + name);
+    else
+        log_warning("failed to config extra cell config bit %s to zero (ignored, maybe update icestorm ?)", name.c_str());
 }
 
 void configure_extra_cell(chipconfig_t &config, const Context *ctx, CellInfo *cell,
```

ice40/bitstream.cc
```diff
@@ -153,7 +153,11 @@ static void set_ec_cbit(chipconfig_t &config, const Context *ctx, const BelConfi
             return;
         }
     }
-    NPNR_ASSERT_FALSE_STR("failed to config extra cell config bit " + name);
+    if (value)
+        NPNR_ASSERT_FALSE_STR("failed to config extra cell config bit " + name);
+    else
+        log_warning("failed to config extra cell config bit %s to zero (ignored, maybe update icestorm ?)",
```
gatecat, line 159: the log warning needs a \n at the end

smunaut: ack

Issue #804, opened 2021-08-24
DCSC doesn't propagate clock constraints


thirtythreeforty:
#754 added support for ECP5's DCSC primitive, but clock constraints are not propagated through it. I have an input clock running at 350MHz, however the output clock is assigned a default:

```
Info: Annotating ports with timing budgets for target frequency 12.00 MHz
```

the clock net (`cap2`) erroneously passes timing:
```
Info: Max frequency for clock          'cap2_clk': 348.80 MHz (PASS at 12.00 MHz)
Info: Max frequency for clock '$glbnet$sdrio_clk': 75.08 MHz (PASS at 60.00 MHz)
Warning: Max frequency for clock  '$glbnet$cap1_clk': 346.38 MHz (FAIL at 350.02 MHz)
Info: Max frequency for clock  '$glbnet$init_clk': 333.11 MHz (PASS at 96.01 MHz)
```

rowanG077:
@thirtythreeforty I know it's a while back. If you are still waiting for this please check if https://github.com/YosysHQ/nextpnr/pull/1090 works for you!

PR #1090, opened 2023-02-01
ecp5: Propagate clock constraints through DCSC

rowanG077:
I added a TODO. If `modesel` is tied to `0` and certain modes are selected we can push through either of the clock constraints directly. I couldn't quickly see how I can do this by grepping the codebase.  

 
Fixes https://github.com/YosysHQ/nextpnr/issues/804

### DIFF
--- a/ecp5/pack.cc
+++ b/ecp5/pack.cc
@@ -2602,6 +2602,15 @@ class Ecp5Packer
         auto MHz = [&](delay_t a) { return 1000.0 / ctx->getDelayNS(a); };
 
         auto equals_epsilon = [](delay_t a, delay_t b) { return (std::abs(a - b) / std::max(double(b), 1.0)) < 1e-3; };
+        auto equals_epsilon_pair = [&](DelayPair& a, DelayPair& b) {
+            return equals_epsilon(a.min_delay, b.min_delay)
+                && equals_epsilon(a.max_delay, b.max_delay);
+        };
+        auto equals_epsilon_constr = [&](ClockConstraint& a, ClockConstraint& b) {
+            return equals_epsilon_pair(a.high, b.high)
+                && equals_epsilon_pair(a.low, b.low)
+                && equals_epsilon_pair(a.period, b.period);
+        };
 
         pool<IdString> user_constrained, changed_nets;
         for (auto &net : ctx->nets) {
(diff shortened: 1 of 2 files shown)


ecp5/pack.cc
```diff
@@ -2706,6 +2721,80 @@ class Ecp5Packer
...
                 } else if (ci->type == id_DCCA) {
                     copy_constraint(ci, id_CLKI, id_CLKO, 1);
+                } else if (ci->type == id_DCSC) {
+                    if ((!ci->ports.count(id_CLK0) && !ci->ports.count(id_CLK1)) || !ci->ports.count(id_DCSOUT))
+                        continue;
+                    auto mode = str_or_default(ci->params, id_DCSMODE, "POS");
+
+                    // TODO: We can do this simplification if we know
+                    // MODESEL is tied to 0. I don't know how I can check this
+                    // if (mode == "CLK0_LOW" || mode == "CLK0_HIGH" || mode == "CLK0") {
```
rowanG077, line 2731: @gatecat 
Technically if `CLKx_LOW` then `low.max_delay` is infinity. SImilar case for `CLKx_HIGH` then `high.max_delay` is infinity. I felt it's not useful to model this here. But I'm happy to proven wrong.

ecp5/pack.cc
```diff
@@ -2706,6 +2721,86 @@ class Ecp5Packer
...
+                            );
+                        } else {
+                            derived_constr->low = DelayPair(
+                                std::min(c1->low.min_delay, c2->low.min_delay),
+                                std::max(c1->low.max_delay, c2->low.max_delay)
+                            );
+                        }
+
+                        if (mode == "POS") {
+                            derived_constr->high = DelayPair(
```
rowanG077, line 2780: @gatecat 

This maybe overconstrained too. At the transition point between clocks the worst case is that the clock will remain high or low for an additional period in the domain you are switching to. Since this only happens when transitioning it might be oke to remove it?

ecp5/pack.cc
gatecat: Yes, I don't see any case where this would matter.

ecp5/pack.cc
gatecat: No, I think this is a useful thing to have

ecp5/pack.cc
```diff
@@ -2706,6 +2721,86 @@ class Ecp5Packer
                     copy_constraint(ci, id_CLK1, id_ECSOUT, 1);
                 } else if (ci->type == id_DCCA) {
                     copy_constraint(ci, id_CLKI, id_CLKO, 1);
+                } else if (ci->type == id_DCSC) {
+                    if ((!ci->ports.count(id_CLK0) && !ci->ports.count(id_CLK1)) || !ci->ports.count(id_DCSOUT))
+                        continue;
+                    auto mode = str_or_default(ci->params, id_DCSMODE, "POS");
+
+                    // TODO: We can do this simplification if we know
+                    // MODESEL is tied to 0. I don't know how I can check this
```
gatecat, line 2730: to detect "tied to 0" or "tied to 1" at this stage, you can see if the connected net is called `$PACKER_GND_NET` or `"$PACKER_VCC_NET` accordingly.

rowanG077: Fixed

ecp5/pack.cc
```diff
@@ -36,6 +36,18 @@ static bool is_nextpnr_iob(Context *ctx, CellInfo *cell)
            cell->type == ctx->id("$nextpnr_iobuf");
 }
 
+static bool net_is_constant(const Context *ctx, NetInfo *net, bool &value)
+{
+    if (net == nullptr)
+        return false;
+    if (net->name == ctx->id("$PACKER_GND_NET") || net->name == ctx->id("$PACKER_VCC_NET")) {
```
gatecat, line 43: I think you can use `net->name.in(...)` as shorthand here

rowanG077: Done

rowanG077:
I didn't refactor the `is_constant_net` into the `base_ctx` since it doesn't seem that those names are the same for every arch. `fpga_interchange` uses `ctx->chip_info->constants->vcc_net_name` for example.

Issue #498, opened 2020-09-12
Feature: Option to Fail on Warning


rvalles:
When running nextpnr from a build script, I'd like to be able to possibly have my script stop if there's any warning.

This is currently not easy to do as even with warnings the returncode is zero.

Thus, a -Werror style option would be very useful (example: bash "set -e" would suffice).

PR #1092, opened 2023-02-06
common: Implement Werror flag

rowanG077:
Fixes https://github.com/YosysHQ/nextpnr/issues/498


rowanG077:
Definitely. The behavior is now as you suggest.

gatecat (approved):

PR #1112, opened 2023-02-23
CMake: detect platform support for threads

whitequark:
Together with https://github.com/YosysHQ/nextpnr/pull/1111 this change makes it possible to build nextpnr with or without parallel refinement support depending on whether the platform has threads available (in case of WASI, you can have WASI with or without threads).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,26 +42,25 @@ endif()
 
 set(Boost_NO_BOOST_CMAKE ON)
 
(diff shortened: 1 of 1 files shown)


whitequark:
The detection functionality seems to work. Nextpnr built with threads not so much, but it's a wasm-specific error (in fact I'm not sure if you can even cause it from C++ deliberately), so this should be good to merge.

PR #1113, opened 2023-02-23
WIP: Added SDC parser

rowanG077:
Still contains some logging. The parser is heavily based on the `PDC` parser. It's a really basic one with no support for wildcards and other stuff. For now it can be used to get nets.

In the future we really want something more sophisticated.

### DIFF
--- a/common/kernel/timing.cc
+++ b/common/kernel/timing.cc
@@ -30,7 +30,8 @@
 NEXTPNR_NAMESPACE_BEGIN
 
 namespace {
-std::string clock_event_name(const Context *ctx, ClockDomainKey &dom) {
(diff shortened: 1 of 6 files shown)


common/kernel/sdc.cc
```diff
@@ -0,0 +1,438 @@
...
+#include "log.h"
+#include "nextpnr.h"
+
+#include "sdc.h"
+#include <algorithm>
+#include <iterator>
+
+NEXTPNR_NAMESPACE_BEGIN
+
+struct TCLEntity
```
whitequark, line 39: The language's called "Tcl", not "TCL".

whitequark (changes requested):

Issue #1122, opened 2023-03-15
Wire looks shorted in GUI (?)


trayres:
Hi everyone, 
If I select W1 in my design on the GUI, it looks like it's connected to X60/Y49/C1
![image](https://user-images.githubusercontent.com/168285/225413463-e16b19e1-9667-4bc8-877d-154fc280d6b8.png)

If I move my mouse over slightly, it looks like the GUI is informing me that this is (in fact) connected to $PACKER_GND_NET.
![image](https://user-images.githubusercontent.com/168285/225414673-b7957e58-15f8-4858-8ee8-19419809b42f.png)

W1 is correctly shown on X60/Y49/A1_SLICE however:
![image](https://user-images.githubusercontent.com/168285/225414950-2e160ec5-c17f-4e5f-99ed-283fdb15e9d6.png)

How can I check to make sure W1 is not shorted to $PACKER_GND_SLICE (looking through the config files?) 
I have been doing that and I *think* there isn't actually a shorting problem, but I want to confirm (...obviously).

Thank you!

gatecat:
I've checked on a known good design and this is definitely a GUI issue rather than a design issue - I have also double checked that the `Context::check()` function run in nextpnr after every pass will never allow a short circuit in the design as it ensures each wire is only used by one net.

gatecat:
See https://github.com/YosysHQ/nextpnr/pull/1123

trayres:
This appears fixed, thank you!

PR #1123, opened 2023-03-16
ecp5: Update GUI rendering to match arch changes

gatecat:
- Render split SLICEs properly
 - Render LUT permutation properly with a pseudo-switchbox, instead of showing wires wrongly shorted as was noticed in #1122


trayres:
I'm not sure if this is related:
https://github.com/YosysHQ/nextpnr/issues/1133

PR #1125, opened 2023-03-17
cmake: Make HeAP placer always-enabled

gatecat:
Having the HeAP placer as optional didn't make sense when it's now considerably more stable than the SA placer, and I've never noticed a need for this option in the real world - Eigen3 is considerably lighter than other current dependencies for nextpnr.


Ravenslofty (approved):

Ravenslofty:
Maybe a bit more controversial of a take: we should also consider deprecating `sa` as a standalone placer.

PR #1128, opened 2023-03-20
ice40: Fix BRAM NegClk bitstream logic

gatecat:
The NegClk bits are read-bottom for everything other than the 1k density devices, based on some testing with icecube.

Fixes YosysHQ/Yosys#3707


niels-moller:
> The NegClk bits are read-bottom for everything other than the 1k density devices, based on some testing with icecube.

Thanks for investigating! What's a proper place to document this finding? Is icestorm/docs/ still the canonical place, or somewhere else? 
 
> Fixes [YosysHQ/yosys#3707](https://github.com/YosysHQ/yosys/issues/3707)

gatecat:
> Thanks for investigating! What's a proper place to document this finding? Is icestorm/docs/ still the canonical place, or somewhere else?

Yeah, I also put a PR in there at https://github.com/YosysHQ/icestorm/pull/313

PR #1129, opened 2023-03-22
gowin: Add support for OSER primitives

yrabbit:
* placement of OSER4, OVIDEO, OSER8 and OSER10 primitives is supported;
* primitives are implemented for the GW1N-1, GW1NZ-1, GW1NSR-4C, GW1NR-9, GW1NR-9C chips;
* the initial support for special HCLK clock wires is implemented to the extent necessary for OSER primitives to function;
* output to both regular IO and TLVDS_OBUF is supported;
* tricks required for IOLOGIC to work on one side of the -9 and -9C chips are taken into account;
* various edits, such as using idf() instead of the local buffer.

Compatible with old apicula bases.


gowin/arch.cc
```diff
@@ -2245,6 +2312,54 @@ void Arch::fix_pll_nets(Context *ctx)
...
+                }
+                if (user.port != id_FCLK) {
+                    continue;
+                }
+                WireId dstWire = ctx->getNetinfoSinkWire(net, user, 0);
+                if (ctx->verbose) {
+                    log_info("   Cell:%s, port:%s, wire:%s\n", user.cell->name.c_str(this), user.port.c_str(this),
+                             dstWire.c_str(this));
+                }
+                for (PipId const &pip : getPipsUphill(dstWire)) {
```
gatecat, line 2347: why the `const &`?

gowin/pack.cc
```diff
@@ -891,13 +895,142 @@ static void pack_iologic(Context *ctx)
...
+                    NPNR_ASSERT_FALSE("The pins for IDES/OSER must be specified explicitly.");
+                }
+
+                Loc loc = ctx->getBelLocation(ctx->getBelByNameStr(iob_bel->second.as_string()));
+                loc.z += BelZ::iologic_z;
+                ci->setAttr(id_BEL, ctx->getBelName(ctx->getBelByLocation(loc)).str(ctx));
+                BelId bel = ctx->getBelByLocation(loc);
+                if (bel == BelId()) {
+                    log_info("No bel for %s at %s\n", ctx->nameOf(ci), iob_bel->second.as_string().c_str());
+                    NPNR_ASSERT_FALSE("Can't place IDES/OSER here");
```
gatecat, line 935: this should probably be a nicer combined `log_error` error message rather than a `log_info` followed by assertion if it's something a user might trigger accidentally (like trying to use OSER at an IO not capable of it).

gowin/pack.cc
```diff
@@ -891,13 +895,142 @@ static void pack_iologic(Context *ctx)
...
+                    use_diff_io = true;
+                } else {
+                    ci->setAttr(id_OBUF_TYPE, std::string("SBUF"));
+                }
+
+                // disconnect Q output: it is wired internally
+                delete_nets.insert(ci->ports.at(output).net->name);
+                q0_dst->disconnectPort(id_I);
+                ci->disconnectPort(output);
+                bool have_XXX = ctx->bels[ctx->getBelByNameStr(iob_bel->second.as_string())].pins.count(id_XXX_1);
```
gatecat, line 966: can you use a nicer name for this internal pin than `XXX` ?

gatecat: this should probably also be `ctx->bels.at()`?

gowin/arch.cc
gatecat: why not just `PipId`? it's small enough that passing by value will be faster than passing by reference?

gowin/pack.cc
gatecat: something like `X_ALWAYS_HIGH` or `X_ALWAYS_LOW` ?

PR #1140, opened 2023-04-12
gowin: implement IDES16 and OSER16 primitives

yrabbit:
These are very cumbersome primitives that take up two cells and consequently 4 IOLOGIC bels.
The primitives are implemented for the chips that contain them and are supported by apicula GW1NSR-4C, GW1NR-9 and GW1NR-9C.


gowin/arch.cc
```diff
@@ -1745,6 +1809,92 @@ Arch::Arch(ArchArgs args) : args(args)
...
                 }
             } break;
+            case ID_OSER16: {
+                if (skip_aux_oser16(device, row, col)) {
+                    break;
+                }
+                belname = idf("R%dC%d_OSER16", row + 1, col + 1);
+                addBel(belname, id_OSER16, Loc(col, row, BelZ::oser16_z), false);
+
+                IdString const oser16_in_ports[] = {id_RESET, id_PCLK, id_D0,  id_D1,  id_D2,  id_D3,
```
gatecat, line 1819: I think the `const` here should be before `IdString` not after it

yrabbit: Sorry, I missed it. 
It's a habit, I'm just a stickler for east style :)
https://hackingcpp.com/cpp/design/east_vs_west_const.html

PR #1143, opened 2023-04-18
gowin: add a common mechanism for placing ports

yrabbit:
If the port is in a different cell than the primitive, then use the alias mechanism. 
Considerably compact code for OSC as an example.

### DIFF
--- a/gowin/arch.cc
+++ b/gowin/arch.cc
@@ -1273,7 +1273,7 @@ WireId Arch::get_make_port_wire(const DatabasePOD *db, const BelsPOD *bel, int r
         int srccol = alias_src->src_col;
         IdString srcid = IdString(alias_src->src_id);
         wire = wireToGlobal(srcrow, srccol, db, srcid);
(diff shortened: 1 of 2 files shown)


gowin/arch.cc
```diff
@@ -1258,6 +1258,28 @@ static bool skip_aux_oser16(std::string device, int row, int col)
     return false;
 }
 
+WireId Arch::get_make_port_wire(const DatabasePOD *db, const BelsPOD *bel, int row, int col, IdString port)
+{
+    IdString wirename = IdString(pairLookup(bel->ports.get(), bel->num_ports, port.hash())->src_id);
+    IdString wire = idf("R%dC%d_%s", row + 1, col + 1, wirename.c_str(this));
+    if (wires.count(wire) == 0) {
```
gatecat, line 1265: `if (!wires.count(wire))` is more idiomatic across the rest of the nextpnr codebase

yrabbit: done

PR #1146, opened 2023-05-02
MachXO3 support

mmicko:


machxo2/pack.cc
```diff
@@ -19,326 +18,1567 @@
...
+        new_cells.push_back(std::move(zero_cell));
     }
-}
 
-static bool is_nextpnr_iob(Context *ctx, CellInfo *cell)
-{
-    return cell->type == ctx->id("$nextpnr_ibuf") || cell->type == ctx->id("$nextpnr_obuf") ||
-           cell->type == ctx->id("$nextpnr_iobuf");
-}
+    int lookup_delay(const std::string &del_mode)
```
gatecat, line 1225: are you actually supporting the IO DELAYs? even if you are, have you checked these values actually make sense for machxo2?

machxo2/pack.cc
```diff
@@ -19,326 +18,1567 @@
...
+            tp = net->driver;
+            return true;
+        }
+        return false;
     }
 
-    for (auto pcell : packed_cells) {
-        ctx->cells.erase(pcell);
+    // Pass to pack LUT5s into a newly created slice
+    void pack_lut5xs()
```
gatecat, line 290: have you tested/used this at all?

machxo2/bitstream.cc
```diff
@@ -18,264 +18,741 @@
...
+    {
+        std::string tile = ctx->get_pip_tilename(pip);
+        std::string tile_type =
+                ctx->chip_info->tiletype_names[ctx->tile_info(pip)->pip_data[pip.index].tile_type].get();
+        std::string source = get_trellis_wirename(pip.location, ctx->getPipSrcWire(pip));
+        std::string sink = get_trellis_wirename(pip.location, ctx->getPipDstWire(pip));
+        cc.tiles[tile].add_arc(sink, source);
+
+        // Special case pips whose config bits are spread across tiles.
+        if (source == "G_PCLKCIBVIQT0" && sink == "G_VPRXCLKI0") {
```
gatecat, line 282: could there well be more cases of this?

mmicko: This is actually from old / existing code 
( https://github.com/YosysHQ/nextpnr/blob/master/machxo2/bitstream.cc#L132 ) so far it looks like it is this one that requires copy

machxo2/pack.cc
mmicko: Removed this for now, easy to restore, did not update yosys part anyway so it is not used for now.

PR #1149, opened 2023-05-11
Remove budget based PNR

rowanG077:
TODO:

- [x] Test on real world design to see if everything still functions as expected

### DIFF
--- a/common/kernel/timing.cc
+++ b/common/kernel/timing.cc
@@ -36,7 +36,7 @@ std::string clock_event_name(const Context *ctx, ClockDomainKey &dom) {
         value = "<async>";
     else
         value = (dom.edge == FALLING_EDGE ? "negedge " : "posedge ") + dom.clock.str(ctx);
-    
+
     return value;
 }
 } // namespace
@@ -501,7 +501,6 @@ void TimingAnalyser::reset_times()
             dp.second.hold_slack = std::numeric_limits<delay_t>::max();
             dp.second.max_path_length = 0;
             dp.second.criticality = 0;
-            dp.second.budget = 0;
         }
         port.second.worst_crit = 0;
         port.second.worst_setup_slack = std::numeric_limits<delay_t>::max();
@@ -1060,7 +1059,6 @@ struct Timing
                     if (portClass == TMG_ENDPOINT || portClass == TMG_IGNORE || portClass == TMG_CLOCK_INPUT) {
                         // Skip
                     } else {
-                        auto budget_override = ctx->getBudgetOverride(net, usr, net_delay);
                         // Iterate over all output ports on the same cell as the sink
                         for (auto port : usr.cell->ports) {
                             if (port.second.type != PORT_OUT || !port.second.net)
@@ -1073,12 +1071,8 @@ struct Timing
                             auto &data = net_data[port.second.net][start_clk];
                             auto &arrival = data.max_arrival;
                             arrival = std::max(arrival, usr_arrival + comb_delay.maxDelay());
-                            if (!budget_override) { // Do not increment path length if budget overridden since it
-                                                    // doesn't
-                                // require a share of the slack
-                                auto &path_length = data.max_path_length;
-                                path_length = std::max(path_length, net_length_plus_one);
-                            }
+                            auto &path_length = data.max_path_length;
+                            path_length = std::max(path_length, net_length_plus_one);
                         }
                     }
                 }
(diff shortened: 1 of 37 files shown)


common/place/place_common.cc
```diff
@@ -51,7 +51,7 @@ wirelen_t get_net_metric(const Context *ctx, const NetInfo *net, MetricType type
             continue;
         if (timing_driven) {
             delay_t net_delay = ctx->predictArcDelay(net, load);
-            auto slack = load.budget - net_delay;
+            auto slack = -net_delay;
```
rowanG077, line 54: @gatecat This is one of the only places I had to remove `budget` where I'm not sure I need to replace it with something else.

gatecat: Looks good - this is rarely if ever used code these days anyway (and quite possibly ripe for a cleanup)

test.v
```diff
@@ -0,0 +1,27 @@
+module foo
```
gatecat, line 1: this file looks accidentally committed?

gatecat:
Friendly ping, have you had another chance to look at this?

rowanG077:
Yes! I will update after the weekend.

rowanG077:
It now prints out the critical path using only the new API. I have compared the outputs of a real world on the ECP5. ~60% utilisation with 5 clocks and the reports match one to one. What is missing is that async paths aren't printed yet.

I want to separate out the code that prints critical paths since it's quite large and makes it harder to understand what code is just for the analysis and what part is just for printing. If you are oke with this I want to move all of that to a `timing_reporter.cc` file.

common/kernel/timing.cc
```diff
@@ -1377,115 +878,120 @@ CriticalPath build_critical_path_report(Context *ctx, ClockPair &clocks, const P
...
+        else
+            value = (e.edge == FALLING_EDGE ? std::string("negedge ") : std::string("posedge ")) + e.clock.str(ctx);
+        if (int(value.length()) < field_width)
+            value.insert(value.length(), field_width - int(value.length()), ' ');
+        return value;
+    };
+
+        // Print critical paths
     if (print_path) {
+        if (ctx->idstring_str_to_idx == nullptr) {
```
gatecat, line 986: hmm, where is this coming from?

shell.nix
```diff
@@ -0,0 +1,52 @@
+{ pkgs ? import <nixpkgs> {} }:
```
gatecat, line 1: I would prefer this is added in a separate PR (and you would agree to take on responsibility for maintaining it)

gatecat:
Some other comments:
 - has your testing covered designs with a critical path from posedge to negedge or vice versa?
 -  there's a CI failure as we have -Wall there: `error: ‘const char* nextpnr_generic::{anonymous}::edge_name(nextpnr_generic::ClockEdge)’ defined but not used [-Werror=unused-function]
   33 | const char *edge_name(ClockEdge edge) { return (edge == FALLING_EDGE) ? "negedge" : "posedge"; }`
 - there's another CI failure because himbaechel needs `getBudgetOverride` removed now too
 - it looks like this might remove the slack histogram if I've gone through the diff correctly, it would be nice to have that back

shell.nix
rowanG077: That's fine!

common/kernel/timing.cc
rowanG077: It was some debug code. There are more debug prints I will remove once I have validated everything to be equivalent.

rowanG077:
> - has your testing covered designs with a critical path from posedge to negedge or vice versa?

Good point will test this out. Do you have a "real world" design that uses this? Of course I can write smaller test cases myself but I want to ensure that it reports the exact same paths even if there are lot's of paths. 

>  - there's a CI failure as we have -Wall there: error: ‘const char* nextpnr_generic::{anonymous}::edge_name(nextpnr_generic::ClockEdge)’ defined but not used [-Werror=unused-function]    33 | const char *edge_name(ClockEdge edge) { return (edge == FALLING_EDGE) ? "negedge" : "posedge"; }
> - there's another CI failure because himbaechel needs getBudgetOverride removed now too

Right this is still WIP. Code style is also a mess still

> it looks like this might remove the slack histogram if I've gone through the diff correctly, it would be nice to have that back

Correct! I need to recover slack histogram and the full critical nets.

[ ] Implement slack histogram and critical nets
[ ] Extract timing reporting into `timing_report.cc` OR refactor `report.cc` @gatecat. Let me know what you prefer.

kittennbfive:
Hi,

will this PR change the critical timing report that nextpnr is providing? (I am asking because i was working on a "beautify"-script for the report, so if the format will change it's pointless for me to continue my work.)

gatecat:
The plan is for no user visible changes to the report format.

kittennbfive:
Thank you for this really quick answer!

gatecat (approved):
LGTM

PR #1152, opened 2023-05-16
 rename c++20 keyword s/requires/requires_range.

nqbit:
See https://en.cppreference.com/w/cpp/language/requires for more details.


gatecat:
Thanks for spotting this one!

PR #1155, opened 2023-05-16
mistral: improve timing calculation

Ravenslofty:
There are a number of timing-related goodies in here:
- the Mistral analogue simulator is now used for (admittedly-slow) timing signoff; it usually finds a few megahertz.
- routing timings have been redone; the old timings turned out to be about 11% more pessimistic than necessary.
- `predictDelay` and `estimateDelay` have been fitted to a sample design; this also helps Fmax a bit.
- HeAP has a parameter tweaked to account for the asymmetry of the Cyclone V.

Is this production-grade? Not quite, but I'd like to believe we're getting a bit closer.

### DIFF
--- a/mistral/bitstream.cc
+++ b/mistral/bitstream.cc
@@ -20,6 +20,7 @@
 #include "log.h"
 #include "nextpnr.h"
 #include "util.h"
+#include "timing.h"
 
 NEXTPNR_NAMESPACE_BEGIN
 namespace {
@@ -379,6 +380,7 @@ struct MistralBitgen
         write_routing();
         write_cells();
         write_labs();
+        ctx->bitstream_configured = true;
     }
 };
 } // namespace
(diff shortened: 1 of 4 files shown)


mistral/bitstream.cc
```diff
@@ -387,6 +389,13 @@ void Arch::build_bitstream()
 {
     MistralBitgen gen(getCtx());
     gen.run();
+    // This is a hack to run timing analysis yet again after the bitstream is
+    // configured in Mistral, because the analogue simulator won't work until
+    // it has a bitstream in the library.
+    //
+    // A better solution would be to move a lot of this bitstream code to
+    // {un,}bind{Bel, Pip} and friends, but we're not there yet.
+    timing_analysis(getCtx(), true, true, true, true, true);
```
gatecat, line 398: Can you add a `log_info` statement here along the lines of "Running signoff timing analysis...", to explain what it's doing and the reason for getting a second Fmax at the end?

Issue #1158, opened 2023-05-17
Segfault when clocking a FF from a ring oscillator


meiniKi:
Clocking FFs from a ring oscillator leads to a segmentation fault due to an infinite recursion of `find_net_drivers`.
https://github.com/YosysHQ/nextpnr/blob/7a827b1c787edb2362fffea8f072a95e331fe976/common/kernel/timing.cc#L299

1. Is there a better way to spot infinite recursions due to circle or lasso-shaped driver networks from a ring oscillator than maintaining a trace, as I did here https://github.com/meiniKi/nextpnr/commit/e69d7ebd33cac8d3599007ab103f3628ebae0b0f ?
2. The ring oscillator may be tapped at different stages; can this lead to side effects or failed timing checks?

gatecat:
1. That commit looks good - do you want to submit it as a PR?

2. As this is only used to find simply related clocks (e.g. due to clock gating primitives), I don't expect any timing check problems here. Note that nextpnr won't automatically try and determine frequency or phase for ring oscillators, they would have to be constrained manually anyway.

meiniKi:
Great. Thanks for checking @gatecat .

PR #1160, opened 2023-05-19
Fix segfault when clocking a FF from a ring oscillator

meiniKi:
Fixes https://github.com/YosysHQ/nextpnr/issues/1158

### DIFF
--- a/common/kernel/timing.cc
+++ b/common/kernel/timing.cc
@@ -296,8 +296,8 @@ void TimingAnalyser::identify_related_domains()
 
     // For each clock net identify all nets that can possibly drive it. Compute
     // cumulative delays to each of them.
-    std::function<void(const NetInfo *, std::set<IdString> &, dict<IdString, delay_t> &, delay_t)> find_net_drivers =
-            [&](const NetInfo *ni, std::set<IdString> &net_trace, dict<IdString, delay_t> &drivers, delay_t delay_acc) {
+    std::function<void(const NetInfo *, pool<IdString> &, dict<IdString, delay_t> &, delay_t)> find_net_drivers =
+            [&](const NetInfo *ni, pool<IdString> &net_trace, dict<IdString, delay_t> &drivers, delay_t delay_acc) {
                 // Get driving cell and port
                 if (ni == nullptr)
                     return;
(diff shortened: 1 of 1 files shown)


common/kernel/timing.cc
```diff
@@ -296,8 +296,8 @@ void TimingAnalyser::identify_related_domains()
 
     // For each clock net identify all nets that can possibly drive it. Compute
     // cumulative delays to each of them.
-    std::function<void(const NetInfo *, dict<IdString, delay_t> &, delay_t)> find_net_drivers =
-            [&](const NetInfo *ni, dict<IdString, delay_t> &drivers, delay_t delay_acc) {
+    std::function<void(const NetInfo *, std::set<IdString> &, dict<IdString, delay_t> &, delay_t)> find_net_drivers =
```
Ravenslofty, line 299:
```suggestion
    std::function<void(const NetInfo *, pool<IdString> &, dict<IdString, delay_t> &, delay_t)> find_net_drivers =
```

common/kernel/timing.cc
```diff
@@ -296,8 +296,8 @@ void TimingAnalyser::identify_related_domains()
...
+            [&](const NetInfo *ni, std::set<IdString> &net_trace, dict<IdString, delay_t> &drivers, delay_t delay_acc) {
```
Ravenslofty, line 300:
```suggestion
            [&](const NetInfo *ni, pool<IdString> &net_trace, dict<IdString, delay_t> &drivers, delay_t delay_acc) {
```

common/kernel/timing.cc
```diff
@@ -371,7 +378,8 @@ void TimingAnalyser::identify_related_domains()
             continue;
 
         dict<IdString, delay_t> drivers;
-        find_net_drivers(ni, drivers, 0);
+        std::set<IdString> net_trace;
```
Ravenslofty, line 381:
```suggestion
        pool<IdString> net_trace;
```

Ravenslofty (commented):
nextpnr tends to avoid `std::set` because it can introduce nondeterminism across platforms. Does it matter in this case? Perhaps not, but better safe than sorry.

meiniKi:
> nextpnr tends to avoid `std::set` because it can introduce nondeterminism across platforms. Does it matter in this case? Perhaps not, but better safe than sorry.

Thanks for the hint.

PR #1162, opened 2023-05-24
router2 alternative weights

Ravenslofty:
I've been tinkering with r2 over the past few days, and I think I have a set of weights that reaches a much higher maximum frequency than the defaults, using `mistral`, anyway.

The idea is to have more, smaller-step iterations, in the hopes of better negotiation between wires, and at least empirically, this seems to work out.

### DIFF
--- a/common/kernel/command.cc
+++ b/common/kernel/command.cc
@@ -211,6 +211,8 @@ po::options_description CommandHandler::getGeneralOptions()
     general.add_options()("router2-tmg-ripup",
                           "enable experimental timing-driven ripup in router (deprecated; use --tmg-ripup instead)");
 
+    general.add_options()("router2-alt-weights", "use alternate router2 weights");
+
     general.add_options()("report", po::value<std::string>(),
                           "write timing and utilization report in JSON format to file");
     general.add_options()("detailed-timing-report", "Append detailed net timing data to the JSON report");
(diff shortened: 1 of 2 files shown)


rowanG077:
That's pretty neat! Do you think there would be worth in using an optimizer to find (even) more optimal values? Maybe even per arch? I would have no problem in throwing a month or two of compute at it with a set of real world designs.

rowanG077:
WOW that's a pretty insane improvement. 25% higher fmax is nothing to sneeze at. 

I would have hoped that you could still do it by simply throwing enough compute at it. In contrast to chess we don't really have a real-time requirement here. So essentially having a sub-optimal optimizer might still be worth it in terms of cost/benefit.

Ravenslofty:
Mmm, it's interesting to see how our approaches are very different.

In computer chess, we have tried:
- a bunch of manual tuning using [SPRT](https://en.wikipedia.org/wiki/Sequential_probability_ratio_test) comparisons. 
- [CLOP](https://www.remi-coulom.fr/CLOP/), which relies on very little information ("between these two parameter configs, is one better, equal, or worse than the other?").
- but instead one can use comparison tests of `param - X` against `param + X` which gives one a very noisy graident approximation, so instead one can use [SPSA](https://en.wikipedia.org/wiki/Simultaneous_perturbation_stochastic_approximation)

rowanG077:
I'm no expert by any means. I have used optimization processes like this in the context of computer vision and object detection in pointclouds.  The advantage there is that the gradients are smooth, I have no clue whether that's the case for weights in router2.

Ravenslofty:
I think if one can use SPSA to differentiate games of chess, then router2 shouldn't be too much of an obstacle.

PR #1170, opened 2023-06-05
mistral: improve ALM packing in LABs

Ravenslofty:
The current ALM packing code overestimates LAB TD congestion due to not accounting for sharing across ALMs. 

Recently, I tried to resolve this by hashing the signal names, but this actually underestimated LAB TD congestion, because the TD-to-GOUT (ALM input) crossbar is not fully populated. The solution to this is to hash signal-and-ALM-input pairs, which still results in a distinct reduction in LAB use for attosoc, from 133 to 123 LABs.

I am a bit concerned about the impact on performance here; while it's still "fine" for small designs, perhaps @gatecat might have an alternative implementation based on the existing `unique_input_count` code minus some incrementally-updated factor for signal sharing.

This code also does not take into account LD usage, which would need some further thinking.

### DIFF
--- a/mistral/lab.cc
+++ b/mistral/lab.cc
@@ -534,6 +534,15 @@ void Arch::update_alm_input_count(uint32_t lab, uint8_t alm)
     alm_data.unique_input_count = total_inputs;
 }
 
+const std::array<int, 6> tdmux_set = {
+    0, // A
+    1, // B
+    0, // C
+    1, // D
+    2, // E
+    3, // F
+};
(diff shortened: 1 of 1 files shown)


Ravenslofty:
I figured you'd have a better way of doing this, but I'm glad the code got the general idea across.

Lunaphied:
Closing as stale - was in draft since 2023, @gatecat suggested incremental approach but no progress in 3 years.

PR #1177, opened 2023-06-12
router2: revisit nodes with lower cost

Ravenslofty:
This seems like one of the many things that gatecat wanted to get around to doing on router2, but didn't have the time. It turns out to be a surprisingly small patch, but it's still a reasonable Fmax boost for nextpnr-mistral.

### DIFF
--- a/common/route/router2.cc
+++ b/common/route/router2.cc
@@ -94,6 +94,7 @@ struct Router2
         // Visit data
         PipId pip_fwd, pip_bwd;
         bool visited_fwd = false, visited_bwd = false;
+        float cost_fwd = 0.0, cost_bwd = 0.0;
     };
 
     Context *ctx;
@@ -522,6 +523,8 @@ struct Router2
             flat_wires[w].pip_bwd = PipId();
             flat_wires[w].visited_fwd = false;
             flat_wires[w].visited_bwd = false;
+            flat_wires[w].cost_fwd = 0.0;
+            flat_wires[w].cost_bwd = 0.0;
         }
         t.dirty_wires.clear();
     }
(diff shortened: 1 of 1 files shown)


common/route/router2.cc
```diff
@@ -559,25 +563,27 @@ struct Router2
...
         if (!wd.visited_fwd && !wd.visited_bwd)
             t.dirty_wires.push_back(wire);
         wd.pip_bwd = pip;
         wd.visited_bwd = true;
+        wd.cost_bwd = cost;
     }
 
-    bool was_visited_fwd(int wire) { return flat_wires.at(wire).visited_fwd; }
-    bool was_visited_bwd(int wire) { return flat_wires.at(wire).visited_bwd; }
+    bool was_visited_fwd(int wire, float cost) { return flat_wires.at(wire).visited_fwd && flat_wires.at(wire).cost_fwd <= cost; }
```
gatecat, line 585: at some point it might be worth looking at adding an epsilon like router1 to avoid minimally-useful revisits, but that's probably a future exercise

gatecat:
Thanks for doing this! Happy to merge if you think it's ready there.

At some point we should also revisit the terminating condition for the arc router loop (also in the other direction):
https://github.com/YosysHQ/nextpnr/blob/863ad4cee865a0b66c66b7204216c9879f5673e9/common/route/router2.cc#L704-L707

We would set an estimated iteration count heuristic or similar (probably, in first iterations at least for lower-criticality-nets, the same as before, terminating as soon as any route is find) that enables it to keep searching for new midpoints with a lower total delay for a few more iterations in some cases.

But that's not needed to be part of this PR imo.

PR #1179, opened 2023-06-14
router2: fix 8935c186

Ravenslofty:


whitequark:
Thanks!

whitequark:
Looks like the fix doesn't work: https://github.com/YoWASP/nextpnr/actions/runs/5273760173/jobs/9537541164

Ravenslofty:
Then I'm not sure what to do.

gatecat:
You could just use the value via C++ numeric_limits (as generally used elsewhere in nextpnr) instead, perhaps with a helper constant somewhere.

gatecat:
@Ravenslofty ping?

Issue #1173, opened 2023-06-09
gowin: incorrect implementation for LUTs with constant 1'b0 inputs


meiniKi:
Implementing non-optimized LUTs that have constant `1'b0` inputs causes incorrect results. This only happens when instantiating a LUT manually, as in

```Verilog
(* keep *) LUT2 #(
  .INIT (4'b0001)
) i_inverter (
  .I0 ( a ),
  .I1 ( 1'b0 ),
  .F  ( c ) 
);
```
because Yosys optimizes to LUT1/LUT2/LUT3 when inferring logic. 

Using the LUT2 instantiation from above, the packer receives the configuration "0001" and replicates it to obtain a 16-bit configuration string [here](https://github.com/YosysHQ/apicula/blob/d350cb7cee68144fdef159e8104bb7652ed91fdc/apycula/gowin_pack.py#L591). Thus, the configuration for the LUT is correct, independent of the valuation of I2 and I3.

When looking at the JSON output of nextpnr-gowin, GND is assumed to be the default value.

```json
          "connections": {
            "LSR": [  ],
            "CE": [  ],
            "Q": [  ],
            "F": [ 3877689 ],
            "CLK": [  ],
            "D": [  ],
            "C": [  ],
            "B": [  ],
            "A": [ 3877684 ]
          }
```

```json
        "$PACKER_GND_NET": {
          "hide_name": 1,
          "bits": [ 3877702 ] ,
          "attributes": {
            "ROUTING": " "
          }
        }
```

LUTs in gowin FPGAs, however, seem to have `1'b1` as the default input value. Changing the `INIT` parameter to `4'b0100` implements a correct design, although `I1` is set to `1'b0`. This happens only when inputs are hard-wired to `1'b0`. LUT1/LUT2/LUT3 are correct due to the config replication (same behavior for any valuation).

If my observation is correct, I propose this fix 7997f2f40f7b8624850bdece23453b5041a4ad7a.

gatecat:
@yrabbit can you have a look at this for me?

yrabbit:
> @yrabbit can you have a look at this for me?
Very interesting! I will try to give my thoughts today.

yrabbit:
@gatecat 
A completely unused and unprogrammed LUT, that is, by default, is logically INIT='16'b1111 1111 1111 1111'. So the output F of the empty, unused, unprogrammed LUT will always be 1 regardless of the state of the inputs.

[...]

PR #1180, opened 2023-06-15
gowin: fixes default networks

meiniKi:
Proposes a fix for #1173 

Instead of routing constant `1'b0` and `1'b1` nets to LUTs, the LUT configuration is modified such that any value at that input those pin leads to the same output.

I removed `is_lc(ctx, uc)` from the `if` statement, so all nets are routed explicitly in this case. Should we handle that case differently?

### DIFF
--- a/gowin/pack.cc
+++ b/gowin/pack.cc
@@ -636,7 +636,45 @@ static void set_net_constant(const Context *ctx, NetInfo *orig, NetInfo *constne
             CellInfo *uc = user.cell;
             if (ctx->verbose)
                 log_info("%s user %s\n", ctx->nameOf(orig), ctx->nameOf(uc));
-            if ((is_lut(ctx, uc) || is_lc(ctx, uc)) && (user.port.str(ctx).at(0) == 'I') && !constval) {
+
+            if ((is_lut(ctx, uc)) && (user.port.str(ctx).at(0) == 'I') && !constval) {
+                auto it_param = uc->params.find(id_INIT);
+                if (it_param == uc->params.end())
+                    log_error("No initialization for lut found.\n");
+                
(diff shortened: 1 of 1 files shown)


gowin/pack.cc
```diff
@@ -636,7 +636,45 @@ static void set_net_constant(const Context *ctx, NetInfo *orig, NetInfo *constne
             CellInfo *uc = user.cell;
             if (ctx->verbose)
                 log_info("%s user %s\n", ctx->nameOf(orig), ctx->nameOf(uc));
-            if ((is_lut(ctx, uc) || is_lc(ctx, uc)) && (user.port.str(ctx).at(0) == 'I') && !constval) {
+
+            if ((is_lut(ctx, uc)) && (user.port.str(ctx).at(0) == 'I') && !constval) {
```
gatecat, line 640: I think the ` && !constval` term isn't actually needed here, as the same logic can be applied when an input is tied high as well as low

meiniKi: Right, should I leave it in the function signature to be consistent with the other architectures although `constval` is not used?

gatecat: No, packing code is entirely arch-specific so there's no need to worry there

yrabbit:
> I removed `is_lc(ctx, uc)` from the `if` statement, so all nets are routed explicitly in this case. Should we handle that case differently?

And this is a very reasonable solution because SLICE type cells are formed in particular as a result of ALU implementation, and there this one LUT is cut into two unequal parts and they are used as independent LUTs. So interfering with their tables would be disastrous :)

yrabbit:
@gatecat A practical check on examples from apicula showed no regressions.

PR #1182, opened 2023-06-19
router2: fix 8935c186 (again)

Ravenslofty:
Hopefully for the last time?

(Perhaps nextpnr needs a webassembly CI job or something.)

cc @whitequark @gatecat

### DIFF
--- a/common/route/router2.cc
+++ b/common/route/router2.cc
@@ -31,9 +31,9 @@
 #include <algorithm>
 #include <boost/container/flat_map.hpp>
 #include <chrono>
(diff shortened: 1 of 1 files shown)


whitequark:
> (Perhaps nextpnr needs a webassembly CI job or something.)

That would be ideal. I think it would be easy enough to pilfer the build scripts from YoWASP? I don't have time to do this right now (very busy shipping the next Amaranth version) but I'm happy to help out anyone taking it on, as I've stepped on probably just about every rake this cross-build has to offer.

PR #1184, opened 2023-07-01
gowin: Himbaechel arch

yrabbit:
gowin P&R implementation as Himbaechel sub-architecture.


himbaechel/uarch/gowin/gowin.cc
```diff
@@ -7,230 +7,234 @@
...
-	ctx->cells.at(ctx->id("leds_OBUF_O_1"))->setAttr(ctx->id("BEL"), std::string("X0Y15/IOBB"));
-	ctx->cells.at(ctx->id("leds_OBUF_O_2"))->setAttr(ctx->id("BEL"), std::string("X0Y20/IOBB"));
-	ctx->cells.at(ctx->id("leds_OBUF_O_3"))->setAttr(ctx->id("BEL"), std::string("X0Y21/IOBB"));
-	ctx->cells.at(ctx->id("leds_OBUF_O_4"))->setAttr(ctx->id("BEL"), std::string("X0Y24/IOBB"));
-	ctx->cells.at(ctx->id("leds_OBUF_O_5"))->setAttr(ctx->id("BEL"), std::string("X0Y25/IOBB"));
-	ctx->cells.at(ctx->id("rst_IBUF_I"))->setAttr(ctx->id("BEL"), std::string("X0Y4/IOBA"));
-	assign_cell_info();
+void GowinImpl::prePlace()
+{
+    ctx->cells.at(ctx->id("leds_OBUF_O"))->setAttr(ctx->id("BEL"), std::string("X0Y14/IOBA"));
```
gatecat, line 24: I know this is test code, but testing with multiple designs would probably be easier manually instantiating the IO buffers and then setting the bel attribute in the Verilog

himbaechel/uarch/gowin/gowin.cc
```diff
@@ -7,230 +7,234 @@
...
+    const dict<IdString, Property> vcc_params;
+    const dict<IdString, Property> gnd_params;
+    h.replace_constants(CellTypePort(id_GOWIN_VCC, id_V), CellTypePort(id_GOWIN_GND, id_G), vcc_params, gnd_params);
+
+    // disconnect the constant LUT inputs
+    mod_lut_inputs();
+
+    // Constrain directly connected LUTs and FFs together to use dedicated resources
+    int lutffs = h.constrain_cell_pairs(
+            pool<CellTypePort>{{id_LUT1, id_F}, {id_LUT2, id_F}, {id_LUT3, id_F}, {id_LUT4, id_F}},
```
gatecat, line 52: As a style point, probably make these `pool<CellTypePort>` variables rather than having them inline as long initialiser lists in the function parameters.

gatecat:
Some tiny comments, but overall I'm really happy with the direction this is going in!

himbaechel/uarch/gowin/gowin.cc
yrabbit: I know :) will be fixed

yrabbit:
> Some tiny comments, but overall I'm really happy with the direction this is going in!

I'll need to get to pins, banks, and packages soon.

And would really like to know the name of the loaded chipdb sometime

gatecat:
> And would really like to know the name of the loaded chipdb sometime

`chip_info->name.get()` should work

yrabbit:
Ok. I missed that.

gatecat:
I would suggest planning to split `gowin.cc` into multiple files eventually, in particular a `pack.h`/`pack.cc` that contain the packing-related steps. The structural approach of the packer helper class taken in the FABulous Viaduct arch: https://github.com/YosysHQ/nextpnr/blob/master/generic/viaduct/fabulous/pack.cc should be applicable here too.

himbaechel/uarch/gowin/gowin.cc
```diff
@@ -0,0 +1,400 @@
...
+        }
+        if (alu) {
+            const auto &alu_data = fast_cell_info.at(alu->flat_index);
+            if (ff_data.ff_d != alu_data.alu_sum) {
+                return false;
+            }
+        }
+        if (adj_ff) {
+            // DFFs must be same type or compatible
+            if (ff->type != adj_ff->type && ((ff->type.in(id_DFFS) && !adj_ff->type.in(id_DFFR)) ||
```
gatecat, line 363: maybe it would be neater to put this in a `ffs_compatible` inline function or similar?

yrabbit: > maybe it would be neater to put this in a `ffs_compatible` inline function or similar?
This is really a one-time piece ... well, I'll move it tomorrow.

yrabbit: > maybe it would be neater to put this in a `ffs_compatible` inline function or similar?

Done

himbaechel/uarch/gowin/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
+message(STATUS "Configuring Himbaechel-Gowin uarch")
+cmake_minimum_required(VERSION 3.5)
+project(himbaechel-gowin-chipdb NONE)
+
+find_package(Python3 3.5 REQUIRED COMPONENTS Interpreter)
+set(ALL_HIMBAECHEL_GOWIN_DEVICES GW1N-1 GW1NZ-1 GW1NS-2 GW1N-4 GW1N-9 GW1N-9C GW1NS-4 GW2A-18 GW2A-18C)
+set(HIMBAECHEL_GOWIN_DEVICES ${ALL_HIMBAECHEL_GOWIN_DEVICES} CACHE STRING
```
gatecat, line 7: can we have this default to none for now so building nextpnr himbaechel doesn't require the gowin binary?

I still want to work a bit more on the user experience for building nextpnr-himbaechel and device databases, then maybe we can revisit this when nextpnr-himbaechel has replaced the old nextpnr-gowin.

himbaechel/uarch/gowin/pack.cc
```diff
@@ -0,0 +1,1373 @@
...
+    {
+        if (ci.type.in(id_ODDR, id_ODDRC, id_OSER4, id_IDDR, id_IDDRC, id_IDES4)) {
+            return nullptr;
+        }
+        IdString aux_name = create_aux_iologic_name(ci.name, idx);
+        BelId bel = get_aux_iologic_bel(ci);
+        BelId io_bel = gwu.get_io_bel_from_iologic(bel);
+        if (!ctx->checkBelAvail(io_bel)) {
+            if (!is_diff_io(io_bel)) {
+                log_error("Can't place %s at %s because of %s\n", ctx->nameOf(&ci), ctx->nameOfBel(bel),
```
gatecat, line 486: can you add a short reason to this error message, something like `because of a conflict with %s` rather than just `because of %s`?

himbaechel/uarch/gowin/pack.cc
```diff
@@ -0,0 +1,1373 @@
...
+            ci->renamePort(id_Q1, id_Q9);
+            ci->renamePort(id_Q0, id_Q8);
+            break;
+        case ID_IDES4:
+            ci->renamePort(id_Q3, id_Q9);
+            ci->renamePort(id_Q2, id_Q8);
+            ci->renamePort(id_Q1, id_Q7);
+            ci->renamePort(id_Q0, id_Q6);
+            break;
+        case ID_IVIDEO:
```
gatecat, line 520: nitpick: I think `for` loops might be tidier here?

himbaechel/uarch/gowin/pack.cc
```diff
@@ -0,0 +1,1373 @@
...
+                        NPNR_ASSERT(ci->cluster == ClusterId());
+                        ci->cluster = cin_block_ci->name;
+                        ci->constr_abs_z = false;
+                        ci->constr_x = alu_chain_len / 6;
+                        ci->constr_y = 0;
+                        ci->constr_z = alu_chain_len % 6;
+                        // XXX I2 is pin C which must be set to 1 for all ALU modes except MUL
+                        // we use only mode 2 ADDSUB so create and connect this pin
+                        ci->addInput(id_I2);
+                        ci->connectPort(id_I2, ctx->nets[ctx->id("$PACKER_VCC")].get());
```
gatecat, line 1114: nitpick: ` ctx->nets.at(ctx->id("$PACKER_VCC")).get()` rather than `[]` - this way it will fail if for some reason the net doesn't exist rather than silently creating a `nullptr` net and then failing in a harder to debug way

himbaechel/uarch/gowin/gowin_arch_gen.py
```diff
@@ -0,0 +1,701 @@
...
+                # gobal carry chain
+                if x > 1 and chip.tile_type_at(x - 1, y).extra_data.tile_class == chip.strs.id('LOGIC'):
+                    nodes.append([NodeWire(x, y, f'CIN0'),
+                                  NodeWire(x - 1, y, f'COUT5')])
+
+            for node in nodes:
+                chip.add_node(node)
+
+            # VCC and VSS sources in the all tiles
+            global_nodes.setdefault('GND', []).append(NodeWire(x, y, 'VSS'))
```
gatecat, line 195: as a heads up to myself - having a node that fans out to all tiles will cause problems with current deduplication. you don't need to change anything right now, but beware the database will be bigger than expected for this reason (the same is possibly true for global nodes, I will need to see how well the apicula structure compacts)

gatecat (commented):
This is looking very good! Most of the comments are really small changes. I'm happy to merge them straight away, then I think there are a few changes on my side next:

  - timing support in himbächel
  - better user experience for building and finding chipdbs at runtime (similar to how the yosys `share` directory finding approach works)
  - improved ux for deduplicated global constants

himbaechel/uarch/gowin/gowin_arch_gen.py
yrabbit: Fixed.

yrabbit: Base sizes are still acceptable :)

bases

gatecat: That's really nice - I suspect this is probably only a big problem when talking about the millions of LUTs then ^^

Will put this as more of a slightly background thing compared to the other TODOs on my side.

yrabbit:
Apicula release 0.9.0a1 with Himbaechel support is in place, what about merge this one?
What did you decide with HIMBAECHEL_GOWIN_DEVICES?

gatecat:
Yes let's fly with this for now. The way in which nextpnr-himbaechel is started (in particular how chipdb paths work) will probably change in a week or two, so just bear that in mind.

himbaechel/uarch/gowin/CMakeLists.txt
whitequark: Could you tell me more about the build process of nextpnr-himbaechel and device databases? This is something I care about in YoWASP. It is not feasible to ship one binary that includes every binary and every chipdb due to size constraints (about 100 MB) and I would rather avoid configuring and building nextpnr several times if at all feasible.

gatecat: The idea is to provide a unified framework for arches to share one nextpnr binary with a reduced amount of per-arch code, and with a relatively compact deduplicated representation. Gowin will move there first and I will also look at xo2/3 moving to it. Longer term if it works out there will be more migrations and new arches using it.

It is going to have the chipdb binaries as data files (which should in every case be significantly less than 100MB). These will be configurable in the build process and in the next few days there should be a nice discovery process for those so you can pass the name of a device on the command line and it will load the right subarch and chipdb automatically (it'll be a relative path based approach similar to Yosys, I need to look exactly what we do for different arches and build environments there still).

I would guess, once it reaches a point of being end user ready (i.e. the legacy nextpnr-gowin is formally deprecated and then removed in favour of this), we'd have one nextpnr-himbaechel package and then another containing the chip databases per arch.

whitequark: Got it. Can you make sure every architecture (~ user-visible package) gets its own directory under `share/` and nothing outside of this directory is instaled? Then I will be able to ship this nicely in YoWASP.

gatecat: Yeah, that matches what I was planning

whitequark: Fantastic! Then what I will do in `yowasp-nextpnr-himbaechel` is to pull in all of the chipdbs that `yowasp-nextpnr-${ARCH}` have registered using the standard Python "entry points" (plugin) mechanism and mount them under `share/`. Ideally that in itself will be enough.

Do you plan to have e.g. a `nextpnr-ecp5` runner that proxies to `nextpnr-himbaechel --uarch ecp5`? This could be done by snooping `argv[0]` (checking if it ends with `-ecp5`, etc) and it would greatly simplify migration for Amaranth.

gatecat: Yes, this could definitely be done as part of the migration path. Although I realise a bit more work around the himbaechel startup and API is needed to support arbitrary command line arguments, which is needed both for backwards compatibility and a friendly CLI in general.

PR #1198, opened 2023-08-28
Widelut support for Mach XO2/XO3/XO3D

mmicko:
code is 100% same as in ECP5


mmicko:
Sure, will make it disabled by default on synth_lattice (for xo2/xo3/xo3d) and add option to enable it

PR #1199, opened 2023-08-29
VLO,VHI support for ECP5

mmicko:


mmicko:
VLO and VHI will be used by synth_lattice only if explicitly instantiated, so by default it is still all as before

PR #1202, opened 2023-08-31
gowin: Himbaechel. Fix the device selection

yrabbit:
Slightly change the Gowin device selection mechanism for database generation. 
By default, nothing is generated as before.


yrabbit:
Thanks!

PR #1213, opened 2023-09-12
himbaechel: Fix for Python 3.9

gatecat:
@yrabbit can you confirm this is enough to make it work for you?


yrabbit:
check in three hours

yrabbit:
oh, there are still problems with my computer - unfortunately I won’t be able to check it until the weekend;(

gatecat:
It seems to run fine with a local install of Python 3.9; still, let me know if you hit any further problems.

yrabbit:
The apicula examples (which build the nextpnr master branch) have compiled. There is a very high probability that everything will be fine.

PR #1217, opened 2023-09-16
gowin: Himbaechel. Support DragonFlyBSD

yrabbit:
We add support right here so that later I don’t have to make patches to the ports.


gatecat:
Does Yosys need a similar PR, as the code should be the same?

yrabbit:
> Does Yosys need a similar PR, as the code should be the same?

I've almost never made changes to Yosys and can't immediately switch to that code yet. Perhaps if no one does this before, I'll read what's going on there.
For now I'm editing the code for selecting the base in himbaechel-gowin :)

PR #1218, opened 2023-09-16
gowin: Himbaechel. Specify the chip variant.

yrabbit:
For GW2A-18 and GW1N-9 you need to specify the family in addition to partno.

### DIFF
--- a/himbaechel/uarch/gowin/gowin.cc
+++ b/himbaechel/uarch/gowin/gowin.cc
@@ -81,9 +81,9 @@ void GowinImpl::init_database(Arch *arch)
     if (args.options.count("family")) {
         family = args.options.at("family");
     } else {
-        bool GW2 = args.device == "GW2A-LV18PG256C8/I7";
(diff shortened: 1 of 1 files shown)


himbaechel/uarch/gowin/gowin.cc
```diff
@@ -81,16 +81,19 @@ void GowinImpl::init_database(Arch *arch)
     if (args.options.count("family")) {
         family = args.options.at("family");
     } else {
-        bool GW2 = args.device == "GW2A-LV18PG256C8/I7";
+        bool GW2 = args.device.rfind("GW2A", 0) == 0;
```
gatecat, line 84: nitpick - why `rfind` and not `find`?

yrabbit: find() with pos parameter only looks in that position or later.
rfind()  with pos parameter only looks in that position or earlier.
if pos == 0 then rfind is faster?

gatecat: Right, makes sense, thanks for explaining

PR #1183, opened 2023-06-22
Remove deprecated timing engine

rowanG077:
This PR removes the old timing engine and fully replaces it with the new one.

Things done to ensure reports match the old one:

- [x] Compare reports with a real world design and ensure there are no difference
- [x] Write a small python script to compare JSON reports. With `detailed_net_timings` this simply became infeasible to do manually. They match.

There are some net differences in the report. But these were all nets with the same path delay as the old one. For example the old report reported a critical path from `some_module.outport[3]` to `some_module.inport[3]` which becomes a critical path from `some_module.outport[0]` to `some_module.inport[0]`. The delays are fully matching.

PR #1219, opened 2023-09-25
TimingEngine: Fix argument order and slack histogram reporting

rowanG077:
Fixes the bug @Ravenslofty reported in https://github.com/YosysHQ/nextpnr/pull/1183#issuecomment-1732692287

Also fixes missing slack histogram printing which I noticed while debugging this.


gatecat:
Thanks!

PR #1222, opened 2023-09-28
MachXO2/XO3/XO3D timing data use

mmicko:
require https://github.com/YosysHQ/prjtrellis/pull/234 and https://github.com/YosysHQ/prjtrellis-db/pull/14


mmicko:
Updated trellis, looks all fine after merge

machxo2/arch.cc
```diff
@@ -532,6 +631,337 @@ DecalXY Arch::getGroupDecal(GroupId group) const
...
+        if (port == id_O)
+            return TMG_STARTPOINT;
+        return TMG_IGNORE;
+    } else if (cell->type == id_DCCA) {
+        if (port == id_CLKI)
+            return TMG_COMB_INPUT;
+        if (port == id_CLKO)
+            return TMG_COMB_OUTPUT;
+        return TMG_IGNORE;
+/*    } else if (cell->type == id_DCSC) {
```
gatecat, line 768: can you remove the commented out sections here?

machxo2/arch.cc
```diff
@@ -362,35 +367,127 @@ BelId Arch::getBelByLocation(Loc loc) const
...
+    std::pair<int, int> dst_loc;
+    if (wire_loc_overrides.count(dst)) {
+        dst_loc = wire_loc_overrides.at(dst);
+    } else {
+        dst_loc = est_location(dst);
+    }
+
+    int dx = abs(src_loc.first - dst_loc.first), dy = abs(src_loc.second - dst_loc.second);
+
+    return (120 - 22 * device_speed) *
```
gatecat, line 407: if you have more trouble routing with this PR, it might be worth just trying increasing the 120 and seeing what happens - as xo2 is slower than ecp5, the delay estimates and predictions may well need to be higher too

machxo2/arch.cc
mmicko: this is now removed

machxo2/arch.cc
mmicko: added higher value, but also made sure that we always use actual speed, so speed 5 is same for XO3 (where only 5 and 6 are available) and for XO2 ( where we have all speeds from 1 to 6)

PR #1223, opened 2023-09-28
kernel/timing: use std::numeric_limits::lowest() instead of min()

rowanG077:
`std::numeric_limits<delay_t>::min()` returns the smallest number larger then 0 for `float` which broke the worst endpoint search for timing reporting for any arch that uses `delay_t = float` or ``delay_t = double`` 

I have found some other suspect spots in the code. @gatecat please go over them whether they are correct. (I'm commenting these in a new comment)

### DIFF
--- a/common/kernel/timing.cc
+++ b/common/kernel/timing.cc
@@ -770,7 +770,7 @@ void TimingAnalyser::build_detailed_net_timing_report()
 std::vector<CellPortKey> TimingAnalyser::get_worst_eps(domain_id_t domain_pair, int count)
 {
     std::vector<CellPortKey> worst_eps;
-    delay_t last_slack = std::numeric_limits<delay_t>::min();
+    delay_t last_slack = std::numeric_limits<delay_t>::lowest();
     auto &dp = domain_pairs.at(domain_pair);
     auto &cap_d = domains.at(dp.key.capture);
     while (int(worst_eps.size()) < count) {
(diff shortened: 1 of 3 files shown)


rowanG077:
https://github.com/YosysHQ/nextpnr/blob/c07ca64ebefc9d796ae0e11dc83a2c11e96a8b8b/common/route/router1.cc#L814

https://github.com/YosysHQ/nextpnr/blob/c07ca64ebefc9d796ae0e11dc83a2c11e96a8b8b/common/route/router1.cc#L832

https://github.com/YosysHQ/nextpnr/blob/c07ca64ebefc9d796ae0e11dc83a2c11e96a8b8b/common/route/router1.cc#L944

https://github.com/YosysHQ/nextpnr/blob/c07ca64ebefc9d796ae0e11dc83a2c11e96a8b8b/fpga_interchange/cost_map.cc#L321

mmicko:
Thanks @rowanG077  it explains why my code did not work fine before https://github.com/YosysHQ/nextpnr/pull/1222/commits/88c4504ff419b63ac4d8efc88f8ffef2b152c376 
Tried now this change of yours and putting it back to float and I get timing info fine, which was not the case at all before

mmicko:
@gatecat no worries, xo2/3 stays on int for compatibility.

rowanG077:
@gatecat Changed those to `::lowest()` as well. Ready to be merged

PR #1224, opened 2023-09-29
placer_static: Initial prototype

gatecat:
This is not yet at a point where it will bring benefits for real-world designs, but it is now a complete and working placer flow including a legaliser.

The primary remaining issue is that we aren't doing enough to prepare the placement for legalisation (like converging carry chains, dealing with control set issues, etc etc) which means there's a big jump in wirelength and so degrade in QoL at this point that the detail placer than has to try and deal with (but isn't designed for).

Other issues are that the heuristics for adjusting penalties etc are currently very crude, and it is likely that in some cases the solver just won't converge. There is also scope for multithreading in lots of places (perhaps with OpenMP), although runtime isn't that bad. The FFTs do mean there is a relatively high constant overhead for small designs on large devices compared to other placer approaches, though.

### DIFF
--- a/3rdparty/oourafft/fftsg.cc
+++ b/3rdparty/oourafft/fftsg.cc
@@ -1,0 +1,3305 @@
+///////////////////////////////////////////////////////////////////////////////
+// BSD 3-Clause License
+//
+// Copyright (c) 2018-2020, The Regents of the University of California
+// All rights reserved.
+//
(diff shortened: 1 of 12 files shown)


gatecat:
This isn't yet _beating_ HeAP, and it's still missing considerable sophistication, but it's also not doing much worse, so let's merge it and follow up with improvements in future PRs.

PR #1225, opened 2023-10-03
gowin: Himbaechel. Initial BSRAM support

yrabbit:
Only pROM/pROMX9 for now


yrabbit:
Compatible with older apicula bases that don't know anything about BSRAM.

PR #1232, opened 2023-10-28
himbaechel: Generation speedup and improvements

gatecat:
Some changes split out of my ongoing nextpnr-xilinx backport-to-himbaechel work:

 - Significant speedup and memory footprint reduction in database generation with some evil tricks
 - Adding a chipdb "schema" version number
 - Adding extra data for PIPs
 - Adding some helper functions for package pin lookup

@yrabbit if you have a chance can you check this breaks nothing on Gowin?

### DIFF
--- a/himbaechel/arch.cc
+++ b/himbaechel/arch.cc
@@ -32,6 +32,8 @@
 
 NEXTPNR_NAMESPACE_BEGIN
 
+static constexpr int database_version = 2;
+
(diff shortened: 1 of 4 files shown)


yrabbit:
This did not break the compilation of any of the himbaechel-gowin examples. My favorite ATTOSOC was loaded into Tangnano9k, Tangnano20k and Tangprimer20k boards and showed that the resulting images were working properly.

Extra information for PIPs will come in very handy - I previously regretted that there is no such mechanism :)

PR #1233, opened 2023-10-29
ice40: add static placer support

Ravenslofty:


ice40/arch.cc
```diff
@@ -650,6 +651,45 @@ bool Arch::place()
...
+            comb.cell_area[id_ICESTORM_LC] = StaticRect(1.0f, 0.125f);
+            comb.bel_area[id_ICESTORM_LC] = StaticRect(1.0f, 0.125f);
+            comb.spacer_rect = StaticRect(1.0f, 0.125f);
+        }
+
+        {
+            cfg.cell_groups.emplace_back();
+            auto &comb = cfg.cell_groups.back();
+            comb.name = getCtx()->id("RAM");
+            comb.cell_area[id_ICESTORM_RAM] = StaticRect(2.0f, 1.0f);
```
gatecat, line 669: a vertical shape (1x2 instead of 2x1) would make more sense here, considering the iCE40 arch

ice40/arch.cc
```diff
@@ -650,6 +651,45 @@ bool Arch::place()
...
+            cfg.cell_groups.emplace_back();
+            auto &comb = cfg.cell_groups.back();
+            comb.name = getCtx()->id("GB");
+            comb.cell_area[id_SB_GB] = StaticRect(0.5f, 0.5f);
+            comb.bel_area[id_SB_GB] = StaticRect(0.5f, 0.5f);
+            comb.spacer_rect = StaticRect(0.5f, 0.5f);
+        }
+
+        {
+            cfg.cell_groups.emplace_back();
```
gatecat, line 684: is this really needed? I think static should be able to place singletons okay without this being configured anyway

Ravenslofty: Yes, the assert I enhanced trips without this.

gatecat: Okay, let's leave this for now and then I'll fix it later.

PR #1235, opened 2023-11-07
himbaechel: Adding a xilinx uarch for xc7 with prjxray

gatecat:
This PR ports the xray/xc7 side (i.e. everything without rapidwright/ultrascale) of the old nextpnr-xilinx repo into himbächel.

This is not finished or working, and it's a long way from being anything production-grade, but it's a useful test of the himbächel framework for something more complex.

~~As well as it just not quite producing working bitstreams yet~~ blinky at least works now, there are several other key issues to be investigated and solved:

 - no timing data
 - router isn't tuned (delay estimates etc) so performs terribly
 - router2 doesn't work at all because of some problem relating to how the new constants API is implemented (this arch is also the first serious test of that)
 - some IOLOGIC code is missing
 - the database build performance and memory footprint needs improving for bigger devices, maybe moving away from python eventually
 - it'd be nice to use the "new" support for cell/bel pin mappings and bel buckets instead of the approach nextpnr-xilinx took at the time of just changing cell type and port names as needed


hansfbaier:
Would be great so see how well this performs with one of the larger parts, like the Kintex 420/480T which I have contributed to prjxray.

Issue #1236, opened 2023-11-07
Routing failure on MachXO3


povik:
On nextpnr 4a7e58a, trellis 36c615, trellis-db 4dda149 I got the following

```
Info: Routing globals...
ERROR: Failed to route net 'clk$TRELLIS_IO_IN$glb_clk' from X24/Y15/G_CLKO1_DCC to X4/Y4/CLK1_SLICE using dedicated routing.
0 warnings, 1 error
```

when routing a simple design with a registered multiplier

```
read_verilog <<EOF
module top(input [16:0] a, input [16:0] b, output reg [33:0] o, input clk);
	reg [16:0] ar;
	reg [16:0] br;

	always @(posedge clk) begin
		ar <= a;
		br <= b;
		o <= $signed(ar) * $signed(br);
	end
endmodule
EOF
synth_lattice -family xo3 -json multest.json
exec -- nextpnr-machxo2 --json multest.json --device=LCMXO3LF-9400C-6BG484C
```

If I increase the dimension of the multiplier, the routing stops failing, which suggests to me this is a bug in nextpnr. Attaching logs and the JSON result of synthesis.

[log.txt](https://github.com/YosysHQ/nextpnr/files/13281483/log.txt)
[multest.json](https://github.com/YosysHQ/nextpnr/files/13281545/multest.json)

gatecat:
@mmicko I would guess this a problem with some clock routing resources being missing so some DCCs aren't working properly, would you be able to look into this?

mmicko:
Problem occurs when there are no constraints, so `promote_globals` work fine and add DCC and route trough it, but later `place_globals` fails, since there are no constraints (BEL attribute).

gatecat:
Can we make the global placement robust enough that it works even in this case? I imagine the same case of unconstrained position would occur e.g. with a global driven from a fabric clock divider (sure, often bad practice, but still sometimes useful).

mmicko:
Ah. Missed the comment. Will need to think more about possible solution.

PR #1238, opened 2023-11-07
machxo2: ignore promoting to global when bel is not constrained

mmicko:
This fixes https://github.com/YosysHQ/nextpnr/issues/1236 by ignoring nets that are connected to non-constrained IO port.


gatecat:
This is very much not a valid solution to the problem. Please see my comment on the issue.

PR #1239, opened 2023-11-09
machxo2: less pessimistic delay prediction

Ravenslofty:
This puts the post-placement delay prediction in a band that matches the output Fmax a bit better. Though I only tuned this using one design, the figure only has to be roughly correct anyway.


mmicko:
@Ravenslofty at line 404 there is same constant used for `Arch::estimateDelay`, also note that `device_speed` can be between 1 and 6 depending on model so worth of checking with at least 5 and 6 that are for XO3, for XO2 there is whole range 1-6

Ravenslofty:
Truthfully the formulae here looks like a mess and I'd prefer to use something else; Mistral has a very simple but effective [formula](https://github.com/YosysHQ/nextpnr/blob/master/mistral/delay.cc#L417) for example.

Since calibrating these values is difficult I'm tempted to write a fake router which can produce a lookup table.

PR #1241, opened 2023-11-13
gowin: Himbaechel. Diff io BUGFIX.

yrabbit:
Fixed incorrect use of attributes instead of parameters.


uis246:
OSER10/10. Works with my project.

PR #1246, opened 2023-11-17
heap: improved net weighting

Ravenslofty:
This might be a little controversial.

While *doing research*, I found [Agashiwala, 2015](https://conservancy.umn.edu/handle/11299/175486), which describes a quadratic analytical placer. While the paper itself wasn't too interesting, I found Formula 3.8 on timing-driven net weighting, which intrigued me because the formula in nextpnr scales the weight down with users, and this one scales the weight up with users.

So, I compared it to the old formula on 100 runs each of dragonmux's [HeadphoneAmp](https://github.com/dragonmux/HeadphoneAmp), and got this:

firefox_IzQan83c3B

and my SPRT calculator is happy to accept this as an improvement for this design.

My question is: given how heap is the default placer everywhere in nextpnr, what's the criteria for "seems sufficiently unlikely to regress people's designs"?

### DIFF
--- a/common/place/placer_heap.cc
+++ b/common/place/placer_heap.cc
@@ -752,8 +752,7 @@ class HeAPPlacer
                                                                        std::abs(o_pos - this_pos)));
 
                     if (user_idx) {
-                        weight *= (1.0 + cfg.timingWeight * std::pow(tmg.get_criticality(CellPortKey(port)),
-                                                                     cfg.criticalityExponent));
+                        weight = tmg.get_criticality(CellPortKey(port)) * ni->users.entries() * double(std::abs(o_pos - this_pos));
                     }
 
                     // If cell 0 is not fixed, it will stamp +w on its equation and -w on the other end's equation,


Ravenslofty:
Okay, this is quite weird; I keep forgetting that criticality is "backwards" [lower = more critical] in nextpnr, which means this is...increasing the weight of low-criticality, high-fanout nets and this is improving the maximum frequency of the design by a statistically significant amount. *eh?*

gatecat:
>  lower = more critical

This definitely shouldn't be true

Ravenslofty:
Am I misunderstanding how criticality in nextpnr works? Because I swear it is from things like router2's `crit_weight` being `1 - crit` (ish).

(it's true I haven't slept yet, but I don't think I'm *that* off the mark)

gatecat:
> router2's crit_weight being 1 - crit

because we want to lower the congestion penalty for more critical nets, favouring a direct route for these.

Ravenslofty:
In the hope that the dramatic wire length increase could be soothed with a wire length term, I added that to the formula.

Instead of easing wire length, this happened:
image

which means this now has a 2.2 MHz increase in median Fmax, compared to 1.2 MHz without the wire length term.

PR #1250, opened 2023-11-22
himbaechel: Fix when more then 32k unique node shapes

gatecat:
Himbächel was always intended to support up to 2³² unique node shapes, hence it being split across two words, but as the multi-purpose structure was being encoded as signed shorts, this would fail in the lower part as soon as 32768 were exceeded.


mmicko (approved):
LGTM. Works with tiles with over 45000 wires

PR #1257, opened 2023-12-08
placer1: Fix various bitrot

gatecat:
Various APIs have changed over time and the updates to placer1 along the way weren't reflecting semantic changes in how validity checking works, meaning that sometimes either initial placement or constraint legalisation would get stuck in an endless or near-endless loop.

### DIFF
--- a/common/place/place_common.cc
+++ b/common/place/place_common.cc
@@ -19,6 +19,7 @@
 
 #include "place_common.h"
 #include <cmath>
+#include "fast_bels.h"
 #include "log.h"
 #include "util.h"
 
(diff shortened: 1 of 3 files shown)


Ravenslofty:
So, I tried this patch, and noticed that it consistently gives `timing cost = 0`, which means the resulting placements perform very badly.

gatecat:
This is also the case without this patch (insofar as SA actually works with out it), though, right?

I suspect this is a totally different issue, possibly relating to the ability to set Fmax in the GUI going missing with various timing reforms.

PR #1258, opened 2023-12-09
static/ice40: bug fixes for ultraplus

Ravenslofty:
- for UP `static` throws an assert over `WARMBOOT` being too big, so it was shrunk a little.
- added DSP bel config (again, the width is set to be less than 1 due to it being too big otherwise).
- fix the IO spacer size to be the size of a cell.
- finally, the iCE40 has the LUT and FF packed together into an `ICESTORM_LC`, so the logic group offset needs to be changed.

### DIFF
--- a/common/place/placer_static.cc
+++ b/common/place/placer_static.cc
@@ -1267,7 +1267,7 @@ class StaticPlacer
                 penalty *= 1.025;
             if (!legalised_ip) {
                 float ip_overlap = 0;
-                for (int i = 2; i < int(groups.size()); i++)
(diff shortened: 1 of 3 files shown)


common/place/placer_static.cc
```diff
@@ -1261,21 +1261,26 @@ class StaticPlacer
         prepare_density_bins();
         initialise();
         bool legalised_ip = false;
+#ifdef ARCH_ICE40
```
gatecat, line 1264: can we make this a config param, not an ifdef?

gatecat: can you remove the now redundant `logic_groups` variable, too?

PR #1260, opened 2023-12-14
Add Rust FFI bindings

Ravenslofty:
As an important first step to getting awooter implemented and merged, we need to get the Rust FFI bindings in. I was a little unsure of how to integrate them, since C++ needs to call into Rust, but this gives an example of how it would work.

(said example is intended to be removed before merge, but exists as a demonstration that the integration works)


rust/rust.cc
```diff
@@ -0,0 +1,323 @@
...
+#include "nextpnr.h"
+
+namespace {
+    USING_NEXTPNR_NAMESPACE;
+
+    template<typename T>
+    uint64_t wrap(T thing) {
+        static_assert(sizeof(T) <= 8, "T is too big for FFI");
+        uint64_t b = 0;
+        memcpy(&b, &thing, sizeof(T));
```
gatecat, line 29: I'd be curious if you could check whether a `reinterpret_cast` type approach works out faster than `memcpy` or not? I'd expect the compiler can optimise this away anyway though? Might be worth a check in godbolt or whatever.

rust/rust.cc
```diff
@@ -0,0 +1,323 @@
...
+    float npnr_context_estimate_delay(const Context *const ctx, uint64_t src, uint64_t dst) { return ctx->getDelayNS(ctx->estimateDelay(unwrap_wire(src), unwrap_wire(dst))); }
+    float npnr_context_get_pip_delay(const Context *const ctx, uint64_t pip) { return ctx->getDelayNS(ctx->getPipDelay(unwrap_pip(pip)).maxDelay()); }
+    float npnr_context_get_wire_delay(const Context *const ctx, uint64_t wire) { return ctx->getDelayNS(ctx->getWireDelay(unwrap_wire(wire)).maxDelay()); }
+    float npnr_context_delay_epsilon(const Context *const ctx) { return ctx->getDelayNS(ctx->getDelayEpsilon()); }
+    Loc npnr_context_get_pip_location(const Context *const ctx, uint64_t pip) { return ctx->getPipLocation(unwrap_pip(pip)); }
+    bool npnr_context_check_pip_avail_for_net(const Context *const ctx, uint64_t pip, NetInfo *net) { return ctx->checkPipAvailForNet(unwrap_pip(pip), net); }
+
+    // This method's in C++ temporarily, while I figure out some better way of getting a pip iterator.
+    Loc npnr_context_get_pip_direction(const Context *const ctx, uint64_t _pip) {
```
gatecat, line 152: Is `DownhillIterWrapper` etc not yet useful for this?

rust/rust.cc
```diff
@@ -0,0 +1,323 @@
...
+        ctx->nets.at
+        ctx->nets.size
+        ctx->rng64
+        ctx->setting<bool>
+        ctx->setting<float>
+        ctx->setting<int>
+        ctx->sorted_shuffle
+    */
+
+    void npnr_log_info(const char *const format) { log_info("%s", format); }
```
gatecat, line 126: this argument is probably better called something other than `format`

Also, I'm not sure if the `const` after the `*` is actually needed here?

rust/rust.cc
```diff
@@ -0,0 +1,323 @@
...
+        auto wire_vec = std::vector<WireId>{};
+        for (auto wire : ctx->getWires()) {
+            wire_vec.push_back(wire);
+        }
+        wire_vec.shrink_to_fit();
+        auto size = wire_vec.size();
+        *wires = wire_vec.data();
+        auto dummy = std::vector<WireId>{};
+        // Yes, by memcpying over `wire_vec` we leak memory.
+        std::memcpy(&wire_vec, &dummy, sizeof(dummy));
```
gatecat, line 210: Could you avoid the memcpy situation here by doing something like `std::vector<WireId> *wire_vec = new std::vector<WireId>{};` `...` and then just never deleting it? That way the vector will also just stay around forever.

gatecat: Another option: iterate over wires once to get the number of wires; allocate an array of WireId of that size; and then iterate again to populate it? This could well be faster than a vector dynamically resizing itself several times; and also means ownership of the allocated memory can be cleanly passed to Rust that could safely delete it if it so desired (and such a thing is possible?)

Final thought here: should this be a list of `uint64_t` wrapped WireIds rather than just a list of WireIds?

rust/rust.cc
```diff
@@ -0,0 +1,323 @@
...
+    const char *npnr_context_name_of(const Context *const ctx, IdString str) { return ctx->nameOf(str); }
+    const char *npnr_context_name_of_pip(const Context *const ctx, uint64_t pip) { return ctx->nameOfPip(unwrap_pip(pip)); }
+    const char *npnr_context_name_of_wire(const Context *const ctx, uint64_t wire) { return ctx->nameOfWire(unwrap_wire(wire)); }
+    bool npnr_context_verbose(const Context *const ctx) { return ctx->verbose; }
+
+    uint64_t npnr_context_get_netinfo_source_wire(const Context *const ctx, const NetInfo *const net) { return wrap(ctx->getNetinfoSourceWire(net)); }
+    uint64_t npnr_context_get_netinfo_sink_wire(const Context *const ctx, const NetInfo *const net, const PortRef *const sink, uint32_t n) { return wrap(ctx->getNetinfoSinkWire(net, *sink, n)); }
+
+    uint32_t npnr_context_nets_leak(const Context *const ctx, int **names, NetInfo ***nets) {
+        auto name_vec = std::vector<int>{};
```
gatecat, line 226: Similar comment as to the `WireId` approach, except here you wouldn't even need a first loop to get the size because you could just call `ctx->nets.size()`...

rust/rust.cc
```diff
@@ -0,0 +1,323 @@
...
+        ctx->getGridDimY
+        ctx->id
+        ctx->nameOf
+        ctx->unbindWire
+        ctx->verbose
+
+        UNNECESSARY:
+        ctx->getDelayNS - all FFI calls go through it anyway.
+
+        TODO:
```
gatecat, line 95: This comment seems out of date with what's actually implemented now

CMakeLists.txt
```diff
@@ -249,7 +256,10 @@ aux_source_directory(3rdparty/json11 EXT_JSON11_FILES)
 aux_source_directory(3rdparty/oourafft EXT_OOURAFFT_FILES)
 aux_source_directory(frontend/ FRONTEND_FILES)
 
+aux_source_directory(rust/ RUST_FILES)
```
gatecat, line 259: This seems to be repeated both in and out of the `USE_RUST` guard?

ice40/arch.cc
```diff
@@ -642,6 +643,12 @@ std::vector<GroupId> Arch::getGroupGroups(GroupId group) const
 
 bool Arch::place()
 {
+#ifdef USE_RUST
```
gatecat, line 646: Suggestion to avoid this `ifdef`: add a Python binding for this function conditional on `USE_RUST` so it can be called as desired

rust/rust.cc
Ravenslofty: The compiler can absolutely optimise this away, and `memcpy` is pretty much what the standard requires for strict aliasing. Not that the rest of this code is necessarily standard-kosher.

ice40/arch.cc
Ravenslofty: > (said example is intended to be removed before merge, but exists as a demonstration that the integration works)

That's meant to cover this; I never intended that to get merged, but as a "look, you can compile it yourself and see that it works" type example.

rust/rust.cc
gatecat: Thanks for confirming!

ice40/arch.cc
gatecat: Yep, I saw, it was just my suggestion if you wanted to keep the demo accessible but remove this code

Ravenslofty: Okay, pybind11 confuses me, so I'm going to have to ask for your help with writing said binding.

uis246:
+12 gigs to build nextpnr

Ravenslofty:
Which architecture are you using to measure that?

Ravenslofty:
nextpnr-ice40 with `cmake -D ARCH=ice40 -D USE_RUST=OFF` uses 2GiB single-threaded:
```
427.10user 16.26system 5:14.81elapsed 140%CPU (0avgtext+0avgdata 2087964maxresident)k
0inputs+4464904outputs (10major+8772302minor)pagefaults 0swaps
```

nextpnr-ice40 with `cmake -D ARCH=ice40 -D USE_RUST=ON`:
```
418.73user 17.09system 5:13.81elapsed 138%CPU (0avgtext+0avgdata 2087900maxresident)k
8inputs+4590320outputs (24major+8957110minor)pagefaults 0swaps
```

I think the memory bottleneck is the BBAs or such, honestly.

uis246:
> I think they meant disk space.

Yes, disk space to compiler rust compiler.

I'm fine if it is completely optional.

whitequark:
Why do you need to compile rustc? I assume you don't build clang from scratch whenever you build Yosys.

rust/rust.cc
```diff
@@ -0,0 +1,231 @@
...
+        *pips = pip_vec->data();
+        // Yes, by never deleting pip_vec, we leak memory.
+        return size;
+    }
+
+    uint64_t npnr_context_get_wires_leak(const Context *ctx, uint64_t **wires) {
+        auto size = size_t{};
+        for (auto _wire : ctx->getWires())
+            size++;
+        auto *wire_vec = new std::vector<uint64_t>{};
```
gatecat, line 117: I think this can be simplified yet further to

`uint64_t *wires = new uint64_t[size];`

ice40/arch.cc
gatecat: I think just adding the necessary include at the top of the file and then adding

```
m.def("example_printnets", example_printnets);
```
after say (and with the necessary ifdef guard)
https://github.com/YosysHQ/nextpnr/blob/56587859d3596c4a42d5f28ae51898ee09cea2c1/common/kernel/pybindings.cc#L251

might work - although it might be you will need to make a function that calls example_printnets taking `Context&` instead of `Context*`

rust/rust.cc
```diff
@@ -0,0 +1,229 @@
...
+    uint64_t npnr_context_get_netinfo_source_wire(const Context *ctx, const NetInfo *net) { return wrap(ctx->getNetinfoSourceWire(net)); }
+    uint64_t npnr_context_get_netinfo_sink_wire(const Context *ctx, const NetInfo *net, const PortRef *sink, uint32_t n) { return wrap(ctx->getNetinfoSinkWire(net, *sink, n)); }
+
+    uint32_t npnr_context_nets_leak(const Context *ctx, int **names, NetInfo ***nets) {
+        auto size = ctx->nets.size();
+        *names = new int[size];
+        *nets = new NetInfo*[size];
+        auto idx = 0;
+        for (auto& item : ctx->nets) {
+            *names[idx] = item.first.hash();
```
gatecat, line 145: nitpick: can you use `.index` rather than `.hash()`

ice40/arch.cc
gatecat: ping?

Ravenslofty: with your patch, this now works :p

PR #1268, opened 2024-01-03
upgrade to C++17

Ravenslofty:
I got a bit annoyed at the lack of C++17 in nextpnr, so here's a PR to upgrade the C++ standard, as well as a fix to hashlib to remove the deprecated `std::iterator` usage.

### DIFF
--- a/common/kernel/hashlib.h
+++ b/common/kernel/hashlib.h
@@ -380,8 +380,8 @@ template <typename K, typename T, typename OPS> class dict
         using iterator_category = std::forward_iterator_tag;
         using value_type = std::pair<K, T>;
         using difference_type = std::ptrdiff_t;
-        using pointer = std::pair<K, T>*;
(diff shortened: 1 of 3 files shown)


common/kernel/hashlib.h
```diff
@@ -818,6 +828,11 @@ template <typename K, typename OPS> class pool
         const_iterator(const pool *ptr, int index) : ptr(ptr), index(index) {}
 
       public:
+        using iterator_category = std::forward_iterator_tag;
+        using value_type = K;
+        using difference_type = std::ptrdiff_t;
+        using pointer = K*;
```
gatecat, line 834: I think this should be `const K*` and `'const K&`

Ravenslofty: Fixed, although my understanding based on the defaults of `std::iterator` is that non-`const` was the behaviour here, and I didn't want to change that without your approval.

PR #1269, opened 2024-01-03
Fix: misc warnings

dragonmux:
This PR seeks to address some of the sea of unused parameter warnings created when compiling with `-Wall -Wextra  -Wpedantic` on GCC, and the Rust FFI bindings warnings while improving const and reference correctness in this code.

Tested to the best of our abilities with the iCE40 backend.

### DIFF
--- a/rust/rust.cc
+++ b/rust/rust.cc
@@ -16,6 +16,7 @@
  *  OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  */
 
+#include <array>
 #include <numeric>
 #include "log.h"
 #include "nextpnr.h"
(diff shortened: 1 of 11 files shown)


rust/rust.cc
```diff
@@ -101,31 +100,31 @@ extern "C" {
...
+    uint64_t npnr_context_get_wires_leak(const Context *const ctx, uint64_t **const wires) {
+        const auto ctx_wires{ctx->getWires()};
+        const auto size{
+            std::accumulate(ctx_wires.begin(), ctx_wires.end(), /*initial value*/ size_t{},
+                [](size_t value, const auto &/*wire*/) { return value + 1U; }
+        )};
         *wires = new uint64_t[size];
         auto idx = 0;
-        for (auto wire : ctx->getWires()) {
+        for (const auto &wire : ctx_wires) {
```
gatecat, line 127: there isn't a need for making this a reference, as `WireId` is not expected (and indeed, required not to be in this case), more than a 64 bit value

dragonmux: If this is done without being `const auto &`, you will always copy and as this is not a trivial type (which is why the warnings btw) that will always generate a call to the copy constructor, even though that might be able to fit in a 64-bit register in your mental model. This is expensive and avoidable - similarly, it will generate destructor calls after, potentially invalidating the wrapped object state that's just been copied.

This construction guarantees this operation to remain cheap, no matter how `WireId` winds up implemented in the future, and the wrapped objects to remain valid post-loop.

gatecat: Everything in nextpnr would break, at a minimum from a performance PoV, if the ID types had non-trivial copies or destructors - the entire principle throughout nextpnr to pass them by value.

dragonmux: One of the backends makes WireId a IdString, another it's CycloneV::rnode_t, etc (couldn't reason about triviality).. there are several where it's two `int32_t`'s, which will spill to two registers here. Using `const auto &` prevents this and any possible impact and eliminates a copy. Using a `const auto &` here also ensures that we only copy once - in wrap() - from the storage in the holder type directly into the FFI boundary storage.

Iff WireId was fully POD in all cases, GCC would not have warned. It's telling us that the type is non-POD (so also non-trivial) even if it is cheap-ish to copy.

gatecat: `rnode_t` is just a typedef to an `int32_t` and an `IdString` is a class with a single `int32_t` member. 

I think most of this "cheap-ish" copies should be optimised away, but hopefully so will the reference here, so fine, I won't bother arguing further.

whitequark: > there are several where it's two int32_t's, which will spill to two registers here

That doesn't make the type non-POD. I see you use the same argument strategy here as you did on the Amaranth channels. It is equally unwelcome.

PR #1270, opened 2024-01-04
Add trivially copyable invariant for ID types

gatecat:
This has always been a design intent of nextpnr, and is now strictly required by #1260.


Ravenslofty:
...we should also have the invariant that `sizeof({Bel,Pip,Wire}Id) <= 8)` or else #1260 would only partially capture some of those types.

gatecat:
True - but for now I would suggest putting that in the Rust bindings rather than here.

PR #1273, opened 2024-01-16
Rust bindings cleanup

Ravenslofty:
I apologise for the code churn, but this will make it more pleasant to use, instead of fighting raw pointers everywhere.

- a mutex is added to guard manipulation of the architecture backend; this might get broken into finer-grained locking in the future, but it'll do for now.
- the big change is that the FFI function declarations take references instead of pointers, because pretty much all functions would choke on a null pointer anyway; this reduces the number of functions that clippy demands to be marked as `unsafe` as a helpful side-effect.
- while doing this, I decided to change the API of `netinfo_users_leak` a bit, because I'm reasonably sure `PortRef`s shouldn't be modified (and then changed `npnr_portref_cell` to reflect this).
- since `Nets` doesn't contain any raw pointers, I can remove the manual implementations of `Send`/`Sync` (in hindsight this commit is badly named)

### DIFF
--- a/rust/nextpnr/src/lib.rs
+++ b/rust/nextpnr/src/lib.rs
@@ -32,7 +32,7 @@ pub struct NetInfo {
 }
 
 impl NetInfo {
-    pub fn driver(&mut self) -> &mut PortRef {
+    pub fn driver(&mut self) -> Option<&mut PortRef> {
         unsafe { npnr_netinfo_driver(self) }
     }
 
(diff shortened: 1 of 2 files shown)


rust/nextpnr/src/lib.rs
```diff
@@ -32,7 +32,7 @@ pub struct NetInfo {
 }
 
 impl NetInfo {
-    pub fn driver(&mut self) -> *mut PortRef {
+    pub fn driver(&mut self) -> &mut PortRef {
```
gatecat, line 35: hmm, I think this is one case where a null could exist - probably this should be converting the pointer to a ref in an option or something?

PR #1278, opened 2024-01-23
gowin: Himbaechel. Deal with SP BSRAM ports.

yrabbit:
The OCE signal in the SP(X)9B primitive is intended to control the built-in output register. The documentation states that this port is invalid when READ_MODE=0 is used. However, it has been experimentally established that you cannot simply apply VCC or GND to it and forget it
- the discrepancy between the signal on this port and the signal on the CE port leads to both skipping data reading and unnecessary reading after CE has switched to 0.
Here we force these ports to be connected to the network, except in the case where the user controls the OCE signal using non-constant signals.

Also:
  * All PIPs for clock spines are made inaccessible to the common router - in general, using these routes for signals that have not been processed by a special globals router is fraught with effects that are difficult to detect.
  * The INV primitive has been added purely to speed up development - this primitive is not generated by Yosys, but is almost always present in vendor output files.

### DIFF
--- a/himbaechel/uarch/gowin/globals.cc
+++ b/himbaechel/uarch/gowin/globals.cc
@@ -40,6 +40,8 @@ struct GowinGlobalRouter
 
     GowinGlobalRouter(Context *ctx) : ctx(ctx) { gwu.init(ctx); };
 
+    bool checkPipAvail(PipId pip) const { return gwu.is_global_pip(pip) || ctx->checkPipAvail(pip); };
+
     // allow io->global, global->global and global->tile clock
     bool global_pip_filter(PipId pip) const
     {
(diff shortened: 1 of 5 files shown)


himbaechel/uarch/gowin/globals.cc
```diff
@@ -40,6 +40,8 @@ struct GowinGlobalRouter
 
     GowinGlobalRouter(Context *ctx) : ctx(ctx) { gwu.init(ctx); };
 
+    bool checkPipAvail(PipId pip) const { return gwu.is_global_pip(pip) || ctx->checkPipAvail(pip); };
```
gatecat, line 43: Forcing a PIP to be always available is likely to cause nasty problems, are you sure you want to be doing this?

gatecat: Ah, my error, this is inside the global router not the himbaechel API implementation, should be okay, but can you give this function a different name so it doesn't look like the nextpnr arch one? (say `global_pip_available`)

yrabbit: You're right, now I myself don't like their similarity. I'll change it.

yrabbit: Plus I broke the style rules, it was sloppy. :(

PR #1279, opened 2024-01-24
static: Add a basic threadpool

gatecat:
Not seeing much speedup (or even parallel CPU usage yet), but it's a starting point there.


whitequark:
Not a blocker for this PR, but can we add a Wasm build to nextpnr?

gatecat:
Yeah, I'd very much support having that in CI

PR #1283, opened 2024-01-28
gowin: Himbaechel. Global router BUGFIX.

yrabbit:
Ignore networks without users.

Addressed https://github.com/YosysHQ/apicula/issues/232


yrabbit:
Oh, this is quite expected - in your first message it is indicated that the router cannot route from the PLL output to X33Y8/A6.
 Here the destination is not a clock input or IO pin and such networks are currently given to the general purpose router. Perhaps in the future I will complicate the dedicated router or not - I somehow don’t really like using the clock as an input for the LUTs :smile:

Thank you.

Issue #1275, opened 2024-01-20
Gowin: failed route io with ddr 


andryblack:
failed route IOBUF with ODDR and IDDR together:
```
module test_io (
    input sys_clk,  // 27 Mhz, crystal clock from board
    input sys_resetn,
    input button,   // 0 when pressed

    output [5:0] led,

    inout test_io
);

wire test_i,test_o,test_oen;
IOBUF io_buf(
    .I(test_i),
    .O(test_o),
    .IO(test_io),
    .OEN(test_oen)
);
ODDR oddr_rwds(
    .CLK(sys_clk), .D0(button), .D1(~button), .TX(button), .Q0(test_i), .Q1(test_oen)
);
IDDR iddr_rwds(
    .CLK(sys_clk), .D(test_o), .Q0(led[0]), .Q1(led[1])
);
endmodule
```

nextpnr-himbaechel:

ERROR: Can't place iddr_rwds at X0Y16/IOLOGICA because it's already taken by oddr_rwds

nextpnr-gowin:

ERROR: Cell 'iddr_rwds' cannot be bound to bel 'R29C15_IOLOGICB' since it is already bound to cell 'oddr_rwds'

yrabbit:
Your problem has been considered and patches for apicula and nextpnr-himbaechel have been prepared.
Such complex changes will require some dancing with the sequence of accepting a PR in one water product, releasing and accepting a PR in another.

It's slow official way. But you can manually fix gowin_pack yourself, and also fix and compile nextpnr.

https://github.com/YosysHQ/apicula/pull/227

https://github.com/yrabbit/nextpnr/commit/1ccd4e56da3b753c61d94bc66b049cf091fa728b

andryblack:
Thanks, its build input output with manual placed IOBUF, but now crashed with generated:
 ```
module test_io (
    input sys_clk,  // 27 Mhz, crystal clock from board
    input sys_resetn,
    input button,   // 0 when pressed

    output [5:0] led,

    inout test_io
);


wire test_i,test_o,test_oen;
ODDR oddr_rwds(
    .CLK(sys_clk), .D0(button), .D1(~button), .TX(button), .Q0(test_i), .Q1(test_oen)
);
IDDR iddr_rwds(
    .CLK(sys_clk), .D(test_o), .Q0(led[0]), .Q1(led[1])
);
// IOBUF io_buf(
//     .I(test_i),
//     .O(test_o),
//     .IO(test_io),
[...]

PR #1290, opened 2024-02-09
gowin: Himbaechel. Allow to combine IOLOGIC.

yrabbit:
Corrects the situation when it is impossible to use IOBUF with two IOLOGIC elements at the same time - input and output.

Addresses https://github.com/YosysHQ/nextpnr/issues/1275

This is done by dividing one IOLOGIC Bel into two - input IOLOGIC and output IOLOGIC plus checking for compatibility of the cells located there.

At the moment, this check is simple and allows only the combination of DDR and DDRC primitives.


yrabbit:
Checks are complete, but Apicula release is needed to get the base chips correct.

yrabbit:
OK. apicula has a release.

PR #1295, opened 2024-02-29
Himbaechel GUI

mmicko:
This extends Himbaechel API with additional methods for drawing bels, wires and pips. There is also configuration attribute to control if some of these are used and also do change in coordinate system (y axis).

Was mostly mimic what we do on ECP5 and XO2/3, and was not sure about equivalent for tile_wire (mapped to GfxTileWireId) so used flags for now, since it anyway keeps some additional custom info. 

First few commits are mostly cleanup and fixes of some old standing GUI issues.

Would add at least basic things in Xilinx and Gowin when we agree on API level. Please do comment and suggest changes.


himbaechel/uarch/xilinx/examples/arty-a35/blinky.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
 set -ex
 yosys -p "synth_xilinx -flatten -abc9 -nobram -arch xc7 -top top; write_json blinky.json" blinky.v
-nextpnr-himbaechel --device xc7a35tcsg324-1 -o xdc=arty.xdc --json blinky.json -o fasm=blinky.fasm --router router2
+../../../../../nextpnr-himbaechel --device xc7a35tcsg324-1 -o xdc=arty.xdc --json blinky.json -o fasm=blinky.fasm --router router2
```
gatecat, line 4: this breaks if someone does an out of tree build

himbaechel/himbaechel_api.h
```diff
@@ -102,6 +109,20 @@ struct HimbaechelAPI
     virtual bool isClusterStrict(const CellInfo *cell) const;
     virtual bool getClusterPlacement(ClusterId cluster, BelId root_bel,
                                      std::vector<std::pair<CellInfo *, BelId>> &placement) const;
+
+    // Graphics
+    virtual uint32_t gfxAttributes() { return 0; }
+
+    virtual void gfxTileBel(std::vector<GraphicElement> &g, int x, int y, int z, int w, int h,
```
gatecat, line 116: can you perhaps think of some slightly more descriptive names for these API functions, and also add some documentation about them, so I can review them better?

himbaechel/himbaechel_api.h
```diff
@@ -102,6 +109,20 @@ struct HimbaechelAPI
     virtual bool isClusterStrict(const CellInfo *cell) const;
     virtual bool getClusterPlacement(ClusterId cluster, BelId root_bel,
                                      std::vector<std::pair<CellInfo *, BelId>> &placement) const;
+
+    // Graphics
+    virtual uint32_t gfxAttributes() { return 0; }
```
gatecat, line 114: I really don't like this function, can you think of a better way here, like perhaps a struct that the architecture configures?

himbaechel/himbaechel_api.h
mmicko: sure, will update these, but probably not today

mmicko:
@gatecat Documentation is not added yet, but main question I have is if using tilewire as additional data is fine (need to bump db version in that case as well) ? not sure if there is other way we can actually use to determine which wire to draw.

himbaechel/uarch/example/gfx.h
```diff
@@ -0,0 +1,158 @@
...
+ */
+
+#ifndef HIMBAECHEL_EXAMPLE_GFX_H
+#define HIMBAECHEL_EXAMPLE_GFX_H
+
+#include "nextpnr.h"
+
+NEXTPNR_NAMESPACE_BEGIN
+
+enum GfxTileWireId
```
gatecat, line 27: this will need to be in a uarch-specific namespace, otherwise there will be conflicts

himbaechel/himbaechel_dbgen/chip.py
```diff
@@ -857,3 +865,18 @@ def write_bba(self, filename):
             bba.ref('chip_info')
             self.serialise(bba)
             bba.pop()
+
+    def read_gfx_h(self, filename):
```
gatecat, line 869: this is kind of horrid

himbaechel/himbaechel_dbgen/chip.py
```diff
@@ -144,6 +144,7 @@ class TileWireData:
     index: int
     name: IdString
     wire_type: IdString
+    tile_wire: int
```
gatecat, line 147: this should be given a name that makes it's very clear that it's gui-specific, otherwise it will be very unclear to anyone looking at the structures

himbaechel/arch.cc
```diff
@@ -504,4 +504,61 @@ IdString Arch::get_tile_type(int tile) const
...
 }
 
+std::vector<GraphicElement> Arch::getDecalGraphics(DecalId decal) const
+{
+    std::vector<GraphicElement> ret;
+    if (decal.type == DecalId::TYPE_BEL) {
+        BelId bel(decal.tile, decal.index);
+        GraphicElement::style_t style = decal.active ? GraphicElement::STYLE_ACTIVE : GraphicElement::STYLE_INACTIVE;
+        uarch->drawBel(ret, style, getBelType(bel), getBelLocation(bel));
+    } else if (decal.type == DecalId::TYPE_WIRE) {
```
gatecat, line 514: if a wire spans multiple tiles, as it can in the himbaechel deduplication scheme, this would need to deal with all the constituent tile wires not just the index one

stacksmith:
Is this close to merging?  I am pretty desperate to see things, and `nextpnr-gowin --gui `just crashes all the time now.

himbaechel/arch.cc
mmicko: @gatecat other points are addressed, except this one. Is there already some other part of code where there is also looping the list of constituent tile wires ? if so would like to be consistent.

himbaechel/himbaechel_dbgen/chip.py
```diff
@@ -229,9 +232,13 @@ def add_bel_pin(self, bel: BelData, pin: str, wire: str, dir: PinType):
 
     def create_wire(self, name: str, type: str="", const_value: str=""):
         # Create a new tile wire of a given name and type (optional) in the tile type
+        gfx_wire_id = 0
+        if (name) in self.gfx_wire_ids:
```
gatecat, line 236: what are the brackets here for?

himbaechel/himbaechel_dbgen/chip.py
```diff
@@ -166,6 +167,7 @@ def serialise_lists(self, context: str, bba: BBAWriter):
     def serialise(self, context: str, bba: BBAWriter):
         bba.u32(self.name.index)
         bba.u32(self.wire_type.index)
+        bba.u32(self.gfx_wire_id)
```
gatecat, line 170: this is a breaking change for existing binaries so it needs a db version bump:

https://github.com/YosysHQ/nextpnr/blob/9c2d96f86ed56b77c9c325041b67654f26308270/himbaechel/arch.cc#L36

and

https://github.com/YosysHQ/nextpnr/blob/9c2d96f86ed56b77c9c325041b67654f26308270/himbaechel/himbaechel_dbgen/chip.py#L834

himbaechel/himbaechel_dbgen/chip.py
mmicko: ah, that is leftover from past. removing it now

himbaechel/himbaechel_dbgen/chip.py
gatecat: this needs to go to 5, 4 is the current version on main

mmicko: ok to rebase it  now, so it is easier then to spot things like this ?

gatecat: sounds like a good plan!

gatecat:
Thanks, sorry for the delay in getting this merged!

PR #1296, opened 2024-02-29
Support building nextpnr with nix flakes

RCoeurjoly:
So far GUI is disabled, because nixpkgs has only qt6, and not qt5.

If enabling the GUI is required, with a little more love it should be doable.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,69 @@
+{
+  description = "A flake for the nextpnr FPGA place and route tool";
+
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+    flake-utils.url = "github:numtide/flake-utils";
+  };
+
+  outputs = { self, nixpkgs, flake-utils, ... }:
+    flake-utils.lib.eachDefaultSystem (system:
+      let
+        pkgs = import nixpkgs {
+          inherit system;
+          overlays = [];
+        };
+        # Define custom overrides or additional packages
+        myPythonPackages = pkgs.python3Packages.override {
+          overrides = self: super: {
+            apycula = super.apycula.override {
+              # Place any necessary overrides here
+            };
+          };
+        };
+        enableGui = false;
+      in
+      {
+        packages = {
+          nextpnr = pkgs.stdenv.mkDerivation {
+            pname = "nextpnr";
+            version = "0.6";
+            src = ./.;
+
+            nativeBuildInputs = [ pkgs.cmake ]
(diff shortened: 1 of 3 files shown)


flake.nix
```diff
@@ -0,0 +1,69 @@
...
+              pkgs.python3
+              pkgs.eigen
+              myPythonPackages.apycula
+              pkgs.icestorm
+              pkgs.trellis
+            ] ++ (pkgs.lib.optional enableGui pkgs.qt6.qtbase)
+              ++ (pkgs.lib.optional pkgs.stdenv.cc.isClang pkgs.llvmPackages.openmp);
+
+            cmakeFlags = [
+              "-DARCH=generic;ice40;ecp5;gowin"
```
whitequark, line 46: why not machxo2 or nexus?

rowanG077:
I would like this see the flake replace the `shell.nix`. Important there are the shell hooks so cmake can find the deps in a dev shell.

flake.nix
RCoeurjoly: Thanks for pointing this out. I changed it to target all architectures.

Targeting all+alpha results in the following error:

> -- Configuring architecture: mistral
> CMake Error at mistral/family.cmake:4 (add_subdirectory):
>   add_subdirectory given source "/tools" which is not an existing directory.
> Call Stack (most recent call first):
>   CMakeLists.txt:342 (include)
> 
> 
> CMake Error at mistral/family.cmake:5 (add_subdirectory):
>   add_subdirectory given source "/generator" which is not an existing
>   directory.
> Call Stack (most recent call first):
>   CMakeLists.txt:342 (include)
> 
> 
> CMake Error at mistral/family.cmake:6 (add_subdirectory):
>   add_subdirectory given source "/libmistral" which is not an existing
>   directory.
> Call Stack (most recent call first):
>   CMakeLists.txt:342 (include)

flake.nix
```diff
@@ -0,0 +1,85 @@
...
+                platforms = platforms.all;
+                maintainers = with maintainers; [ ];
+              };
+            };
+          };
+
+          packages.default = self.packages.${system}.nextpnr;
+
+          devShell = pkgs.mkShell {
+            buildInputs = [ self.packages.${system}.default ];
```
rowanG077, line 76:
```
- buildInputs = [ self.packages.${system}.default ];
- inputsFrom = [ self.packages.${system}.default ];
```

rowanG077: If I change that and enter the flake dev shell using `nix develop` I can succesfully build nextpnr in that shell.

RCoeurjoly: Yes, changed.

PR #1300, opened 2024-03-18
gowin: Add support for DSP primitives.

yrabbit:
For the following primitives:
  - PADD9
  - PADD18
  - MULT9X9
  - MULT18X18
  - MULT36X36
  - MULTALU18X18
  - MULTALU36X18
  - MULTADDALU18X18
  - ALU54D 
  
packing and processing of fixed wires between macro and between DSP blocks is implemented.
Clusters of DSP and macro blocks are processed using custom placement of cluster elements.


yrabbit:
It may seem that there are a lot of repetitions between the code for individual primitives, but these things are not very easy to use, and although they all work in test examples, I think it’s too early to optimize - it’s easier to quickly correct errors that will come to light during real use.

gatecat:
This all looks good! And there's nothing wrong with uarches updating cell_bel_pins, it was designed exactly for cases like these.

PR #1309, opened 2024-04-07
Gowin. Add PLL pads.

yrabbit:
If the CLKIN input of the PLL is connected to a special pin, then it makes sense to try to place the PLL so that it uses a direct implicit non-switched connection to this pin.

The transfer of information about pins for various purposes has been implemented (clock input signal, feedback, etc), but so far only CLKIN is used.


himbaechel/uarch/gowin/pack.cc
```diff
@@ -2639,6 +2641,26 @@ struct GowinPacker
                         ci.renamePort(ctx->idf("FDLY[%d]", i), ctx->idf("FDLY%d", i));
                     }
                 }
+                // If CLKIN is connected to a special pin, then it makes sense
+                // to try to place the PLL so that it uses a direct connection
+                // to this pin.
+                if (ci.bel == BelId()) {
+                    NetInfo *ni = ci.getPort(id_CLKIN);
+                    if (ni && ni->driver.cell) {
+                        BelId pll_bel = gwu.get_pll_bel(ni->driver.cell->bel, id_CLKIN_T);
```
gatecat, line 2650: I think this code will fail in the `get_pll_bel` call if the driver is unplaced - say logic, or another PLL, or whatever - the test should probably be `ni && ni->driver.cell && ni->driver.cell->bel != BelId()`

yrabbit: oh yes, you're right.

PR #1312, opened 2024-04-17
Enable user to override setupArchContext in Himbaechel arch

mmicko:
This enables  Himbaechel uarch to override and implement additional operations on already constructed context, like loading of generated bitstream back to nextpnr

### DIFF
--- a/himbaechel/himbaechel_api.h
+++ b/himbaechel/himbaechel_api.h
@@ -57,7 +57,9 @@ struct PlacerHeapCfg;
 
 struct HimbaechelAPI
 {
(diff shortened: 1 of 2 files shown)


himbaechel/himbaechel_api.h
```diff
@@ -58,6 +58,7 @@ struct PlacerHeapCfg;
 struct HimbaechelAPI
 {
     virtual void init(Context *ctx);
+    virtual void setupArchContext() {}
```
gatecat, line 61: Can you add a bit of documentation for this?

PR #1314, opened 2024-04-20
Fix path to shared fallback on windows

lushaylabs:
The nextpnr directory was missing from the fallback shared directory on windows only causing the chip db to not be found when using nextpnr-himbaechel through OSS-Cad-suite

### DIFF
--- a/common/kernel/command.cc
+++ b/common/kernel/command.cc
@@ -197,7 +197,7 @@ void init_share_dirname()
         npnr_share_dirname = proc_share_path;
         return;
     }
-    proc_share_path = proc_self_path + "..\\share\\";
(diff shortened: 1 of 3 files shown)


whitequark (commented):
Nextpnr already links to Boost, doesn't it? I feel like that has something that improves path manipulation over `#ifdef` + `stringf("%s%c%s")`, which is just awful.

lushaylabs:
Thank you for the review, I cleaned up the changes using boost as you suggested and tested it on Windows and Mac

gatecat:
I believe this needs to be updated in respect of https://github.com/YosysHQ/nextpnr/pull/1315

PR #1315, opened 2024-04-26
Fixed header files for boost 1.85.0

pdaehne:
nextpnr does not compile with the newest boost version 1.85.0 because the header file "boost/filesystem/convenience.hpp" does not exist anymore (removed because of deprecation). See

https://www.boost.org/users/history/version_1_85_0.html

This pull request replaces convenience.hpp with the correct header file.

### DIFF
--- a/bba/main.cc
+++ b/bba/main.cc
@@ -19,7 +19,7 @@
  */
 
 #include <assert.h>
-#include <boost/filesystem/convenience.hpp>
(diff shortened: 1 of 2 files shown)


whitequark (approved):
[RIP 2002-2024 boost/filesystem/convenience](https://www.mail-archive.com/boost@lists.boost.org/msg02101.html)

PR #1325, opened 2024-05-16
nexus: Use a toposort when preplacing clock primitives

### LINKED ISSUE #1317: Routing fails on Crosslink Nexus: "Failed to route arc 0.0 of net 'clk'"
Struggling with an issue where nextpnr fails while trying to build a design for the CrosslinkNX-40. The full error message is: 
`ERROR: Failed to route arc 0.0 of net 'clk', from X86/Y53/JCLKOP_PLL_CORE_I_PLL_LMMI to X49/Y56/JCLKI_DCC_DCC2.`

I am trying to narrow down my design to a minimum reproducible example but its proving tricky. The design I am working on is for an I2S controller and the issue seems to have something to do with how I am dividing the system clock to the serial bus clock. I am using the Crosslink PLL to generate a 64MHz system clock which I divide to a 4MHz serial clock. 

The TX and RX modules build just fine on their own, but when I try and use both of them in the same design I get the error above. Ill keep working toward something concise that reproduces the issue, but I am finding it hard to grasp which piece of my design is the one actually causing the error. 

I should add, the design is synthesized and place and routed with the Radiant toolchain just fine. So I am pretty confident this is not a fundamental issue with my HDL but some interaction between the HDL and the tool.

gatecat:
Fixes #1317


polymerizedsage:
This seems to fix both of the build issues I was having. Thanks for the quick fix!

PR #1327, opened 2024-05-19
Clock prerouting Gowin fix?

adrianparvino:
I'm encountering a problem with router2 and the global router conflicting with router2: https://gist.github.com/adrianparvino/0dbe37f9199616af909d9f686c3ae69a I *think* it's trying to connect the PLL output to the global wire twice.


gatecat:
Fixed in https://github.com/YosysHQ/nextpnr/pull/1636

PR #1328, opened 2024-05-22
router2: additional heatmap data

Ravenslofty:
Currently, router2 only tracks how many wires of each type have given levels of overuse ("congestion by wiretype"). This PR extends that to cover how many wires of each type are used at all ("utilisation by wiretype") and how many wires are used in each X/Y coordinate ("utilisation by coordinate"),

It may be a good idea to sum overuse by wiretype or coordinate, but it's a little unclear if this would be useful information yet.

### DIFF
--- a/common/route/router2.cc
+++ b/common/route/router2.cc
@@ -1154,7 +1154,7 @@ struct Router2
         return success;
     }
 
-    void write_wiretype_heatmap(std::ostream &out)
+    void write_congestion_by_wiretype_heatmap(std::ostream &out)
     {
         dict<IdString, std::vector<int>> cong_by_type;
         size_t max_cong = 0;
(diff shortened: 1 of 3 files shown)


common/route/router2.cc
```diff
@@ -1180,6 +1180,37 @@ struct Router2
...
     }
 
+    void write_utilisation_by_wiretype_heatmap(std::ostream &out)
+    {
+        dict<IdString, int> util_by_type;
+        for (auto &wd : flat_wires) {
+            IdString type = ctx->getWireType(wd.w);
+            if (wd.curr_cong > 0) {
+                if (util_by_type.count(type))
+                    util_by_type[type] += 1;
```
gatecat, line 1190: I think this should be increasing by `curr_cong` not by a fixed 1, because if two nets are using a wire, you want it to have a utilisation of 2, too

Ravenslofty: Maybe there should be both. If you increase by `curr_cong`, then you can imagine a situation early in routing where 50% of the wires have `curr_cong == 2` and 50% of the wires have `curr_cong == 0`; it would report 100% utilisation, when there's still plenty of space to spread out to. I think that would be a different situation to "actually *all* of the wires here are used and the router is struggling", and a fixed-1 count would let one distinguish between those situations.

gatecat:
This got forgotten about, it all looks okay to me, shall we merge it?

Ravenslofty:
Yes please!

PR #1337, opened 2024-07-04
Gowin. Taking into account the features of ROM

yrabbit:
For pROM(X9) primitives in images generated by Gowin IDE, there is an interesting recommunication of inputs depending on the data bit depth. For example, in some cases, a high logical level may be applied to the Write Enable input, which, let’s say, is not entirely usual for Read Only memory.

Here we will do similar manipulations.

In addition, several minor bug fixes are included:

 - Fixed bit numbering for non-X9 series primitives.
 - Fixed decoder generation for BLKSEL - do not assume unused inputs are connected to GND.
 - Use default values for BSRAM parameters - don't assume their mandatory presence.

### DIFF
--- a/himbaechel/uarch/gowin/pack.cc
+++ b/himbaechel/uarch/gowin/pack.cc
@@ -1,6 +1,7 @@
 #include "design_utils.h"
 #include "log.h"
 #include "nextpnr.h"
+#include "util.h"
 
 #define HIMBAECHEL_CONSTIDS "uarch/gowin/constids.inc"
 #include "himbaechel_constids.h"
@@ -1324,7 +1325,7 @@ struct GowinPacker
     {
         int num = (bit_width == 9 || bit_width == 18 || bit_width == 36) ? 36 : 32;
         for (int i = 0, j = offset; i < num; ++i, ++j) {
-            if (((i + 1) % 9) == 0 && (bit_width == 16 || bit_width == 32)) {
+            if (((j + 1) % 9) == 0 && (bit_width == 16 || bit_width == 32)) {
                 ++j;
             }
             ci->renamePort(ctx->idf(from, i), ctx->idf(to, offset ? j % 36 : j));
@@ -1390,6 +1391,15 @@ struct GowinPacker
             }
             ++idx;
         }
+        switch (idx) {
+        case 1:
+            init |= init << 2; /* fallthrough */
+        case 2:
+            init |= init << 4;
+            break;
+        default:
+            break;
+        }
         lut->setParam(id_INIT, init);
 
         new_cells.push_back(std::move(lut_cell));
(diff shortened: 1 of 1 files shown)


himbaechel/uarch/gowin/pack.cc
```diff
@@ -1602,10 +1609,19 @@ struct GowinPacker
...
+            ci->connectPort(id_WREB, vcc_net);
 
-            for (int i = 0; i < 14; ++i) {
-                ci->renamePort(ctx->idf("AD[%d]", i), ctx->idf("ADA%d", i));
-                ci->copyPortTo(ctx->idf("ADA%d", i), ci, ctx->idf("ADB%d", i));
+            // disconnect lower address bits for ROM
+            static int rom_ignore_bits[] = {2, 4, 8, 16, 32};
+            static int romx9_ignore_bits[] = {9, 9, 9, 18, 36};
+            for (unsigned int i = 0; i < 14; ++i) {
+                if (i < sizeof(rom_ignore_bits) && ((ci->type == id_pROM && bit_width >= rom_ignore_bits[i]) ||
```
gatecat, line 1618: that `sizeof` doesn't look correct - you could either do `sizeof(rom_ignore_bits)/sizeof(int)` or better still `std::array` or `std::vector`

himbaechel/uarch/gowin/pack.cc
```diff
@@ -1798,6 +1823,15 @@ struct GowinPacker
             bsram_fix_blksel(ci, new_cells);
         }
 
+        // XXX
```
gatecat, line 1826: assuming this `XXX` is supposed to indicate some kind of hack, can you add a bit of info about what the exact workaround is here, too?

himbaechel/uarch/gowin/pack.cc
```diff
@@ -1798,6 +1823,15 @@ struct GowinPacker
...
+        NetInfo *oce_net = ci->getPort(id_OCE);
+        if (oce_net == nullptr || oce_net->name == ctx->id("$PACKER_VCC") || oce_net->name == ctx->id("$PACKER_GND")) {
+            if (oce_net != nullptr) {
```
gatecat, line 1829: why is this checking `oce_net == nullptr` and then `oce_net != nullptr` inside of it?

yrabbit: Corrected.

himbaechel/uarch/gowin/pack.cc
yrabbit: Definitely.
Synchronizing CE and OCE helps and the memory works properly in complex systems such as RISC-V emulation and i8080 emulation (with 32K RAM and 16K BSRAM based ROM), but there is no theoretical basis for this fix, so it is a hack.

PR #1342, opened 2024-07-14
Gowin. Implement clock management primitives.

yrabbit:
DQCE and DCS primitives are added.

DQCE allows the internal logic to enable or disable the clock network in the quadrant. When clock network is disabled, all logic drivern by this clock is no longer toggled, thus reducing the total power consumtion of the device.

DCS allows you to select one of four sources for two clock wires (6 and 7). Wires 6 and 7 have not been used up to this point.

Since "hardware" primitives operate strictly in their own quadrants, user-specified primitives are converted into one or more "hardware" primitives as needed.

Also:
  - minor edits to make the most of helper functions like connectPorts()
  - when creating bases, the corresponding constants are assigned to the VCC and GND wires, but for now huge nodes are used because, for an unknown reason, the constants mechanism makes large examples inoperable. So for now we remain on the nodes.

Compatible with older Apicula databases.


rowanG077:
Could you propagate the clock constraint through these primitives? In a similar manner as in https://github.com/YosysHQ/nextpnr/pull/1090/files ?

yrabbit:
I assume that we will add a similar mechanism as soon as we get the clock constraints in principle.

rowanG077:
Ah excuse me I did not realise we have that yet for gowin.

PR #1345, opened 2024-07-29
himbaechel/gowin: add timing information

Ravenslofty:
This wires up enough of the Apicula timing data to give an Fmax for most designs. There are gaps in the implementation, but those gaps also existed in nextpnr-gowin, so it's not much of a loss. 

The Fmax numbers do appear to be *optimistic* though - compared to the vendor tooling the numbers are about 20% higher and I don't entirely trust them. However, if the optimistic numbers say that a design doesn't meet timing, it definitely doesn't.


gatecat (approved):
LGTM, feel free to merge whenever you're happy with it

gatecat:
It would be nice to get BRAM timings added at some point, depending on the clock-to-out time it's not uncommon for it to actually be the critical path.

PR #1348, opened 2024-08-12
SDC parsing support

rowanG077:
The coming 2 months I have a lot of time to work on nextpnr. I want to make the timing changes I make to be small so I will make small PRs to get them merged asap. 

First step is to get SDC support merged with clock frequency support.

### DIFF
--- a/common/kernel/sdc.cc
+++ b/common/kernel/sdc.cc
@@ -27,7 +27,7 @@
 
 NEXTPNR_NAMESPACE_BEGIN
 
-struct TclEntity
+struct SdcEntity
 {
     enum EntityType
     {
@@ -39,8 +39,8 @@ struct TclEntity
     IdString name;
     IdString pin; // for cell pins only
 
-    TclEntity(EntityType type, IdString name) : type(type), name(name) {}
-    TclEntity(EntityType type, IdString name, IdString pin) : type(type), name(name), pin(pin) {}
+    SdcEntity(EntityType type, IdString name) : type(type), name(name) {}
+    SdcEntity(EntityType type, IdString name, IdString pin) : type(type), name(name), pin(pin) {}
 
     const std::string &to_string(Context *ctx) { return name.str(ctx); }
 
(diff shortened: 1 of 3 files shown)


rowanG077:
I will be adding test cases to: https://github.com/rowanG077/nextpnr-thesis-tests

rowanG077:
clock constraint with frequency is confirmed working.

common/kernel/sdc.cc
```diff
@@ -0,0 +1,453 @@
...
+        if (type == ENTITY_PIN) {
+            CellInfo *cell = nullptr;
+            if (ctx->cells.count(name)) {
+                cell = ctx->cells.at(name).get();
+            } else {
+                const std::string &n = name.str(ctx);
+                auto pos = n.rfind('.');
+                if (pos == std::string::npos)
+                    return nullptr;
+                // remove one hierarchy layer due to Radiant weirdness around PLLs etc
```
gatecat, line 72: this is a nexus-specific workaround that I don't think needs to be kept in the generic code

common/kernel/sdc.cc
```diff
@@ -0,0 +1,453 @@
...
+
+#include "log.h"
+#include "nextpnr.h"
+
+#include <algorithm>
+#include <iterator>
+
+NEXTPNR_NAMESPACE_BEGIN
+
+struct TclEntity
```
gatecat, line 30: to avoid a collision with arch-specific stuff, can this be called SdcEntity instead?

common/kernel/sdc.cc
```diff
@@ -0,0 +1,453 @@
...
+            }
+        }
+
+        if (from == nullptr) {
+            log_error("-from is required for set_false_path (line %d)\n", lineno);
+        } else if (to == nullptr) {
+            log_error("-to is required for set_false_path (line %d)\n", lineno);
+        }
+
+        log_info("set_false_path from: %s, to: %s\n", from->name.c_str(ctx), to->name.c_str(ctx));
```
gatecat, line 430: this should probably be a warning that the `set_false_path` is currently being ignored

rowanG077: Fixed

common/kernel/sdc.cc
rowanG077: Fixed

PR #1349, opened 2024-08-13
Gowin. Add DHCEN primitive.

yrabbit:
This primitive allows you to dynamically turn off and turn on the networks of high-speed clocks.

This is done tracking the routes to the sinks and if the route passes through a special HCLK MUX (this may be the input MUX or the output MUX, as well as the interbank MUX), then the control signal of this MUX is used.


yrabbit:
I am not enthusiastic about this function - this is a terrible look in the implementation and I would very much like such a function to be part of the API.

https://github.com/yrabbit/nextpnr/blob/05ff80a24eec8b2305a672a4c0d4977d41be3201/himbaechel/uarch/gowin/gowin_utils.h#L41

And I must notice that Xilinx is also spying :)
https://github.com/YosysHQ/nextpnr/blob/0d5d32951c2f2100e436a4fbaf82d23e645ae4ff/himbaechel/uarch/xilinx/xilinx.cc#L326

gatecat:
I don't think this is a massive problem myself, in general if you're normalising wires in the first place then you are also not really dealing with canonical WireIds (there's a general avoidance of using WireIds for non-canonical identifiers).

However, could `get_dhcen_bel` perhaps analyse connectivity (e.g. following pips up/downhill and then using getWireBelPins) instead of doing the wire normalization thing?

yrabbit:
Of course, I can think about making these wires as DHCEN pins, only now it is not attached to them - it only “permits/prohibits” the passage of the signal through them. Does non -existent networks make sense? Hm...

Or I can turn the names of the wires from the very beginning into nomalized right in gowin_arch_gen.py. There is probably a normalization function there, I need to look it up, I don’t remember so offhand.

PR #1357, opened 2024-09-02
Gowin. Implement the UserFlash primitive

yrabbit:
Some Gowin chips have embedded flash memory accessible from the fabric. Here we add primitives that allow access to this memory.


himbaechel/uarch/gowin/pack.cc
```diff
@@ -3072,6 +3076,89 @@ struct GowinPacker
...
+                }
+            }
+            // add invertor
+            int lut_idx = 0;
+            auto add_inv = [&](IdString port, PortType port_type) {
+                if (!port_used(&ci, port)) {
+                    return;
+                }
+
+                CellInfo *lut = ctx->createCell(create_aux_name(ci.name, lut_idx, "_lut$"), id_LUT4);
```
gatecat, line 3131: creating a cell while also iterating over the cells map is not permitted - usually the best thing to do is build up a vector of user flash cells and then iterate over that to do the actual task of packing with creating cells, etc

yrabbit: My mistake. How's that?

PR #1359, opened 2024-09-06
timing: Move towards DelayPairs for timing analysis

rowanG077:
My plan is to first implement hold time violations report. This will happen in the next PR. This PR is only moving more things in the timing engine and the reporting to `DelayPair`. Once that is done I can use the `DelayPairs` to implement a hold time violation report.

After the hold time reports I will add false path min/max delay using the same mechanism. 

TODO:

- [x] Run performance difference on realistic design.
- [x] Verify same timing output

### DIFF
--- a/common/kernel/nextpnr_types.h
+++ b/common/kernel/nextpnr_types.h
@@ -80,7 +80,7 @@ struct PortRef
 // minimum and maximum delay
 struct DelayPair
 {
-    DelayPair(){};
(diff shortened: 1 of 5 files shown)


rowanG077:
Ran a 20K lut design through it 5 times. There seems to be no runtime change or it is drowned out by measurement error.

Timing information reported was the same. I compared the detailed net timings in two json reports using the same seed,

Issue #1043, opened 2022-11-18
Disable global promotion for ECP5


SteveCoul:
For nextpnr-ice40 we can disable promotion of nets to global - there is no such option for ECP5, could this be added?

gatecat:
It should be possible to disable it individually with a `(* noglobal *)` attribute on the net of concern, if that helps?

PR #1360, opened 2024-09-06
Add clock skew and hold time checks 

rowanG077:
Based of https://github.com/YosysHQ/nextpnr/pull/1359

### DIFF
--- a/common/kernel/nextpnr_types.h
+++ b/common/kernel/nextpnr_types.h
@@ -447,11 +447,11 @@ struct CriticalPath
     // Clock pair
     ClockPair clock_pair;
     // Total path delay
+    // if sum < 0 this is a hold/min violation
+    // if delay.maxDelay() >= max_delay this is a setup/max violation
     DelayPair delay;
 
-    // if delay.minDelay() < bound.minDelay() then this is a hold violation
-    // if delay.maxDelay() > bound.maxDelay() then this is a setup violation
-    DelayPair bound;
+    delay_t max_delay;
 
     // Individual path segments
     std::vector<Segment> segments;
--- a/common/kernel/nextpnr_types.h
+++ b/common/kernel/nextpnr_types.h
@@ -78,6 +78,16 @@ struct PortRef
     IdString port;
 };
 
+// Zero checking which works regardless if delay_t is floating or integer
+inline bool is_zero_delay(delay_t delay)
+{
+    if constexpr (std::is_floating_point<delay_t>::value) {
+        return std::fpclassify(delay) == FP_ZERO;
+    } else {
+        return delay == 0;
+    }
+}
+
 // minimum and maximum delay
 struct DelayPair
 {
--- a/common/kernel/nextpnr_types.h
+++ b/common/kernel/nextpnr_types.h
@@ -30,7 +30,6 @@
 #include "archdefs.h"
 #include "hashlib.h"
 #include "indexed_store.h"
-#include "log.h"
 #include "nextpnr_base_types.h"
 #include "nextpnr_namespaces.h"
 #include "property.h"
@@ -201,7 +200,7 @@ enum PortType
     case PORT_INOUT:
         return "PORT_INOUT";
     default:
-        log_error("Impossible PortType");
+        NPNR_ASSERT_FALSE("Impossible PortType");
     }
 }
 
@@ -250,7 +249,7 @@ enum TimingPortClass
     case TMG_IGNORE:
         return "TMG_IGNORE";
     default:
-        log_error("Impossible TimingPortClass");
+        NPNR_ASSERT_FALSE("Impossible TimingPortClass");
     }
 }
 
@@ -268,7 +267,7 @@ enum ClockEdge
     case FALLING_EDGE:
         return "FALLING_EDGE";
     default:
-        log_error("Impossible ClockEdge");
+        NPNR_ASSERT_FALSE("Impossible ClockEdge");
     }
 }
 
@@ -438,7 +437,7 @@ struct CriticalPath
             case Type::HOLD:
                 return "hold";
             default:
-                log_error("Impossible Segment::Type");
+                NPNR_ASSERT_FALSE("Impossible Segment::Type");
             }
         }
 
--- a/common/kernel/nextpnr_types.h
+++ b/common/kernel/nextpnr_types.h
@@ -398,35 +398,35 @@ struct CriticalPath
         // Segment type
         enum class Type
         {
-            CLK2CLK,  // Clock to clock delay
-            CLK_SKEW, // Clock skew
-            CLK_TO_Q, // Clock-to-Q delay
-            SOURCE,   // Delayless source
-            LOGIC,    // Combinational logic delay
-            ROUTING,  // Routing delay
-            SETUP,    // Setup time in sink
-            HOLD      // Hold time in sink
+            CLK_TO_CLK, // Clock to clock delay
+            CLK_SKEW,   // Clock skew
+            CLK_TO_Q,   // Clock-to-Q delay
+            SOURCE,     // Delayless source
+            LOGIC,      // Combinational logic delay
+            ROUTING,    // Routing delay
+            SETUP,      // Setup time in sink
+            HOLD        // Hold time in sink
         };
 
         [[maybe_unused]] static const std::string type_to_str(Type typ)
         {
             switch (typ) {
-            case Type::CLK2CLK:
-                return "CLK2CLK";
+            case Type::CLK_TO_CLK:
+                return "clk-to-clk";
             case Type::CLK_SKEW:
-                return "CLK_SKEW";
+                return "clk-skew";
             case Type::CLK_TO_Q:
-                return "CLK_TO_Q";
+                return "clk-to-q";
             case Type::SOURCE:
-                return "SOURCE";
+                return "source";
             case Type::LOGIC:
-                return "LOGIC";
+                return "logic";
             case Type::ROUTING:
-                return "ROUTING";
+                return "routing";
             case Type::SETUP:
-                return "SETUP";
+                return "setup";
             case Type::HOLD:
-                return "HOLD";
+                return "hold";
             default:
                 log_error("Impossible Segment::Type");
             }
--- a/common/kernel/nextpnr_types.h
+++ b/common/kernel/nextpnr_types.h
@@ -30,6 +30,7 @@
 #include "archdefs.h"
 #include "hashlib.h"
 #include "indexed_store.h"
+#include "log.h"
 #include "nextpnr_base_types.h"
 #include "nextpnr_namespaces.h"
 #include "property.h"
@@ -180,6 +181,20 @@ enum PortType
     PORT_INOUT = 2
 };
 
+[[maybe_unused]] static const std::string portType_to_str(PortType typ)
+{
+    switch (typ) {
+    case PORT_IN:
+        return "PORT_IN";
+    case PORT_OUT:
+        return "PORT_OUT";
+    case PORT_INOUT:
+        return "PORT_INOUT";
+    default:
+        log_error("Impossible PortType");
+    }
+}
+
 struct PortInfo
 {
     IdString name;
@@ -203,7 +218,7 @@ enum TimingPortClass
     TMG_IGNORE,          // Asynchronous to all clocks, "don't care", and should be ignored (false path) for analysis
 };
 
-static std::string tmgPortClass_to_str(TimingPortClass tmg_class)
+[[maybe_unused]] static const std::string timingPortClass_to_str(TimingPortClass tmg_class)
 {
     switch (tmg_class) {
     case TMG_CLOCK_INPUT:
@@ -225,7 +240,7 @@ static std::string tmgPortClass_to_str(TimingPortClass tmg_class)
     case TMG_IGNORE:
         return "TMG_IGNORE";
     default:
-        return "TMG_BOOM";
+        log_error("Impossible TimingPortClass");
     }
 }
 
@@ -234,6 +249,18 @@ enum ClockEdge
     RISING_EDGE,
     FALLING_EDGE
 };
+
+[[maybe_unused]] static const std::string clockEdge_to_str(ClockEdge edge)
+{
+    switch (edge) {
+    case RISING_EDGE:
+        return "RISING_EDGE";
+    case FALLING_EDGE:
+        return "FALLING_EDGE";
+    default:
+        log_error("Impossible ClockEdge");
+    }
+}
 
 struct TimingClockingInfo
 {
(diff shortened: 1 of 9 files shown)


gatecat:
One thing that seems to be missing here is clock delay/skew analysis. As it is inherently unlikely to have a hold time failure without clock skew (because in almost all cases, the clock to out time of a flipflop plus minimum routing is more than the hold time requirement of a DFF input), I think it would be good to do that at the same time.

However, it will require some consideration about performance, because this does add a bit of extra runtime to do the clock side analysis. Probably okay, but worth benchmarking and considering an option to disable it too.

rowanG077:
That's true. I'm in the process of adding it.

The problem I'm facing is that global nets on ECP5 (which I use as my development and testing target since I know most about it) returns zero due to:

https://github.com/YosysHQ/nextpnr/blob/50bd8d09b07d8e62959ae299177310c2e7f9e4e9/common/kernel/context.cc#L117-L123

If I comment that ifdef I do get delays for the clocks but they seem high to me for a clock net. Is this simply a case of ECP5 not having this information?

gatecat:
Yeah, unfortunately so, due to some questionable design decisions at the time they also aren't routed in a way that makes this easy to implement, either...

gatecat:
Only ECP5, although as this is not a feature we've had in the past, I won't 100% guarantee they're actually accurate across all architectures. At a minimum they should be for iCE40, although its global networks are good enough and the device small enough that there's no skew (to test clock skew/hold analysis, you could always disable the use of them and then see a bunch of skew through the globals).

rowanG077:
Can I also add an option to disable global promotion for ECP5 like for ice40? I rather keep my test scripts the same. This will also close: https://github.com/YosysHQ/nextpnr/issues/1043

rowanG077:
I think this is now in a reviewable state. I initially wanted store the clocks delay from all source FFs to sink FFs but this would become large and slow. I now moved the arrival and required times all in reference to the clock source. I think I have considered all cases. 

I did not yet benchmark. I will focus on that tomorrow. But I think the solution I choose shouldn't be very heavy. In the hot spots it's just an add/subtract more. What is more slowly now is the `max_delay_by_domain_pairs` since it now needs to walk the critical path. But since that is only used for reporting I'm fine with that.

rowanG077:
I also changed the timing report to include the type of segment it is. 

```
Info: Critical path report for clock 'clk$TRELLIS_IO_IN' (posedge -> posedge):
Info:       type  curr  total
Info:   clk-skew  52.00  52.00  Net clk$TRELLIS_IO_IN (15,14) -> (16,14)
Info:                           Sink out_TRELLIS_FF_Q_1.CLK
Info:                           Defined in:
Info:                             tests/simple.v:2.16-2.19
Info:   clk-to-q  0.52  52.53  b_reg_TRELLIS_FF_Q_1.Q
Info:    routing  0.00  52.53  Net b_reg[2] (15,14) -> (15,14)
Info:                           Sink b_reg_TRELLIS_FF_Q_1.Q
Info:                           Defined in:
Info:                             /nix/store/2bv02r255w3xgvcl00z1365zw6djv58f-yosys-0.38/bin/../share/yosys/ecp5/cells_map.v:108.23-108.24
Info:     source  0.00  52.53  b_reg_TRELLIS_FF_Q_1.Q
Info:    routing  1.10  53.62  Net b_reg[2] (15,14) -> (16,14)
Info:                           Sink out_TRELLIS_FF_Q_DI_PFUMX_Z_BLUT_LUT4_Z.B
Info:                           Defined in:
Info:                             /nix/store/2bv02r255w3xgvcl00z1365zw6djv58f-yosys-0.38/bin/../share/yosys/ecp5/cells_map.v:108.23-108.24
Info:      logic  0.40  54.02  out_TRELLIS_FF_Q_DI_PFUMX_Z_BLUT_LUT4_Z.OFX
Info:    routing  0.38  54.40  Net out_TRELLIS_FF_Q_DI[2] (16,14) -> (16,14)
Info:                           Sink out_TRELLIS_FF_Q_1.M
Info:                           Defined in:
Info:                             tests/simple.v:13.16-13.29
Info:                             /nix/store/2bv02r255w3xgvcl00z1365zw6djv58f-yosys-0.38/bin/../share/yosys/techmap.v:270.26-270.27
Info:      setup  0.00  54.40  out_TRELLIS_FF_Q_1.M
Info: 0.93 ns logic, 53.47 ns routing
```

This is mainly done to make it more clear. Especially now with this patch there can be negative segments.

common/kernel/nextpnr_types.h
```diff
@@ -180,6 +181,20 @@ enum PortType
     PORT_INOUT = 2
 };
 
+[[maybe_unused]] static const std::string portType_to_str(PortType typ)
```
rowanG077, line 184: These and other enum to string function where mainly used during development. In the past I have often defined them. And now had to again. I would prefer to keep them even if currently unused because they can be handy during development.

common/kernel/timing.cc
```diff
@@ -566,12 +568,18 @@ void TimingAnalyser::walk_forward()
...
             if (sp.second != IdString()) {
                 // clocked startpoints have a clock-to-out time
                 for (auto &fanin : pd.cell_arcs) {
                     if (fanin.type == CellArc::CLK_TO_Q && fanin.other_port == sp.second) {
-                        init_arrival = init_arrival + fanin.value.delayPair();
+                        init_arrival += fanin.value.delayPair();
+                        // Include the clock delay if clock_skew analysis is enabled
+                        if (with_clock_skew) {
+                            auto clock_delay = ports.at(CellPortKey(sp.first.cell, fanin.other_port)).route_delay;
+                            clock_delay.min_delay = clock_delay_fac * clock_delay.min_delay;
```
rowanG077, line 579: The `clock_delay_fac` is just for debugging for me. Once this PR is approved I will remove it. It just makes seeing what delay is the clock skew and what isn't much easier in the reports :).

common/kernel/timing.cc
```diff
@@ -616,20 +626,28 @@ void TimingAnalyser::walk_backward()
...
+                            auto clock_delay = ports.at(CellPortKey(ep.first.cell, fanin.other_port)).route_delay;
+                            clock_delay.min_delay = clock_delay_fac * clock_delay.min_delay;
+                            clock_delay.max_delay = clock_delay_fac * clock_delay.max_delay;
+                            init_required += clock_delay;
+                        }
+                        init_required.min_delay -= fanin.value.maxDelay();
+                    }
                     if (fanin.type == CellArc::HOLD && fanin.other_port == ep.second)
-                        init_setuphold.max_delay -= fanin.value.maxDelay();
+                        init_required.max_delay += fanin.value.maxDelay();
```
rowanG077, line 646: I think the hold time had the wrong sign.

Hold slack is calculated like `arr.value.minDelay() - req.value.maxDelay() + clock_to_clock`. If you say here that hold is negative then it becomes `arr.value.minDelay() + hold_time - <rest of required time> + clock_to_clock` which is not correct.

common/kernel/timing.cc
```diff
@@ -915,23 +1065,29 @@ CriticalPath TimingAnalyser::build_critical_path_report(domain_id_t domain_pair,
...
+        } else {
+            seg_logic.type = CriticalPath::Segment::Type::HOLD;
+            seg_logic.delay -= ep_clk_info.hold;
+        }
+        seg_logic.from = std::make_pair(prev_cell->name, prev_port);
         seg_logic.to = seg_logic.from;
         seg_logic.net = IdString();
         report.segments.push_back(seg_logic);
+
+        report.bound.min_delay = ep_clk_info.hold.min_delay;
```
rowanG077, line 1090: Oops this is wrong now. A hold time violation has negative delay now with the negative segments.

common/kernel/timing.cc
gatecat: Not directly relevant here but some related food for thought - if you use minimum clock/routing delays for the arrival time and maximum for the required time, this is likely to be unnecessarily pessimistic, particularly on the clock side where it is probably that some of the routing is shared so it can't physically have two different delays.

But in general, nextpnr doesn't really have a clear concept of whether min/max delays are overall or for a specific corner (iirc Vivado has fast-min, fast-max, slow-min, slow-max for example which might actually be a better use of the four values DelayQuad than seperate rising/falling delays). This means that as well as shared routing that's guaranteed to have the same delay; as the entire chip will be at essentially the same temperature, voltage and process corner; the difference between delays will never be as high as min vs max.

To avoid this pessimism I think it might be better to just use one corner of routing delays (e.g. min delays everywhere) for hold analysis. Even if this is slightly optimistic it's unlikely for any of the FPGAs we deal with that on-chip variation is going to be that high (but OCV is a field in itself...)

rowanG077: Currently it has no impact since the route delay is just a `delay_t` so the min/max corners are the same. But a comment there would be nice if this ever changes.

rowanG077:
I did 100 runs over the night with 18k lut design current master vs this branch:

```
Benchmark 1: make pnr FORCE_PNR=1 NEXTPNR_PATH="/home/rowan.goemans/Documents/engineering/nextpnr-orig/build/nextpnr-ecp5"
  Time (mean ± σ):     60.214 s ±  4.609 s    [User: 66.815 s, System: 0.310 s]
  Range (min … max):   49.514 s … 82.134 s    100 runs
 
Benchmark 2: make pnr FORCE_PNR=1 NEXTPNR_PATH="/home/rowan.goemans/Documents/engineering/nextpnr/build/nextpnr-ecp5"
  Time (mean ± σ):     60.108 s ±  4.370 s    [User: 66.670 s, System: 0.305 s]
  Range (min … max):   51.732 s … 76.546 s    100 runs
```

Doesn't seem to make an appreciable difference as expected. clock skew analysis was enabled during routing and final reporting. Since this doesn't have a real performance impact I would say  just keep it on during routing and final reporting.

rowanG077:
I one of my larger designs for an ECP5 with disabling global routing I do get hold time violations, ex:

```
ERROR: Hold time violation for clock 'posedge clk50':
Info:       type curr  total name
Info:   clk-skew -0.76 -0.76 Net clk50 (57,11) -> (57,9)
Info:                          Sink matrix_inst.TopEntity_topEntity_packetHandler_result_19.TopEntity_topEntity_dmaAsyncStream_result_11.TopEntity_topEntity_asyncPacketFIFO_ds2.TopEntity_topEntity_trueDualPortBlockRamWrapper_1_c$case_scrut.mem.0.5$DPRAM_COMB2.WCK
Info:                          Defined in:
Info:                               verilog/TopEntity.topEntity/matrix.v:129.16-129.25
Info:   clk-to-q  0.52 -0.24 Source matrix_inst.TopEntity_topEntity_packetHandler_result_19.TopEntity_topEntity_dmaAsyncStream_result_11.TopEntity_topEntity_asyncPacketFIFO_ds2.result_5_TRELLIS_FF_Q_9.Q
Info:    routing  0.51  0.27 Net matrix_inst.TopEntity_topEntity_packetHandler_result_19.TopEntity_topEntity_dmaAsyncStream_result_11.a1[23] (57,11) -> (57,9)
Info:                          Sink matrix_inst.TopEntity_topEntity_packetHandler_result_19.TopEntity_topEntity_dmaAsyncStream_result_11.TopEntity_topEntity_asyncPacketFIFO_ds2.TopEntity_topEntity_trueDualPortBlockRamWrapper_1_c$case_scrut.mem.0.5$RAMW_SLICE.D1
Info:                          Defined in:
Info:                               verilog/TopEntity.topEntity/TopEntity_topEntity_dmaAsyncStream.v:25.16-25.24
Info:      logic  0.00  0.27 Source matrix_inst.TopEntity_topEntity_packetHandler_result_19.TopEntity_topEntity_dmaAsyncStream_result_11.TopEntity_topEntity_asyncPacketFIFO_ds2.TopEntity_topEntity_trueDualPortBlockRamWrapper_1_c$case_scrut.mem.0.5$RAMW_SLICE.WDO2
Info:    routing  0.08  0.35 Net matrix_inst.TopEntity_topEntity_packetHandler_result_19.TopEntity_topEntity_dmaAsyncStream_result_11.TopEntity_topEntity_asyncPacketFIFO_ds2.TopEntity_topEntity_trueDualPortBlockRamWrapper_1_c$case_scrut.mem.0.5$RAMW_SLICE$conn$WDO2 (57,9) -> (57,9)
Info:                          Sink matrix_inst.TopEntity_topEntity_packetHandler_result_19.TopEntity_topEntity_dmaAsyncStream_result_11.TopEntity_topEntity_asyncPacketFIFO_ds2.TopEntity_topEntity_trueDualPortBlockRamWrapper_1_c$case_scrut.mem.0.5$DPRAM_COMB2.WD
Info:       hold -0.30  0.06 Source matrix_inst.TopEntity_topEntity_packetHandler_result_19.TopEntity_topEntity_dmaAsyncStream_result_11.TopEntity_topEntity_asyncPacketFIFO_ds2.TopEntity_topEntity_trueDualPortBlockRamWrapper_1_c$case_scrut.mem.0.5$DPRAM_COMB2.WD
Info: 0.23 ns logic, -0.17 ns routing
``` 
So it's nice to see this now reported correctly.

common/kernel/timing_log.cc
```diff
@@ -168,7 +174,27 @@ static void log_crit_paths(const Context *ctx, TimingResult &result)
         log_info("Critical path report for cross-domain path '%s' -> '%s':\n", start.c_str(), end.c_str());
         print_path_report(report);
     }
-};
+
+    // Min delay violated paths
+    // Show maximum of 10
```
gatecat, line 179: Hm, compared to max delay violations where we display one per clock (pair) this seems like a behaviour change. If we are going to show a maximum of 10, we should probably be sorting to show the worst 10, which I might have missed, but I can't see anywhere.

rowanG077: This is true. I also thought to show the worst per clock pair the same way we do for max delay. The other question remaining is whether we should add the hold slack to slack histogram.

rowanG077:
I'm now confident this works correctly. The hold slack now matches the path delay reported which was not the case due to min/max bounds of a delay pair not being chosen correctly.

common/kernel/nextpnr_types.h
```diff
@@ -180,6 +191,20 @@ enum PortType
...
+{
+    switch (typ) {
+    case PORT_IN:
+        return "PORT_IN";
+    case PORT_OUT:
+        return "PORT_OUT";
+    case PORT_INOUT:
+        return "PORT_INOUT";
+    default:
+        log_error("Impossible PortType");
```
gatecat, line 204: These functions are fine, but `NPNR_ASSERT_FALSE("Impossible PortType");` would be more idiomatic nextpnr code than `log_error`, also to avoid bringing `log.h` into here.

common/kernel/timing_log.cc
```diff
@@ -168,7 +165,27 @@ static void log_crit_paths(const Context *ctx, TimingResult &result)
...
+    if (num_min_violations > 0) {
+        log_break();
+        log_info("%zu Hold/min time violations (showing 10 worst paths):\n", num_min_violations);
+        for (size_t i = 0; i < std::min((size_t)10, num_min_violations); ++i) {
+            auto &report = result.min_delay_violations.at(i);
+            log_break();
+            std::string start = clock_event_name(ctx, report.clock_pair.start);
+            std::string end = clock_event_name(ctx, report.clock_pair.end);
+            if (report.clock_pair.start == report.clock_pair.end) {
+                log_nonfatal_error("Hold/min time violation for clock '%s':\n", start.c_str());
```
gatecat, line 181: we should follow https://github.com/YosysHQ/nextpnr/blob/4b63b1115e01590e1ca5ac7a7f9c17811455387e/common/kernel/timing_log.cc#L197-L199 and make this a warning too if `timing/allowFail` is set

rowanG077:
Comments have been handled.

Issue #1361, opened 2024-09-09
Himbaechel for Xilinx target : Unable to place cell DSP48E1


marzoul:
Hi,

When experimenting with the Himbaechel variant for Xilinx Zynq target, nextpnr-himbaechel fails very early with the following message:
`ERROR: Unable to place cell 'user_inst.mul_i.gen_dsp48e1_inst', no BELs remaining to implement cell type 'DSP48E1'`

Indeed the test design instantiates a DSP48E1 cell.
But it seems that the supported cells are named `DSP48E1_DSP48E1` according to the log :
`Info: 	     DSP48E1_DSP48E1:       0/     80     0%`

I guess that BRAMs will have the same issue.

Note that this does not happen with experimental nextpnr-xilinx, this tool implements a dedicated mapping according to its log, it seems this pass is missing in Himbaechel for Xilinx :
```
Info: Packing DSPs..
Info:     Created 1 DSP48E1_DSP48E1 cells from:
Info:              1x DSP48E1
```
Tested with latest commit as of now, compiled on Archlinux.

Archive of the reproducer (with source verilog generated with ghdl, the json produced by Yosys, the netpnr-himbaechel log, and the Makefile to reproduce) : https://cloud.univ-grenoble-alpes.fr/s/pTHEn8ngLgSTpxg

Regards,
Adrien

marzoul:
This issue is now resolved.

PR #1363, opened 2024-09-10
Himbaechel xilinx : Add support of DSP packing for 7-series

marzoul:
Hi,

This pull request imports the DSP packing implementation from project nextpnr-xilinx.
https://github.com/gatecat/nextpnr-xilinx/
Only minimal changes done to the main code to make it fit for Himbaechel code base.

This should address issue https://github.com/YosysHQ/nextpnr/issues/1361
(but not make it work altogether as routing fails later in the flow).

One point to highlight is the addition of `rbegin()` to class `indexed_store`, because as far as I understand the use of `end()` will always generate an out-of-range exception because of calling `at()` with an index equal to the vector size.
So if existing code makes use of `end()` should probably be inspected.

Regards,
Adrien


common/kernel/indexed_store.h
```diff
@@ -209,6 +209,7 @@ template <typename T> class indexed_store
         return it;
     }
     iterator end() { return iterator{this, int32_t(slots.size())}; }
+    iterator rbegin() { return iterator{this, int32_t(slots.size()-1)}; }
```
gatecat, line 212: Hmm, this won't actually work if the last slot isn't occupied (because an item was removed). I think the original code was just an error, and derefencing normal `begin` instead of `rbegin` or `end` should work as it only cares about the one-user case anyway.

himbaechel/uarch/xilinx/pack_dsp_xc7.cc
```diff
@@ -0,0 +1,165 @@
+/*
+ *  nextpnr -- Next Generation Place and Route
+ *
+ *  Copyright (C) 2020  David Shah <dave@ds0.me>
```
gatecat, line 4: Please update this to Myrtle Shah <gatecat@ds0.me>

common/kernel/indexed_store.h
marzoul: Indeed `begin` is better.
But if the last slot may be unoccupied, the first slot may be unoccupied in a similar way, so we should use some other getter that returns the first occupied slot (not yet looked at the code for that).

gatecat: Yup, that's exactly how `begin` behaves - it will always return the first occupied slot, if there is one (which is guaranteed by the previous check in the case we're dealing with here)

marzoul:
Changes done and tested, it still works :-)

himbaechel/uarch/xilinx/pack_dsp_xc7.cc
```diff
@@ -0,0 +1,165 @@
...
+            auto vcc_attr = ctx->id("DSP_VCC_PINS");
+
+            auto gnd_pins = str_or_default(ci->attrs, gnd_attr, "");
+            auto vcc_pins = str_or_default(ci->attrs, vcc_attr, "");
+
+            for (auto &port : ci->ports) {
+                std::string n = port.first.str(ctx);
+
+                if (boost::starts_with(n, "ACIN") || boost::starts_with(n, "BCIN") || boost::starts_with(n, "PCIN")) {
+                    if (port.second.net == nullptr)
```
gatecat, line 112: this should be changed to `port.second.net == nullptr || port.second.net->users.empty()` as we also want to skip the net if it's just dangling

marzoul:
I pushed the extra checks.
Now the routing starts, but does not succeed.

```
Info: Routing..
Info: Setting up routing queue.
Info: Routing 1919 arcs.
Info:            |   (re-)routed arcs  |   delta    | remaining|       time spent     |
Info:    IterCnt |  w/ripup   wo/ripup |  w/r  wo/r |      arcs| batch(sec) total(sec)|
Warning: Failed to find a route for arc 100 of net $PACKER_GND_NET.
Info: Running post-routing legalisation...
ERROR: Expecting string value but got integer 1380270932.
1 warning, 1 error
```

Worth noting is that it is related to net `$PACKER_GND_NET` that is explicitly used in DSP packing.
But this is beyond what I can analyze.

himbaechel/uarch/xilinx/pack_dsp_xc7.cc
```diff
@@ -0,0 +1,165 @@
...
+            auto vcc_attr = ctx->id("DSP_VCC_PINS");
+
+            auto gnd_pins = str_or_default(ci->attrs, gnd_attr, "");
+            auto vcc_pins = str_or_default(ci->attrs, vcc_attr, "");
+
+            for (auto &port : ci->ports) {
+                std::string n = port.first.str(ctx);
+
+                if (boost::starts_with(n, "ACIN") || boost::starts_with(n, "BCIN") || boost::starts_with(n, "PCIN")) {
+                    if (port.second.net == nullptr || port.second.net->users.empty())
```
gatecat, line 112: This is the cause for the failure - the users.empty() check isn't correct or needed here, only in the one case I commented on when finding cascades

marzoul:
Found the option `--debug-router`..
The routing issue (which is just a warning) as about that arc :
```
Routing constant arc 100 on net $PACKER_GND_NET (1905 arcs total):
  value ... GND
  sink ..... X49Y57/DSP48_X0Y1.CARRYCASCIN
  total number of visited nodes: 5
  no route found for this arc
Warning: Failed to find a route for arc 100 of net $PACKER_GND_NET.
```
Indeed `CARRYCASCIN` input is not reachable with routing.
But the crash happens after :
```
Info: Running post-routing legalisation...
ERROR: Expecting string value but got integer 1380270932.
1 warning, 1 error
```
According to gdb stacktrace, it is indeed linked to DSPs:
```
(gdb) bt
#0  nextpnr_himbaechel::log_error (format=0x5555557319c8 "Expecting string value but got integer %d.\n") at /usr/src/debug/nextpnr-himbaechel-git/nextpnr-himbaechel/common/kernel/log.cc:184
#1  0x00005555555c8cc6 in nextpnr_himbaechel::str_or_default<nextpnr_himbaechel::IdString> (ct=..., key=..., def="DIRECT") at /usr/src/debug/nextpnr-himbaechel-git/nextpnr-himbaechel/common/kernel/util.h:65
#2  0x00005555556b29ce in nextpnr_himbaechel::(anonymous namespace)::FasmBackend::write_dsp_cell (this=0x7fffffffad00, ci=<optimized out>) at /usr/src/debug/nextpnr-himbaechel-git/nextpnr-himbaechel/himbaechel/uarch/xilinx/fasm.cc:1524
#3  nextpnr_himbaechel::(anonymous namespace)::FasmBackend::write_ip (this=0x7fffffffad00) at /usr/src/debug/nextpnr-himbaechel-git/nextpnr-himbaechel/himbaechel/uarch/xilinx/fasm.cc:1634
#4  nextpnr_himbaechel::(anonymous namespace)::FasmBackend::write_fasm (this=0x7fffffffad00) at /usr/src/debug/nextpnr-himbaechel-git/nextpnr-himbaechel/himbaechel/uarch/xilinx/fasm.cc:1648
#5  nextpnr_himbaechel::XilinxImpl::write_fasm (this=<optimized out>, filename=...) at /usr/src/debug/nextpnr-himbaechel-git/nextpnr-himbaechel/himbaechel/uarch/xilinx/fasm.cc:1661
#6  0x00005555556e3af7 in nextpnr_himbaechel::XilinxImpl::postRoute (this=0x5555557b0190) at /usr/src/debug/nextpnr-himbaechel-git/nextpnr-himbaechel/himbaechel/uarch/xilinx/xilinx.cc:273
#7  0x000055555564e16b in nextpnr_himbaechel::Arch::route (this=0x5555557b3860) at /usr/src/debug/nextpnr-himbaechel-git/nextpnr-himbaechel/himbaechel/arch.cc:243
#8  0x00005555556fef69 in nextpnr_himbaechel::CommandHandler::executeMain(std::unique_ptr<nextpnr_himbaechel::Context, std::default_delete<nextpnr_himbaechel::Context> >) [clone .constprop.0] (this=this@entry=0x7fffffffc590, 
    ctx=std::unique_ptr<nextpnr_himbaechel::Context> = {...}) at /usr/src/debug/nextpnr-himbaechel-git/nextpnr-himbaechel/common/kernel/command.cc:687
#9  0x00005555555917e3 in nextpnr_himbaechel::CommandHandler::exec (this=0x7fffffffc590) at /usr/src/debug/nextpnr-himbaechel-git/nextpnr-himbaechel/common/kernel/command.cc:759
#10 main (argc=<optimized out>, argv=<optimized out>) at /usr/src/debug/nextpnr-himbaechel-git/nextpnr-himbaechel/himbaechel/main.cc:107
(gdb) 
```
The code at fasm.cc:1524 :
`auto ainput = str_or_default(ci->params, ctx->id("A_INPUT"), "DIRECT");`

marzoul:
Maybe `ci->params` stores the string values as string IDs I reckon ?
So the variant `int_or_default` could be more appropriate for these operations, the stored value is just compared anyway right after.

himbaechel/uarch/xilinx/pack_dsp_xc7.cc
```diff
@@ -0,0 +1,165 @@
...
+            log_error("User %s of net %s is not a DSP block, but %s",
+                user.cell->name.c_str(ctx), ni->name.c_str(ctx), user.cell->type.c_str(ctx));
+    };
+
+    // see if any cascade outputs are connected
+    for (auto port : current_cell->ports) {
+        if (!boost::contains(port.first.str(ctx), "COUT")) continue;
+        NetInfo *cout_net = port.second.net;
+
+        if (cout_net == nullptr) continue;
```
gatecat, line 48: Sorry, I think my previous comment got misunderstood - this is the only place where the empty check is needed, it should be removed down below

gatecat:
No, strings that come from Yosys should be stored as strings in the JSON and therefore kept as string type property values. 

The problem is that GHDL doesn't preserve them as strings, so you get something like this in the Verilog/JSON `.A_INPUT(48'b010001000100100101010010010001010100001101010100)` - nextpnr can't handle this.

With a normal Verilog flow, this should work fine.

marzoul:
I did not notice this ghdl generation issue. Thank you.
With a little post-processing of generated verilog I can have the types expected by nextpnr, that'll do for now.

But nextpnr expects MASK and PATTERN to strings, where the reference instantiation templates from ug953 looks more like a bit vector. So the format generated by ghdl should be a better fit ?
```
    .MASK(48'b001111111111111111111111111111111111111111111111),
    .PATTERN(48'b000000000000000000000000000000000000000000000000),
```

Routing still fails, though. But this time no exploitable stack trace. So I guess the warning has to be fixed solidly first.
```
Warning: Failed to find a route for arc 100 of net $PACKER_GND_NET.
Info: Running post-routing legalisation...
ERROR: Routing design failed.
```

marzoul:
I added a commit to disconnect more cascaded input ports so these are not routed.
Routing now succeeds \o/

gatecat:
> But nextpnr expects MASK and PATTERN to strings, where the reference instantiation templates from ug953 looks more like a bit vector. So the format generated by ghdl should be a better fit ?

This sounds like a problem in nextpnr, they should be handled as bitvectors (probably, as these parameters aren't usually specified, this was never caught before).

marzoul:
The issue is now with cascaded DSPs, this is used for multiplications larger than one DSP.
Reproducer design is in public archive :
https://cloud.univ-grenoble-alpes.fr/s/NpQT72p49kR5mCb

common/place/placer_heap.cc
```diff
@@ -301,17 +301,24 @@ class HeAPPlacer
...
+        // Apply saved solution
         for (auto &sc : solution) {
             CellInfo *cell;
             BelId bel;
             PlaceStrength strength;
             std::tie(cell, bel, strength) = sc;
+            // In case no bel was found, emit a warning instead of continuing to a crash
+            // This helps to investigate and to devise a workaround in user design
+            if (bel == BelId()) {
+                log_warning("No bel found to bind cell '%s' of type '%s'\n", cell->name.c_str(ctx), cell->type.c_str(ctx));
```
gatecat, line 319: I don't think this should just be a warning

marzoul:
Disabling DSP clustering makes placement and routing succeed.
And it seems the current clustering of DSPs should be revised (see comments in commits).

It seems the results are valid solutions, though I can't guarantee that given my little knowledge of placer behaviour.
But at least, this it makes nextpnr produce results. I think this is valuable for both end users (perf metrics give directions to continue their design development), and nextpnr developers (large designs enable to work on performance of algorithms).

This no-clustering behavior also corresponds to how BRAMs are handled.

marzoul:
Finally, I pushed a commit to be more flexible about types of DSP parameters, for interoperability purposes.
Xilinx doc ug953 shows bitvector type for PATTERN and MASK, but yosys seems to generate strings only, and ghdl generates bitvectors.
Xilinx doc ug953 even shows inconsistent type for USE_DPORT.

gatecat:
Unfortunately I haven't had much time to look into this, but I'm somewhat surprised that just disabling clustering is enough to make this work - my understanding was the DSP cascade routing is dedicated, so if the DSPs aren't placed correctly then routing will fail. BRAMs are different because we don't ever actually use any of the functionality that requires dedicated cascade routing for those.

A "proper", albeit somewhat more complex solution, if I understand this correctly, would be to override `getClusterPlacement` to account for the gaps in the column structure and that DSPs don't always have fixed offsets.

marzoul:
Indeed I'm not saying that this is fixed, only that there is no crash on my experiment design.
Obtaining a correct placement by chance nice but not irrelevant. For my humble objective, potentially non-functional results is still better than crash because I can keep going.
A strong warning is certainly desirable.

Are you meaning that placement of cascaded BRAMs is not supposed to be handled properly ?

Some interesting words about cascaded BRAMs : I am testing also testing with 2 cascaded BRAMs, and perhaps also by chance these happen to be placed correctly (runtime not tested). Yosys does generate cascaded BRAMs for a 64k x 1b sync memories so this is easily reproducible (I can provide testcase if this is useful).
However, nextpnr-xilinx crashes in the placer code on cascaded BRAMs. I had to add more constant cascaded port disconnection in nextpnr-xilinx to make it work (pull request to come, for discussion).

marzoul:
Now I need some visibility : What should this PR feature to hope for integration ?
Adding a warning for cascaded DSPs is doable.
But in making the placer correctly support it could be future works (limited time and knowledge of intricacies of this code).

marzoul:
Finally found the root cause : The root DSP was not marked as being in a cluster !
It all seems functional now.

gatecat:
Nice, thanks! This was probably missed when porting the code to the new cluster API at some point.

PR #1367, opened 2024-09-12
ecp5: Allow disabling of global promotion

rowanG077:
Closes https://github.com/YosysHQ/nextpnr/issues/1043

And more easily allows me to continue working on clock skew analysis.


gatecat:
I don't think this needs the "experimental" qualifier, otherwise looks good

PR #1369, opened 2024-09-17
Fix handling of RNG seed

jthornblad:
Hi,

There are a couple of problems with the handling of the RNG seed.

First, the seed value used for the deterministic RNG is being truncated from 64 bits to 32 (int instead of uint64) when written to the header of the output JSON file.

When using the --seed option on the command line, the argument needs to be a signed value.

If using the --seed or --randomize-seed option, it is not the originally supplied or randomized seed value that is written to the output JSON file, but a value that has been run through the rngseed() function, which permutes the value before it is stored and then used for the output JSON file.

This patch fixes these problems.

Regards
Jonas

### DIFF
--- a/common/kernel/command.cc
+++ b/common/kernel/command.cc
@@ -343,7 +343,7 @@ po::options_description CommandHandler::getGeneralOptions()
     general.add_options()("json", po::value<std::string>(), "JSON design file to ingest");
     general.add_options()("write", po::value<std::string>(), "JSON design file to write");
     general.add_options()("top", po::value<std::string>(), "name of top module");
-    general.add_options()("seed", po::value<int>(), "seed value for random number generator");
+    general.add_options()("seed", po::value<uint64_t>(), "seed value for random number generator");
     general.add_options()("randomize-seed,r", "randomize seed value for random number generator");
 
     general.add_options()(
@@ -447,7 +447,7 @@ void CommandHandler::setupContext(Context *ctx)
     }
 
     if (vm.count("seed")) {
-        ctx->rngseed(vm["seed"].as<int>());
+        ctx->rngstate = vm["seed"].as<uint64_t>();
     }
 
     if (vm.count("threads")) {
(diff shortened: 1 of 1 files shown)


common/kernel/property.h
```diff
@@ -42,7 +42,7 @@ struct Property
     };
 
     Property();
-    Property(int64_t intval, int width = 32);
+    Property(int64_t intval, int width = 64);
```
gatecat, line 45: I'm concerned this might end up having unwanted consequences elsewhere - perhaps you could just instantiate it 64 bits wide in the one place where seed is being saved?

jthornblad:
Does this look ok?

common/kernel/command.cc
```diff
@@ -565,7 +565,8 @@ void CommandHandler::setupContext(Context *ctx)
 
     ctx->settings[ctx->id("arch.name")] = std::string(ctx->archId().c_str(ctx));
     ctx->settings[ctx->id("arch.type")] = std::string(ctx->archArgsToId(ctx->archArgs()).c_str(ctx));
-    ctx->settings[ctx->id("seed")] = ctx->rngstate;
+    Property seed_property(ctx->rngstate, 64);
+    ctx->settings[ctx->id("seed")] = seed_property;
```
gatecat, line 569: what about `ctx->settings[ctx->id("seed")] = Property(ctx->rngstate, 64);`? that's closer to the style we tend to use.

jthornblad:
Good suggestion, done.

gatecat:
Thanks!

jthornblad:
I see that the CI is failing for this patch. I'm not quite sure what the problem is, do you have any idea?

gatecat:
Looks like one of the tests is timing out (probably a latent instability)

rowanG077:
Input seed handling has changed which means some fixed seeds in the test suite were causing issues. https://github.com/YosysHQ/nextpnr-tests/pull/12 should fix it.

fontamsoc:
I am still seeing this problem as of:
$ nextpnr-ecp5 --version
"nextpnr-ecp5" -- Next Generation Place and Route (Version nextpnr-0.8-2-g7a3a43e1)

At the top of the generated config file, I have:
Info: Generated random seed: 3034085575128226198

But when I used as follow `nextpnr-ecp5 --seed 3034085575128226198`, I do not get the same run.

mbapplegate:
I am also seeing this issue using apio on:
apio raw -- nextpnr-ice40 --version
"nextpnr-ice40" -- Next Generation Place and Route (Version nextpnr-0.9-60-gb4da86ed)

PR #1370, opened 2024-09-20
apicula: add support for magic sip pins

pepijndevos:
While we have documented the sdram pins of some devices we lack the ability to automatically place them like the vendor tools do. For better compatibility and user friendliness this PR adds the ability to constrain these magic pin locations.

It seems to load the constraints but I have yet to test if sdram actually functions, which isn't super trivial.

### DIFF
--- a/himbaechel/uarch/gowin/cst.cc
+++ b/himbaechel/uarch/gowin/cst.cc
@@ -240,8 +240,31 @@ struct GowinCstReader
     }
 };
 
+
+void add_sip_constraints(Context *ctx, const Extra_package_data_POD *extra) {
+    for(auto cst : extra->cst) {
+        auto it = ctx->cells.find(IdString(cst.net));
+        if (it == ctx->cells.end()) {
+            log_info("Cell %s not found\n", IdString(cst.net).c_str(ctx));
+            continue;
+        }
+        Loc loc = Loc(cst.col, cst.row, cst.bel);
+        BelId bel = ctx->getBelByLocation(loc);
+        if (bel == BelId()) {
+            log_error("Pin not found.\n");
+        }
+        it->second->setAttr(IdString(ID_BEL), std::string(ctx->nameOfBel(bel)));
+    }
+}
+
 bool gowin_apply_constraints(Context *ctx, std::istream &in)
 {
+    // implicit constraints from SiP pins
+    const Extra_package_data_POD *extra = reinterpret_cast<const Extra_package_data_POD *>(ctx->package_info->extra_data.get());
+    if(extra != nullptr) {
+        add_sip_constraints(ctx, extra);
+    }
+
     GowinCstReader reader(ctx, in);
     return reader.run();
 }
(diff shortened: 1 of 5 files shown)


himbaechel/uarch/gowin/cst.cc
```diff
@@ -240,8 +240,31 @@ struct GowinCstReader
...
+        }
+        it->second->setAttr(IdString(ID_BEL), std::string(ctx->nameOfBel(bel)));
+    }
+}
+
 bool gowin_apply_constraints(Context *ctx, std::istream &in)
 {
+    // implicit constraints from SiP pins
+    const Extra_package_data_POD *extra = reinterpret_cast<const Extra_package_data_POD *>(ctx->package_info->extra_data.get());
+    if(extra != nullptr) {
```
gatecat, line 264: this won't actually work, because relptrs are relative so `get()` will add it's own location to the zero offset first. you'd need to add something like an `is_null()` or similar function to RelPtr that checks if the offset is zero (because a relptr is never going to be pointing to itself in any useful context, so zero is useful as a null offset value)

himbaechel/uarch/gowin/cst.cc
```diff
@@ -240,8 +240,31 @@ struct GowinCstReader
     }
 };
 
+
+void add_sip_constraints(Context *ctx, const Extra_package_data_POD *extra) {
```
gatecat, line 244: this can be `static void` or in an anonymous namespace

himbaechel/uarch/gowin/cst.cc
pepijndevos: The concerning part is that I wasn't sure how to check if a RelPtr is null, and found other cases that seemed to do this, but maybe they were subtly different.

You mean add a method to the RelPtr class that checks if the offset is 0?

gatecat: Those places are probably just wrong...

Yes, that's what I'd recommend doing.

pepijndevos: I'm not sure what I found before but it may have been a slightly different pattern that checks if a pointer **to** a RelPtr is nullptr.
```
    auto db_ptr = reinterpret_cast<const RelPtr<DatabasePOD> *>(get_chipdb(chipdb));
    if (db_ptr == nullptr)
        log_error("Failed to load chipdb '%s'\n", chipdb.c_str());
    db = db_ptr->get();
```

gatecat: Yeah, this isn't checking if the result of getting a RelPtr is nullptr, rather if a regular pointer to a RelPtr is nullptr (as it would be if mapping the database failed)

gatecat (approved):
LGTM

gatecat:
Does this still need to be a draft or are we okay merging this?

pepijndevos:
I still need to test it actually works

PR #1373, opened 2024-09-26
Add flag to restart place-and-route on failed target frequency

jthornblad:
* Add --restart-on-failed-target-frequency flag and logic to restart the place-and-route process if the target frequency is not achieved.

* This flag is intended to be used with the --randomize-seed option to generate a new random seed for every run. This can significantly improve the chances of achieving a higher clock frequency compared to using the default seed.

* Move print_net_source out of log_crit_paths() to remove the 'static' keyword, which could otherwise cause a segmentation fault if the context is changed.


rowanG077:
Personally I think this should be out of scope for nextpnr. And should be handled with an external script.

gatecat:
Yeah, I'm in agreement with @rowanG077

Anything in nextpnr itself should be finer-grained, like improving the ripup-and-reroute support in the router to locally improve the critical path.

PR #1375, opened 2024-10-01
Himbaechel xilinx : XDC improvements

marzoul:
Hi,

This pull request brings several improvements to ease the life of users.
- fix crashes
- more messages about unsupported or invalid constraints
- support multiple nets per command
- get_nets tries lowercase as fallback to have consistent behavior with vivado in case of designs originally in VHDL

Regards,
Adrien

### DIFF
--- a/common/kernel/basectx.h
+++ b/common/kernel/basectx.h
@@ -213,7 +213,9 @@ struct BaseCtx
 
     NetInfo *getNetByAlias(IdString alias) const
     {
-        return nets.count(alias) ? nets.at(alias).get() : nets.at(net_aliases.at(alias)).get();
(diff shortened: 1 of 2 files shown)


common/kernel/basectx.h
```diff
@@ -213,7 +213,9 @@ struct BaseCtx
 
     NetInfo *getNetByAlias(IdString alias) const
     {
-        return nets.count(alias) ? nets.at(alias).get() : nets.at(net_aliases.at(alias)).get();
+        if(nets.count(alias) > 0) return nets.at(alias).get();
```
gatecat, line 216: can you run a `make clangformat` on this?

marzoul: Now done.

PR #1376, opened 2024-10-02
Gowin. Fix the port check for connectivity.

yrabbit:
What happens is that it's not enough to check for a network, we also need to make sure that the network is functional: has src and sinks.

And the style edits - they get automatically when I make sure to run clang-format10.


gatecat:
Do you think you'd be able to update to a newer clang-format version (preferably 18)? Otherwise, at least not commit the files that only contain formatting reversions?

Issue #1379, opened 2024-10-02
terminate called after throwing an instance of 'std::out_of_range' during nextpnr_ice40::log_crit_paths


Barabas5532:
nextpnr crashes with the following message: 

```
<previous output removed>
Info: Critical path report for cross-domain path 'posedge CLK$SB_IO_IN' -> 'posedge CLK$SB_IO_IN_$glb_clk':
Info:       type curr  total name
Info: clk-to-clk  1.59  1.59 Net  (25,0) -> (21,9)
Info:                          Sink dut.read_data_SB_DFFE_Q_8_DFFLC.CLK
terminate called after throwing an instance of 'std::out_of_range'
  what():  dict::at()
```

To reproduce, check out my repo and run `command.sh`: https://github.com/Barabas5532/crush/commit/59d3a06fb4c04f1fbbbd76f77e432f0439987abc

I can also explain how to generate the JSON file from the verilog source if required. This command uses a pre-generated JSON file.

The JSON file was produced by yosys version:

```
Yosys 0.43 (git sha1 ead4718e5, g++ 14.2.1 -march=x86-64 -mtune=generic -O2 -fno-plt -fexceptions -fstack-clash-protection -fcf-protection -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -ffile-prefix-map=/build/yosys/src=/usr/src/debug/yosys -fPIC -Os)
```
My nextpnr version:

```
"nextpnr-ice40" -- Next Generation Place and Route (Version nextpnr-0.7-123-g75d2ce6a)
```

<details>
<summary>Backtrace</summary>

```
usr/src/debug/gcc/gcc/libstdc++-v3/libsupc++/eh_throw.cc:81
#1  0x0000555555653ad2 in nextpnr_ice40::dict<nextpnr_ice40::IdString, std::unique_ptr<nextpnr_ice40::NetInfo, std::default_delete<nextpnr_ice40::NetInfo> >, nextpnr_ice40::hash_ops<nextpnr_ice40::IdString> >::at (
    this=0x555563677400, key=...) at /usr/src/debug/nextpnr-git/nextpnr/common/kernel/hashlib.h:607
#2  operator() (__closure=__closure@entry=0x7fffffffc3d0, path=...) at /usr/src/debug/nextpnr-git/nextpnr/common/kernel/timing_log.cc:109
[...]

PR #1380, opened 2024-10-03
ice40: Fix missing clock pin types

gatecat:
I spotted in #1379 an unexpected cross-clock path showing up that the hard IP clock pins had been missed and weren't being driven through the global buffer network.

Maybe this even fixes some (but not all, there are also design issues in them reproducible in icestorm/sim models) of the weird issues people have seen over these years with these IPs...

This does fix the crash in #1379, but there's also no reason that the cross-clock path should crash timing analysis, so there's still something there for @rowanG077 to look into too.


rowanG077:
@gatecat The lookup failure is fixed with https://github.com/YosysHQ/nextpnr/pull/1381.

PR #1381, opened 2024-10-04
timing_log: Handle potentially missing net when reporting crit path

rowanG077:
Caused by the original code path only handling routing which always has a net. Now with the additional clock delay reporting a net is not always available depending on exactly where it came from. Which lead to a faulty lookup. 

Together with https://github.com/YosysHQ/nextpnr/pull/1380 fixes https://github.com/YosysHQ/nextpnr/issues/1379.


gatecat:
Thanks!

PR #1386, opened 2024-10-22
Gowin. FFs placement.

yrabbit:
* Allow clusters to be created from FFs and LUTs;

* Immediately create pass-through LUTs from free LUTs adjacent to FF - at the same time ensure alternating use of LUT inputs;

* In case of constant networks, such pass-through LUTs are disconnected from networks altogether;

* Allow FF to be placed directly into SSRAM slides - this is useful when using synchronous reading.

### DIFF
--- a/himbaechel/uarch/gowin/gowin.cc
+++ b/himbaechel/uarch/gowin/gowin.cc
@@ -740,11 +740,6 @@ void GowinImpl::assign_cell_info()
 void GowinImpl::create_passthrough_luts(void)
 {
     std::vector<std::unique_ptr<CellInfo>> new_cells;
-    // evenly use all 4 LUT inputs
-    const std::vector<std::pair<IdString, int>> lut_tmpl = {
-            {id_I0, 0xaaaa}, {id_I1, 0xcccc}, {id_I2, 0xf0f0}, {id_I3, 0xff00}};
-    int cur_lut_tmpl = 3;
-
     for (auto &cell : ctx->cells) {
         CellInfo *ci = cell.second.get();
         if (is_dff(ci)) {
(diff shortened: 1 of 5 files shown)


himbaechel/uarch/gowin/gowin_utils.cc
```diff
@@ -181,6 +181,15 @@ bool GowinUtils::need_BLKSEL_fix(void)
     return extra->chip_flags & Extra_chip_data_POD::NEED_BLKSEL_FIX;
 }
 
+IdString GowinUtils::create_aux_name(IdString main_name, int idx, const char *str_suffix)
+{
+    std::string sfx("");
+    if (idx) {
+        sfx = std::to_string(idx);
+    }
+    return ctx->id(main_name.str(ctx) + std::string(str_suffix) + sfx);
```
gatecat, line 190: this could use ctx->idf instead

himbaechel/uarch/gowin/gowin.cc
```diff
@@ -725,6 +736,73 @@ void GowinImpl::assign_cell_info()
...
+                    if (d_net->name == ctx->id("$PACKER_GND")) {
+                        lut->setParam(id_INIT, 0x0000);
+                    } else {
+                        lut->setParam(id_INIT, 0xffff);
+                    }
+                } else {
+                    if (ctx->debug) {
+                        log("make a pass-through.\n");
+                    }
+                    IdString lut_input = lut_tmpl.at(cur_lut_tmpl).first;
```
gatecat, line 781: is there a reason for changing which LUT input is used each time? this seems like something that long term might be better dealt with using LUT permutation

himbaechel/uarch/gowin/gowin_utils.cc
yrabbit: done

himbaechel/uarch/gowin/gowin.cc
yrabbit: There is no particular reason. I can use the I3 input to all LUTS if you think it is so convenient for the one who will implement Permutation in the future.

gatecat: It would make the code a bit simpler, if there are no other downsides I would just stick to I3 particularly if permutation is planned for the future

yrabbit: done

PR #1395, opened 2024-11-22
Add expandBoundingBox method to API

mmicko:
Add expandBoundingBox method to main nextpnr API and Himbaechel API to be able to handle architectures and uarch that have different needs for expanding bound box.

Used oportunity to remove Mistral specifics in router2.cc and move it into architecture specific file.

### DIFF
--- a/docs/archapi.md
+++ b/docs/archapi.md
@@ -771,4 +771,6 @@ Router Methods
 
 ### void expandBoundingBox(BoundingBox &bb) const
 
-Updates bounding box to expand search space when routing is failing to find possible route.
+As part of `router2` implementation, during congestion update, every third time a net fails to route, this method is executed to expand the bounding box to increase the search space.
+
+Default implementation expands by one tile in each direction.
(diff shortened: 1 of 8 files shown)


docs/archapi.md
```diff
@@ -765,3 +765,10 @@ Returns `true` if the cell **must** be placed according to the cluster; for exam
 Gets an exact placement of the cluster, with the root cell placed on or near `root_bel` (and always within the same tile). Returns false if no placement is viable, otherwise returns `true` and populates `placement` with a list of cells inside the cluster and bels they should be placed at.
 
 This approach of allowing architectures to define cluster placements enables easier handling of irregular fabrics than requiring strict and constant x, y and z offsets.
+
+Router Methods
+---------------
+
+### void expandBoundingBox(BoundingBox &bb) const
+
+Updates bounding box to expand search space when routing is failing to find possible route.
```
gatecat, line 774: Maybe a little bit more explanation here would be good?

PR #1398, opened 2024-11-29
Gowin. Add IODELAY.

yrabbit:
Input/Output delay (IODELAY) is programmable delay uint in IO block.

This delay line is enabled before/after the IO pad and allows the signal to be delayed statically or dynamically during 0-127 stages each lasting from 18 to 30 picoseconds depending on the chip family.


himbaechel/uarch/gowin/pack.cc
```diff
@@ -618,21 +625,142 @@ struct GowinPacker
...
+            }
+            // There is only one delay line in the IO block, which can be either
+            // input or output.  Define which case we are dealing with.
+            bool is_idelay = false;
+            NetInfo *di_net = ci.ports.at(id_DI).net;
+            NetInfo *do_net = ci.ports.at(id_DO).net;
+            CellInfo *iob = net_driven_by(ctx, di_net, is_iob, id_O);
+            if (iob != nullptr) {
+                NPNR_ASSERT(iob->bel != BelId());
+                NPNR_ASSERT_MSG(di_net->users.entries() == 1, "IODELAY should be the only sink in the network.\n");
```
gatecat, line 666: This is better as a log_error, as it's something the user could conceivably trigger with a design

PR #1399, opened 2024-11-29
Add GroupId related calls to Himbaechel API

mmicko:
This PR adds GroupId support to chip database and enables  drawing of Group decal.
Also have run clangformat and updated affected himbaechel files.
Update chip database version to 6

It includes changes on example architecture to utilize this.


himbaechel/himbaechel_api.h
```diff
@@ -109,10 +109,13 @@ struct HimbaechelAPI
...
-    virtual void drawWire(std::vector<GraphicElement> &g, GraphicElement::style_t style, Loc loc, IdString wire_type, int32_t tilewire, IdString tile_type) {};
+    virtual void drawWire(std::vector<GraphicElement> &g, GraphicElement::style_t style, Loc loc, IdString wire_type,
+                          int32_t tilewire, IdString tile_type) {};
 
-    virtual void drawPip(std::vector<GraphicElement> &g,GraphicElement::style_t style, Loc loc,
-                WireId src, IdString src_type, int32_t src_id, WireId dst, IdString dst_type, int32_t dst_id) {};
+    virtual void drawPip(std::vector<GraphicElement> &g, GraphicElement::style_t style, Loc loc, WireId src,
+                         IdString src_type, int32_t src_id, WireId dst, IdString dst_type, int32_t dst_id) {};
+
+    virtual void drawGroup(std::vector<GraphicElement> &g, IdString group_type, Loc loc) {};
```
mmicko, line 118: @gatecat  Should I change interface by adding group_name or even better we can just propagating GroupId and let it handled depending of implementation ?
Was thinking about this API and for example in case of Ice40 there would be multiple switchboxes of same type, also we can represent warious muxes groups like this and those are drawn same way but on different places depending both on type but also on id.

gatecat: I think just GroupId is fine

mmicko: Thanks, updated code accordingly

PR #1405, opened 2024-12-11
Add attributes to the hierarchical cells

tarikgraba:
Hi,

I did not find a way to retrieve attributes (assigned in the verilog) of hierarchical cells even though yosys preserves them in the json netlist.

This patch adds an "attrs" dictionary to the HierarchicalCell structure and expose it through the python bindings.

Those attributes are imported in the front end after the import of the submodule.

This last point I am not sure of and maybe I am not doing this correctly/efficietly.

I wanted to add this feature to mimic the behavior of Xilinx's LOC/RLOC attributes where, in a hierarchy, you can assign a relative location to primitive cells within a module and a global location for instances of this module.


gatecat:
Thanks!

PR #1407, opened 2024-12-22
Fix gowin ALU SUB mode ports

aerkiaga:
By examining the output of `nextpnr`, I see an `ALU` in mode 1 (`SUB`) is wired like this:

```
Verilog I0    I0
Verilog I1    I1
Verilog I3    I3
```

When parsed by `gowin_pack`, this would result in an `ALU` that computes `I0 - I3`.
This PR swaps ports `I1` and `I3` to produce the following (verified in `nextpnr` output):

```
Verilog I0    I0
Verilog I1    I3
Verilog I3    I1
```

This matches Apicula documentation, as well as `ADDSUB` ports.
Can't test on real hardware at the time (used my Tang Nano 1K and 20K on a project, haven't ordered replacement yet).

### DIFF
--- a/himbaechel/uarch/gowin/pack.cc
+++ b/himbaechel/uarch/gowin/pack.cc
@@ -1423,6 +1423,12 @@ struct GowinPacker
                             ci->renamePort(id_I0, id_I3);
                             ci->renamePort(id_I2, id_I0);
                         }
+                        // XXX mode 1 - SUB
+                        else if (ci->params.at(id_ALU_MODE).as_int64() == 1) {
+                            ci->renamePort(id_I3, id_I2);
+                            ci->renamePort(id_I1, id_I3);
+                            ci->renamePort(id_I2, id_I1);
+                        }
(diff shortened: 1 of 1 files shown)


yrabbit:
It would be nice to correct the connection of I2 to VCC/GND here, depending on whether the mode is MULT or not. :)

aerkiaga:
@yrabbit I would argue that the status quo of having I2 connect to VCC is more useful; it can do the same operation by just fixing I3 down, plus it can take I3 as a second input (which is extremely useful for a multiplier). I suggesting updating `cells_sim.v` in `yosys` instead, I'll send a PR.

aerkiaga:
I've created the aforementioned PR. If you'd rather see the wiring updated instead, I can fix that and propose a new mode for the 2-input `MULT` (while changing unused bits in the regular `MULT`). What do you think?

yrabbit:
I have no objections, if it works then great:)

PR #1414, opened 2025-01-11
Find all components of Python at the same time

whitequark:
This is explicitly recommended by the FindPython module documentation and is required to avoid failed builds on some systems. See: https://cmake.org/cmake/help/latest/module/FindPython.html

See also https://github.com/YosysHQ/prjtrellis/pull/243.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,7 +10,7 @@ cmake_policy(SET CMP0079 NEW)
 # compilation database as not all clang/gcc share the same implicit includes
 # leading to essentially non-working compile_commands.json
 if(CMAKE_EXPORT_COMPILE_COMMANDS)
-  set(CMAKE_CXX_STANDARD_INCLUDE_DIRECTORIES 
+  set(CMAKE_CXX_STANDARD_INCLUDE_DIRECTORIES
       ${CMAKE_CXX_IMPLICIT_INCLUDE_DIRECTORIES})
 endif()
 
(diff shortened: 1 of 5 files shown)


mmicko:
@whitequark This breaks one important use case, and that is oss-cad-suite cross compilation. If you are searching for ```Interpreter``` and ```Development.Embed``` they needs to be from same version/location, but in case of cross-compiling ```Interpreter``` should be local python from host and ```Development.Embed``` points to cross-compiled python libraries of target architecture. Tried splitting this again in two calls and it works fine so other changes are not affecting build.

whitequark:
"It works fine" on your system isn't enough. It does not work fine on mine, where multiple versions of Python are installed.

I also think that oss-cad-suite build could very well specify Python versions via `-D...` arguments, since it's using build scripts anyway. Finding Python automatically is more important where build is done manually, since the experience of having to hunt down these parameters is pretty awful, and people end up just opening issues here when things break.

mmicko:
@whitequark It is not a problem for me to override parameters. I do set ```Python3_INCLUDE_DIR``` and ```Python3_LIBRARY``` to cross-compile location, but that does not work any more in case they are searched together, looks like cmake script ignores overriding it and forces use of Python from location that supports both Interpreter and Development.Embed instead.

whitequark:
I see. Then I think we should detect cross-compiling and split them in that case.

whitequark:
Actually, I have a major additional refactor coming on in #1422 that I'd like to have merged today, do you think we could postpone this until it gets merged? I have one more test to run and then it's ready.

mmicko:
@whitequark Sure no problem to wait. 
Next snippet do the trick for cross compilation, but still there is a problem of compiling nexpnr with specific self-compiled python. Will see which parameters needs to be altered for that  case since  Python3_ROOT_DIR  does not help
```
    if (CMAKE_CROSSCOMPILING)
        find_package(Python3 3.5 REQUIRED COMPONENTS Interpreter)
        find_package(Python3 3.5 REQUIRED COMPONENTS Development.Embed)
    else ()
        find_package(Python3 3.5 REQUIRED COMPONENTS Interpreter Development.Embed)
    endif()
```

mmicko:
@whitequark manage to find a way to make it work without altering cmake files. It is enough to actually override PATH and put locally compiled native python version first, then overrides for include and lib also do work, so this fixes both of my use cases. Will use other PR to make things for with ubunut-latest since that is quite broken

whitequark:
Thank you for the investigation! I'm glad you were able to find a solution.

Issue #1417, opened 2025-01-12
Selftest fails for some Himbächel Gowin devices


whitequark:
While working on adding Himbächel Gowin support to YoWASP, I've noticed that the selftest (`--test` option) frequently fails.

It succeeds with, for example:

```
$ ./nextpnr-himbaechel --device GW1NR-LV9QN88PC6/I5 --vopt family=GW1N-9C --test
```

It fails with, for example but not limited to:

```
$ ./nextpnr-himbaechel --device GW1N-LV1QN48C6/I5 --test
$ ./nextpnr-himbaechel --device GW2A-LV18PG256C8/I7 --test
$ ./nextpnr-himbaechel --device GW2A-LV18PG256C8/I7 --vopt family=GW2A-18C --test
$ ./nextpnr-himbaechel --device GW2A-LV18PG256C8/I7 --vopt family=GW2A-18 --test
```

whitequark:
cc @yrabbit

PR #1418, opened 2025-01-12
Gowin. BUGFIX. Do not create missing wires.

yrabbit:
Erroneously created wires for specific IOs on the underside of some chips.

Fixes https://github.com/YosysHQ/nextpnr/issues/1417

Also cosmetic edits.


whitequark:
That was fast, great work!

PR #1421, opened 2025-01-15
Rationalize and refactor build system (part 1)

whitequark:
This fairly large PR is not intended to introduce functional changes, but rather tidy up the architecture-independent parts of the CMake build system: make sure they're up to date with CMake guidelines (with the new minimum required version being CMake 3.25, the one in Debian stable today) and that the CMake code makes sense and is easy to understand.

The changes in this PR are very conservative; I have removed a few commands but only in places where the outcome should unambiguously stay the same. It is not impossible that it breaks the build on some obscure target but I would be surprised if it does.

The part 2 will refactor and deduplicate architecture-specific code, which will inevitably be more invasive.


whitequark:
The error comes from the combination of `-Werror` and gtest. I'm not sure what specifically exposed it (I don't _think_ I changed gtest stuff?) but I'll just patch it in the vendored copy.

```
In file included from /home/runner/work/nextpnr/nextpnr/3rdparty/googletest/googletest/src/gtest-all.cc:43:
/home/runner/work/nextpnr/nextpnr/3rdparty/googletest/googletest/src/gtest-death-test.cc: In function ‘bool testing::internal::StackGrowsDown()’:
/home/runner/work/nextpnr/nextpnr/3rdparty/googletest/googletest/src/gtest-death-test.cc:1009:24: error: ‘dummy’ may be used uninitialized [-Werror=maybe-uninitialized]
 1009 |   StackLowerThanAddress(&dummy, &result);
      |   ~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~
/home/runner/work/nextpnr/nextpnr/3rdparty/googletest/googletest/src/gtest-death-test.cc:999:6: note: by argument 1 of type ‘const void*’ to ‘void testing::internal::StackLowerThanAddress(const void*, bool*)’ declared here
  999 | void StackLowerThanAddress(const void* ptr, bool* result) {
      |      ^~~~~~~~~~~~~~~~~~~~~
/home/runner/work/nextpnr/nextpnr/3rdparty/googletest/googletest/src/gtest-death-test.cc:1007:7: note: ‘dummy’ declared here
 1007 |   int dummy;
      |       ^~~~~
cc1plus: all warnings being treated as errors
```

whitequark:
I think this should be ready to go.

CMakeLists.txt
```diff
@@ -371,12 +363,12 @@ foreach (family ${ARCH})
         if (USE_RUST)
             target_sources(${target} PRIVATE ${RUST_FILES})
```
gatecat, line 364: looks like this is still using `${RUST_FILES}` which was removed, so I think this would fail with USE_RUST. 

As no-one is actually using the Rust bindings for anything (and if they wanted to, there are probably better ways of doing this, in retrospect), we could also remove them if you think the build system support to keep them is disproportionate.

whitequark: I actually tested `-DUSE_RUST=ON` and it links and runs fine (including with the weird Rust-Python bridge enabled). I haven't tested calling them at runtime since I did get undefined symbol errors before I added the Rust support and that seemed enough.

It didn't cause me any significant effort to keep the support in; I suspect having them around probably doesn't make sense but this isn't a decision that I feel like should be taken based on build system complexity. Rather, should we remove them because they're not even tested on CI?

I can do the removal if you'd like me to.

gatecat: Thanks for confirming that it still builds fine, that's fine for this PR - I will probably think about removal (or adding a quick CI smoketest) once the build system refactoring is done.

cr1901:
>(with the new minimum required version being CMake 3.25, the one in Debian stable today) and that the CMake code makes sense and is easy to understand.

If you're on Ubuntu 22.04 LTS (supported for another 2 years), follow the instructions [here](https://apt.kitware.com/) to update `cmake` to a supported version.

PR #1422, opened 2025-01-15
Rationalize and refactor build system (part 2)

whitequark:
Depends on https://github.com/YosysHQ/nextpnr/pull/1421.

Also allows splitting `nextpnr-himbaechel` per microarchitecture.


cmake/FindIceStorm.cmake
```diff
@@ -0,0 +1,19 @@
+set(icestorm_default_install_prefix ${CMAKE_INSTALL_PREFIX})
+if (DEFINED ICEBOX_ROOT) # for compatibility with old build scripts
```
gatecat, line 2: while we're refactoring things, we can probably just remove this, it's been deprecated for very many years now

CMakeLists.txt
```diff
@@ -52,15 +54,24 @@ else()
...
 else()
     set(BBASM_MODE "string")
 endif()
 
 find_package(Threads)
 if (Threads_FOUND)
-    find_package(TBB QUIET)
+    if (NOT STATIC_BUILD)
+        # TBB is not provided as a statically linked library.
+        find_package(TBB QUIET)
```
gatecat, line 73: as the PR is touching this anyway, might as well remove the optional TBB dependency, it was only used by the now-removed FPGA interface support

himbaechel/CMakeLists.txt
```diff
@@ -0,0 +1,89 @@
...
+    message(FATAL_ERROR "Microarchitecture setting is mandatory")
+endif()
+
+if (HIMBAECHEL_UARCH STREQUAL "all")
+    set(HIMBAECHEL_UARCH ${HIMBAECHEL_UARCHES})
+endif()
+
+foreach (uarch ${HIMBAECHEL_UARCH})
+    if (NOT uarch IN_LIST HIMBAECHEL_UARCHES)
+        message(FATAL_ERROR "Microarchitecture ${uarch} is not a supported Himbächel microarchitectures")
```
gatecat, line 84: nitpick - microarchitecture not microarchitectures

CMakeLists.txt
whitequark: done

himbaechel/CMakeLists.txt
whitequark: done

README.md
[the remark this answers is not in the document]
whitequark: done, and also explained how/why to use split himbächel

gatecat (approved):

mmicko:
@whitequark there are couple of things that are now broken (when building oss-cad-suite but some are general)

minor issues (when doing in-source-tree build):
- .bba and .cc as product are now generated in same directory as others so not covered by .gitignore, previously it was in chipdb subdirectory
- since "generated" subdirectory is not used any more .gitignore needs updating to cover those files

major issues :
- there is no target just to generate BBA files and nothing else for architecture
- there is no way to tell build system to use already generated BBA files from other location

Since generating BBA files takes time but always generate same output (for same endianess) and sometimes specific binaries/scripts only for this, this was used on oss-cad-suite build to minimize build time and lower redudancy.

Using nextpnr-**arch**-chipdb target we can get BBA created (even at the wrong location right now) but it also lose some time on generating .cc files, splitting these two into two targets and making a way to propagate BBA_CHIPDB location for bba files instead would help fixing this.

whitequark:
> when doing in-source-tree build

I would rather disallow in-source-tree builds entirely. They cause many issues and with CMake in particular there are just way too many auxiliary files to keep track of. Projects like LLVM already do that for similar reasons.

@gatecat What do you think?

> * there is no target just to generate BBA files and nothing else for architecture
> * there is no way to tell build system to use already generated BBA files from other location

The change that resulted in this was done quite intentionally since the existing system for generating BBA files was unmaintainable itself and also an unmanageable obstacle to changing how the build system works.

Pre-running `bbasm` is something I have no plans of allowing again. If you are bottlenecked on `.cc` file generation then you should use a compiler with `#embed` file support (Clang 19 or later) or use `-DEXTERNAL_CHIPDB=ON` (which aligns legacy architectures with Himbaechel architectures).

Pre-running `.bba` file generation is something that I decided could be restored later as a capability if it becomes an issue while working on this refactoring. However, I want to say that I am also packaging the entirety of nextpnr for YoWASP, and my builds always finish within about 30 minutes. Is it really a major issue if it doesn't take that much time?

whitequark:
I'm going to add export/import of generated `.bba` files since it was more widely used/needed than I anticipated. It shouldn't take too long.

whitequark:
I've implemented this functionality in https://github.com/YosysHQ/nextpnr/pull/1436, please let me know if it works. The `nextpnr-all-bba` target (or `nextpnr-<arch>-bba`) doesn't depend on anything but the `.bba` files themselves and the `EXPORT_BBA_FILES` directory doesn't contain anything extra. The `.cc` files are generated as a part of the subsequent build with `IMPORT_BBA_FILES` so I think it should solve the issue of that taking a long time also.

I don't know how useful it is, but you can also do a normal build with `-DEXPORT_BBA_FILES` enabled and it will both build nextpnr and export `.bba` files.

@cr1901 You should be able to run `make nextpnr-all-bba -j1` target as a part of a normal build as well, to serialize producing these files.

PR #1426, opened 2025-01-20
Updating CI to work with ubuntu-latest

mmicko:
~~Split Python search again changed in https://github.com/YosysHQ/nextpnr/pull/1414  to support cross compilation environment where python binary may be different version and location than library and includes.~~

Repurposing this PR to be used for making CI work with ubuntu-latest since that is currenly quite broken.


whitequark (changes requested):
NAK, this will just break the systems where the behavior I added is required.

I will accept this change if you detect cross-compilation and split it only in those cases.

whitequark (approved):

PR #1435, opened 2025-01-22
CMake: disallow in-tree builds

whitequark:
In-tree builds pollute the source directory and make version control more difficult to use effectively.

See https://github.com/YosysHQ/nextpnr/pull/1422#issuecomment-2606815347 for prior discussion.


whitequark:
If you try to run `cmake` in the source directory you will now see this:

image

It also creates a `CMakeFiles` directory in the source root, which is annoying but unavoidable (even if you remove it right before the `message()` call it re-creates it again).

whitequark:
Merging now by request of @mmicko.

PR #1436, opened 2025-01-22
CMake: add support for exporting and importing .bba files

whitequark:
This is useful for certain cross-compilation workloads, and to cache rarely changing build products.

To use this functionality, build e.g. as follows:

    cmake . -B build-export -DEXPORT_BBA_FILES=../bba-files -DARCH=all
    cmake --build build-export -t nextpnr-all-bba

    cmake . -B build-import -DIMPORT_BBA_FILES=../bba-files -DARCH=all
    cmake --build build-import

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -237,6 +237,8 @@ add_subdirectory(rust)
 
 add_subdirectory(tests/gui)
 
+add_custom_target(nextpnr-all-bba)
+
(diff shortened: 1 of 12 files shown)


mmicko:
Thanks @whitequark for this, now working as expected. Would be good that #1435 is then also merged since that way I can update all required scripts and make sure all is building as expected.

PR #1438, opened 2025-01-23
CMake: Add include guards when IMPORT_BBA_FILES is used

mmicko:
This one also includes small patch for machxo arch to properly build with GUI enabled


whitequark (approved):

whitequark:
That looks like it breaks non-GUI builds of machxo2.

mmicko:
@whitequark noticed and fixed that

whitequark:
Please feel free to merge once CI passes.

PR #1441, opened 2025-01-23
Update README to represent most recent changes in build system

mmicko:


README.md
```diff
@@ -59,16 +59,17 @@ Getting started
 For iCE40 support, install [Project IceStorm](https://github.com/YosysHQ/icestorm) to `/usr/local` or another location, which should be passed as `-DICESTORM_INSTALL_PREFIX=/usr` to CMake. Then build and install `nextpnr-ice40` using the following commands:
 
 ```
-cmake . -DARCH=ice40
-make -j$(nproc)
-sudo make install
+cmake . -B build -DARCH=ice40
```
gatecat, line 62: I think just doing a mkdir/cd build beforehand would be clearer than the -B/-Cs myself, not sure if you have a strong preference here

mmicko: No preference to be honest. There are always multiple ways, whatever looks better to others is fine by me

README.md
```diff
@@ -59,16 +59,18 @@ Getting started
 For iCE40 support, install [Project IceStorm](https://github.com/YosysHQ/icestorm) to `/usr/local` or another location, which should be passed as `-DICESTORM_INSTALL_PREFIX=/usr` to CMake. Then build and install `nextpnr-ice40` using the following commands:
 
 ```
-cmake . -DARCH=ice40
+mkdir -p build && cd build
+cmake .. -DARCH=ice40
+make nextpnr-all-bba
```
whitequark, line 64: I don't think this is needed?

mmicko: It was not needed before, but now it fails if you run it with -j parameter even for ice40

mmicko: @gatecat found solution for others, but himbaechel is already having issue that it starts bba creation twice if we run with -j
Next change will work for rest of archs.

```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index c3561e3c..1f68a62f 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -337,6 +337,8 @@ function(add_nextpnr_architecture target)
 
     add_library(nextpnr-${target}-chipdb INTERFACE)
 
+    add_dependencies(nextpnr-${target}-chipdb nextpnr-${target}-bba)
+
     target_link_libraries(nextpnr-${target}-core INTERFACE nextpnr-${target}-chipdb)
 
     # CLI/GUI runner
```

mmicko: anyway I have removed this line to prepare README with usage as it should be

whitequark: I thought I fixed this by removing SERIALIZE_CHIPDBS but evidently not... I can't spend time on this for a bit, but once I can I'd really like to review it in depth so it doesn't reoccur later somehow.

README.md
```diff
@@ -212,14 +220,16 @@ Additional notes for building nextpnr
 The following runs a debug build of the iCE40 architecture without GUI, without Python support, without the HeAP analytic placer and only HX1K support:
 
 ```
-cmake . -DARCH=ice40 -DCMAKE_BUILD_TYPE=Debug -DBUILD_PYTHON=OFF -DICE40_HX1K_ONLY=1
+mkdir -p build && cd build
+cmake .. -DARCH=ice40 -DCMAKE_BUILD_TYPE=Debug -DBUILD_PYTHON=OFF -DICE40_HX1K_ONLY=1
```
whitequark, line 224: The `ICE40_HX1K_ONLY` flag is no longer recognized by the build system (not sure since when, I didn't remove it a week ago...)

whitequark (changes requested):

README.md
mmicko: Fixed this

whitequark (approved):

README.md
whitequark: Should be fixed by https://github.com/YosysHQ/nextpnr/pull/1444

PR #1442, opened 2025-01-27
ecp5: Fix constant and inverted CEMUX

gatecat:
In one case, Yosys generates a never-toggling flipflop with constant-1 `CE` and `CEMUX` set to `INV`. This is a bit odd, but nextpnr should still handle this case correctly.

### DIFF
--- a/ecp5/pack.cc
+++ b/ecp5/pack.cc
@@ -924,7 +924,8 @@ class Ecp5Packer
                 if (is_lut(ctx, uc)) {
                     set_lut_input_constant(uc, user.port, constval);
                 } else if (is_ff(ctx, uc) && user.port == id_CE) {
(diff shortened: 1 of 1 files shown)


leocassarani:
@gatecat Thanks for fixing this so quickly! I've just tested your patch with the following repro case and I can confirm that it works as expected on my board (i.e. the LED stays off) — as opposed to nextpnr on `master` where the LED would be on. Much appreciated 🙏 

```verilog
module top (
  input clk,
  output reg led_g = 1'b0
);

  reg a = 1'b0;
  always @(posedge clk)
    a <= !a;

  wire [2:0] b = 3'd3 + { 2'b0, a };

  always @(posedge clk)
    if (b == 3'd0) led_g <= 1'b1;

endmodule
```

PR #1444, opened 2025-01-29
CMake: exclude *-bba and *-chipdb targets from `make all`

whitequark:
Due to the way CMake-generated Makefiles evaluate dependencies, this calls the `.bba` generation custom command twice, which then fails as they both use the same `.bba.new` file as an output and one of them moves it first.

This broke builds using `make -j` but not builds using `make -j nextpnr-himbaechel-example`.


gatecat:
Thanks for fixing this!

PR #1445, opened 2025-01-29
Gowin. Add I3C io buffer.

yrabbit:
A buffer is added that can operate as a normal IOBUF in PUSH-PULL mode or switch to open-drain IOBUF mode.

### DIFF
--- a/himbaechel/uarch/gowin/gowin.h
+++ b/himbaechel/uarch/gowin/gowin.h
@@ -36,6 +36,10 @@ inline bool is_diffio(const CellInfo *cell) { return type_is_diffio(cell->type);
 inline bool type_is_mipi(IdString cell_type) { return cell_type.in(id_MIPI_OBUF, id_MIPI_OBUF_A, id_MIPI_IBUF); }
 inline bool is_mipi(const CellInfo *cell) { return type_is_mipi(cell->type); }
 
+// I3C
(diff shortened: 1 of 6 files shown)


himbaechel/uarch/gowin/gowin.h
```diff
@@ -106,6 +110,7 @@ NPNR_PACKED_STRUCT(struct Tile_extra_data_POD {
     int32_t class_id;
     int16_t io16_x_off;
     int16_t io16_y_off;
+    int32_t i3c_capable;
```
gatecat, line 113: are there likely to be other flags like this you want to store here? instead of using a whole int32_t for this, what about an int32_t flags field with one bit for this and the rest reserved for future, similar, tile flags

uis246:
Could it be working in one state as OBUF, while in other as TBUF with constant output? Is there some LUT? 2xLUT2 for example.

yrabbit:
It's a little unclear. Are you hypothesizing how this is implemented in silicon? At the transistor level or something?

uis246:
Yes.

yrabbit:
It's hard to say. From the fact that I3C buffers are not on all chips, and on those that are, they are located only where MIPI_OBUF or MIPI_IBUF, we can assume that “borrowed” some of the mechanisms of MIPI.

Perhaps it is just a dynamic resistor connection to VCC that provides open-drain.

PR #1455, opened 2025-02-13
gowin: Add deprecation message for nextpnr-gowin

mmicko:
Adds deprecation warining as requested in https://github.com/YosysHQ/nextpnr/pull/1318 
It has to be done as (f)printf since during that time logger is still not active and it is useful to add warning in that particular moment.


whitequark (approved):

whitequark:
I'll leave a message on this PR once I can confirm nextpnr-gowin can be safely removed. (cc @gatecat @yrabbit)

yrabbit:
commit https://github.com/YosysHQ/nextpnr/pull/1318 actualized.

PR #1457, opened 2025-02-24
Add arch API function for pip inversion

Ravenslofty:
GateMate contains inverting switchbox pips, which needs to be handled specially (I suspect Cyclone V does too, but it's been a long while since I looked at that). As part of bringup for that family, add an architecture API function to detect if a pip inverts. 

Inverting pips between two wires are disambiguated from non-inverting for use by e.g. Python by having an `INV` string appended to the name.

### DIFF
--- a/himbaechel/arch.cc
+++ b/himbaechel/arch.cc
@@ -285,24 +285,33 @@ IdStringList Arch::getWireName(WireId wire) const
 
 PipId Arch::getPipByName(IdStringList name) const
 {
-    NPNR_ASSERT(name.size() == 3);
-    int tile = tile_name2idx.at(name[0]);
+    NPNR_ASSERT(name.size() == 3 || (name.size() == 4 && name[3] == id("INV")));
+    const int tile = tile_name2idx.at(name[0]);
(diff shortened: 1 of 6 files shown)


himbaechel/arch.cc
```diff
@@ -285,24 +285,33 @@ IdStringList Arch::getWireName(WireId wire) const
...
+    const int tile = tile_name2idx.at(name[0]);
     const auto &tdata = chip_tile_info(chip_info, tile);
     for (int pip = 0; pip < tdata.pips.ssize(); pip++) {
         if (IdString(tdata.wires[tdata.pips[pip].dst_wire].name) == name[1] &&
-            IdString(tdata.wires[tdata.pips[pip].src_wire].name) == name[2])
-            return PipId(tile, pip);
+            IdString(tdata.wires[tdata.pips[pip].src_wire].name) == name[2]) {
+
+            const auto tmp_pip = PipId(tile, pip);
+            if ((name.size() == 3 && !getCtx()->isPipInverting(tmp_pip)) ||
```
gatecat, line 296: ditto here, `isPipInverting` belongs to `Arch` so no need for `getCtx`

PR #1460, opened 2025-03-01
rust: add getBels() binding

Ravenslofty:
(at what point do I spend the time trying to figure out a pattern for passing these iterators between C++ and Rust that doesn't require leaking an array?)

### DIFF
--- a/rust/nextpnr/src/lib.rs
+++ b/rust/nextpnr/src/lib.rs
@@ -239,16 +239,19 @@ impl Context {
         v
     }
 
-    pub fn wires_leaking(&self) -> &[WireId] {
-        let mut wires = std::ptr::null_mut();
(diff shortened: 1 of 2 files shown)


gatecat:
> (at what point do I spend the time trying to figure out a pattern for passing these iterators between C++ and Rust that doesn't require leaking an array?)

The best solution would probably be to actually try and wrap the iterator somehow

Ravenslofty:
Well, I intended to just add a fairly basic `getBels` binding, but yeah, I ended up porting the iterator wrappers to cover `getBels`/`getPips`/`getWires`.

(While working on this I found out that rustup 1.28 broke corrosion 0.4, so I bumped corrosion to 0.5.1. I hope you don't mind.)

PR #1462, opened 2025-03-07
Gowin. Add HCLK wires to PLL.

yrabbit:
Adds the ability to use high-speed clock lines (together with CLKDIV2 type frequency dividers operating on them) as signals for the CLKIN and CLKFB inputs of the rPLL and PLLVR primitives (these cover the full range of supported Gowin chips).

### DIFF
--- a/himbaechel/uarch/gowin/gowin.cc
+++ b/himbaechel/uarch/gowin/gowin.cc
@@ -287,7 +287,7 @@ void GowinImpl::adjust_dsp_pin_mapping(void)
 
 /*
    Each HCLK section can serve one of three purposes:
-       1. A simple routing path to IOLOGIC FCLK
+       1. A simple routing path to IOLOGIC FCLK or PLL inputs
        2. CLKDIV2
        3. CLKDIV (only one section at any time)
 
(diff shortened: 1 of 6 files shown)


himbaechel/uarch/gowin/gowin.cc
```diff
@@ -319,51 +358,57 @@ void GowinImpl::place_constrained_hclk_cells()
...
-                seen_hclk_users.insert(user.cell->name);
-
-                if (ctx->debug) {
-                    log_info("Custom HCLK Placer: Found HCLK user: %s\n", user.cell->name.c_str(ctx));
                 }
 
-                gwu.find_connected_bels(user.cell, id_FCLK, id_CLKDIV2, id_CLKOUT, 16, bel_candidates);
-                these_options.insert(bel_candidates.begin(), bel_candidates.end());
+                int alias_count = 0;
+                std::set<std::set<BelId>> seen_options;
```
gatecat, line 375: Can you explain exactly what this set-of-a-set is doing? I'm not sure I fully understand

yrabbit: The internal set is CLKDIV2 Bels, which could potentially be used to deliver the signal to sinks located in the service area of the HCLK sections. For example PLL inputs can use sections 1 and 3, and IOLOGIC FCLK inputs from the same network sections 0 and 1, but individual IOs only section 1. 

The external set, on the other hand, simply maintains the uniqueness of these variant Bels sets.

My changes to the existing algorithm were only to ensure proper handling of the case where a single cell is a sink in multiple networks (pins connected to different HCLK networks). I can ask the author of the original commit of this algorithm to comment if there are more questions about the routing method itself.

PR #1464, opened 2025-03-12
Gowin. Implement the DLLDLY primitive.

yrabbit:
DLLDLY is the clock delay primitive that adjust the input clock according to the DLLSTEP signal and outputs the delayed clock.

These primitives are associated with clock pins and are "tapped" between the output of this IBUF and the clock networks, leaving the possibility to connect to the original unshifted signal as well, although the latter is not very practical because it is no longer possible to use fast wires.


yrabbit:
Actualised.

gatecat:
Thanks. Do you have anything urgent (i.e. next one-two days) that you'd like to see in a nextpnr release?

yrabbit:
> Thanks. Do you have anything urgent (i.e. next one-two days) that you'd like to see in a nextpnr release?
Thank you. No, I don't have any additions planned. And possible future bugs found will be corrected as a matter of routine.

yrabbit:
> Thanks. Do you have anything urgent (i.e. next one-two days) that you'd like to see in a nextpnr release?

As usual it didn't take a day for something to show up. 

https://github.com/YosysHQ/nextpnr/pull/1467

This morning managed to run Linux on tangnan20k (!) using the latest yosys/nextpnr/apicula . In the process I found a random error - sometimes a route through DCS can be generated even though this primitive is not provided by the design.

The correct thing would be to generate static DCS automatically, but I would not make such changes before nextpnr release. On the other hand it's easy to disallow such networks and that's what I propose to do here, and I'll add the generation later when I think it through.

PR #1467, opened 2025-03-20
Gowin. Fix non-DCS networks.

yrabbit:
Prohibits the use of Dynamic Clock Selection PIPs for networks where no DCS is present.


gatecat:
Thanks, nice spot!

PR #1470, opened 2025-03-31
Gowin. BUGFIX Use a separate net for segment gates

yrabbit:
We use a temporary separate small network (typically 2 - 3 sinks) for routing from the segment network source to the segment gate. This fixes the rare but unpleasant case of self-intersection when a route to a gate is routed using PIPs after the gate, this is no longer allowed when using a separate small network.

### DIFF
--- a/himbaechel/uarch/gowin/globals.cc
+++ b/himbaechel/uarch/gowin/globals.cc
@@ -795,23 +795,44 @@ struct GowinGlobalRouter
         if (ctx->debug) {
             log_info("    step 3: %s -> \n", ctx->nameOfWire(src_wire));
         }
+        // Create a temporary small network where segment gates will be the sinks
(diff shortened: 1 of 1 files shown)


himbaechel/uarch/gowin/globals.cc
```diff
@@ -795,23 +795,44 @@ struct GowinGlobalRouter
         if (ctx->debug) {
             log_info("    step 3: %s -> \n", ctx->nameOfWire(src_wire));
         }
+        // Create a temporary small network where segment gates will be the sinks
+        IdString gate_net_name = ctx->id(ni->name.str(ctx) + std::string("$gate_net$"));
```
gatecat, line 799: this would be neater using `ctx->idf("%s$gate_net$", ni->name.c_str(ctx))`

yrabbit: Done.

PR #1471, opened 2025-04-01
Update README with instructions to install using guix.

csantosb:
GNU/Guix is a package manager, and includes packages for most of the architectures, no need to handle dependencies.


uis246:
> no need to handle dependencies.

Automagic!

PR #1472, opened 2025-04-13
placer1: add sanity check for try_swap_chain

mmicko:
Had to check both boundaries since lot of getBelByLocation implementations do not check at least one of them and it is possible to hit this case.


gatecat:
Thanks!

PR #1476, opened 2025-04-30
gatemate: Initial SERDES support

mmicko:
This enables basic support for SERDES primitive, and makes sure all parameters are of proper size and have default value.


himbaechel/uarch/gatemate/pack.cc
```diff
@@ -2170,6 +2170,347 @@ void GateMatePacker::pack_ram()
...
+        move_ram_i(&ci, id_RX_BUF_ERR_O);
+        move_ram_i(&ci, id_RX_BYTE_IS_ALIGNED_O);
+        move_ram_i(&ci, id_RX_BYTE_REALIGN_O);
+        move_ram_i(&ci, id_RX_RESET_DONE_O);
+        move_ram_i(&ci, id_RX_EI_EN_O);
+        move_ram_i(&ci, id_RX_CLK_O);
+        move_ram_i(&ci, id_PLL_CLK_O);
+        move_ram_i(&ci, id_REGFILE_RDY_O);
+
+        ci.params[id_RX_BUF_RESET_TIME] = Property(int_or_default(ci.params, id_RX_BUF_RESET_TIME, 3), 5);
```
gatecat, line 2252: maybe it'd be nicer to have a static const array somewhere of all these parameters and their widths/defaults, rather than all this imperative code

mmicko:
Addressed both comments.

PR #1477, opened 2025-05-02
Only add subdirectory tests, if BUILD_TESTS=ON

realchonk:


whitequark (approved):
LGTM. I assume this is due to difficulties with the submodule.

PR #1478, opened 2025-05-12
build: Allow using GoogleTest from the system.

Apteryks:
* CMakeLists.txt: [BUILD_TESTS]: Detect whether GTest is available, and alias 'gtest_main' to GTEST::gtest_main if it is. (gtest_include_dir): Set variable when using bundled GTest.


Apteryks:
Just to be clear, is this project opposed to allowing users to use system libraries instead of the copies
it bundles? Otherwise, what is the rationale for dropping this proposed change?

PR #1480, opened 2025-05-12
build: Allow linking to system-provided imgui and qtimgui libraries.

Apteryks:
* gui/CMakeLists.txt (IMGUI_INCLUDE_DIR, QTIMGUI_INCLUDE_DIR) (IMGUI_SOURCES, QTIMGUI_SOURCES, IMGUI_LIB, QTIMGUI_LIB): New variables.
<nextpnr-${target}-gui> [target_sources]: Use IMGUI_SOURCES and QTIMGUI_SOURCES.
<nextpnr-${target}-gui> [include_directories]: Use IMGUI_INCLUDE_DIR and QTIMGUI_INCLUDE_DIR.
<nextpnr-${target}-gui> [link_libraries]: Use IMGUI_LIB and QTIMGUI_LIB when defined.

### DIFF
--- a/gui/CMakeLists.txt
+++ b/gui/CMakeLists.txt
@@ -46,12 +46,42 @@ target_include_directories(nextpnr-${target}-gui PUBLIC
     ${CMAKE_CURRENT_SOURCE_DIR}
 )
 
+# Detect whether imgui/qtimgui packages are available from the system.
+find_path(IMGUI_INCLUDE_DIR NAMES imgui.h PATH_SUFFIXES imgui)
+find_path(QTIMGUI_INCLUDE_DIR NAMES QtImGui.h PATH_SUFFIXES qtimgui)
+find_library(IMGUI_LIB NAMES imgui)
+find_library(QTIMGUI_LIB NAMES qt_imgui_widgets)
+if (IMGUI_INCLUDE_DIR AND IMGUI_LIB)
+    message("Using system imgui library:
+    IMGUI_INCLUDE_DIR=${IMGUI_INCLUDE_DIR}
+    IMGUI_LIB=${IMGUI_LIB}")
+else()
+    message("Using bundled imgui library")
+    set(IMGUI_INCLUDE_DIR ${CMAKE_SOURCE_DIR}/3rdparty/imgui)
+    set(IMGUI_SOURCES
+        ../3rdparty/imgui/imgui_widgets.cpp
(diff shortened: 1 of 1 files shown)


whitequark:
Why is this necessary? The bundled imgui works just fine, so to me this looks like a complication of the build procedure for no benefit.

Apteryks:
As a policy for packaging on many distributions, bundling libraries is discouraged for various (and good) reasons, where linking to system provided libraries is the preferred approach.  The contributed patch here was made in the context of GNU Guix.  I'm sure it's a build system change that would be welcome and useful to Debian and other distributions too.

The proposed change doesn't impact the status quo in case no system libraries are available for imgui and qtimgui (the build system falls back to use the bundled source copies).

whitequark:
It complicates the build system and introduces combinations of configurations that would have never been seen by anyone building the upstream code. If Debian or Guix want to make this poor policy choice they can also carry the patch downstream.

Apteryks:
Using my distribution maintainer's hat, I disagree about qualifying the policy as "poor": it's applied system wide to Debian, Guix, etc. and helps keeping running systems secure (imagine reviewing and patching security sensitive problems *per package* where each of them may bundle N subprojects -- a maintenance nightmare). But it's your call to make; if you don't like the change, feel free to close this.  The patch is already carried in Guix and will remain so.  It's been tested working with imgui 1.86.

whitequark:
I'm familiar with the challenges involved in maintaining a Linux distribution. I disagree with the blanket nature of the policy philosophically: since nextpnr isn't security-critical software (it is not intended to ever consume any untrusted input) there is no security benefit for unbundling dependencies.

whitequark:
Whether that's useful is in the eye of the beholder. As a package maintainer, if Debian replaces a library that I want to be at a specific version test against it with some other version (that might not even correspond to anything that project has ever released) because of a blanket policy, and issues arise, that should be Debian's problem, not mine. But in practice people will report any issues they encounter to me and often not even mention that they're using a specific distribution's package (much less the fact that this package has been modified by a third party in unknown ways).

I don't think distributions should be applying changes on a whim and without understanding or asking why something is done in a certain way simply because this way the dependency graph looks prettier, and when something similar came up with Debian in the past I've concluded that I might be better off not having software I maintain packaged in Debian.

gatecat:
Agreed, the nextpnr build system is complicated enough without having to pander to these kinds of requirements as well.

PR #1483, opened 2025-05-13
GateMate clock router

Ravenslofty:
With the unique Global Mesh of the CologneChip GateMate, it is vitally important to have good clock routing to avoid hold violations through clock skew.

While this can certainly be improved through better timing analysis, it is already a significant improvement over the current state of the gatemate uarch.

### DIFF
--- a/himbaechel/uarch/gatemate/route_clock.cc
+++ b/himbaechel/uarch/gatemate/route_clock.cc
@@ -143,11 +143,12 @@ void GateMateImpl::route_clock()
                 dest = ctx->getPipDstWire(uh);
                 if (ctx->getBoundWireNet(dest) == glb_net) {
                     NPNR_ASSERT(glb_net->wires.at(dest).pip == uh);
-                    break;
-                }
+                    log_info("                 pip %s --> %s (plane %hhd)\n", ctx->nameOfPip(uh), ctx->nameOfWire(dest), pip_plane(uh));
+                } else {
                 //if (ctx->debug)
                     log_info("            bind pip %s --> %s (plane %hhd)\n", ctx->nameOfPip(uh), ctx->nameOfWire(dest), pip_plane(uh));
-                ctx->bindPip(uh, glb_net, STRENGTH_LOCKED);
+                    ctx->bindPip(uh, glb_net, STRENGTH_LOCKED);
+                }
                 if (dest == sink_wire)
                     break;
             }
--- a/himbaechel/uarch/gatemate/route_clock.cc
+++ b/himbaechel/uarch/gatemate/route_clock.cc
@@ -1,0 +1,127 @@
+/*
+ *  nextpnr -- Next Generation Place and Route
+ *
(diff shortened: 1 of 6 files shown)


himbaechel/uarch/gatemate/route_clock.cc
```diff
@@ -0,0 +1,162 @@
...
+                auto fnd_wire = glb_net->wires.find(curr.wire);
+                if (fnd_wire != glb_net->wires.end()) {
+                    bound_pip = fnd_wire->second.pip;
+                }
+
+                for (auto uh : ctx->getPipsUphill(curr.wire)) {
+                    if (!ctx->checkPipAvailForNet(uh, glb_net))
+                        continue;
+                    WireId src = ctx->getPipSrcWire(uh);
+                    if (backtrace.count(src))
```
gatecat, line 121: this means you won't do any revisits if you subsequently find a cheaper route to a wire - is this intentional?

himbaechel/uarch/gatemate/route_clock.cc
```diff
@@ -0,0 +1,162 @@
...
+    bool operator>(const QueuedWire &rhs) const { return this->delay > rhs.delay; }
+};
+
+} // namespace
+
+void GateMateImpl::route_clock()
+{
+    auto clk_nets = std::vector<NetInfo *>{};
+
+    auto feeds_clk_port = [](PortRef &port) {
```
gatecat, line 50: do clocks for BRAMs etc feed through a CPE? or should they be considered here too?

himbaechel/uarch/gatemate/route_clock.cc
Ravenslofty: Because there's no lookahead, this router is essentially Dijkstra, so I'm pretty sure the first route we find is the cheapest.

himbaechel/uarch/gatemate/route_clock.cc
mmicko: BRAMs have direct global CLK connection, so no need to handle it from here. Just CPEs need it, GPIO and BRAMs have those routed signals directly

himbaechel/uarch/gatemate/route_clock.cc
gatecat: ah right, I missed that, thanks for explaining

himbaechel/uarch/gatemate/route_clock.cc
gatecat: thanks for confirming!

mmicko: There are additional CLK inputs (going trough CPE) that we can use in case that is a problem.

Ravenslofty: I think the only way we could feasibly discover that is with timing analysis...

gatecat: Indeed - just the first thing that comes to mind with an arrangement like this

PR #1479, opened 2025-05-12
Use Windows resources, even if compiler supports #embed.

cr1901:
Recent versions of `g++` available on Windows have `#embed` implemented. `#embed` currently takes priority over Windows resources. However, Rrght now, `#embed` support is not implemented for Windows because of various `#ifdef` gating and [Boost using `wchar_t`](https://www.boost.org/doc/libs/1_70_0/libs/filesystem/doc/reference.html#Operating-system-examples) C-strings for its paths on Windows, whereas `bbasm` otherwise emits UTF-8.

```
[1/11] Building CXX object bba/CMakeFiles/bbasm.dir/main.cc.obj
C:/msys64/home/William/Projects/FPGA/nextpnr/bba/main.cc: In function 'int main(int, char**)':
C:/msys64/home/William/Projects/FPGA/nextpnr/bba/main.cc:455:37: warning: format '%s' expects argument of type 'char*', but argument 3 has type 'const boost::filesystem::path::value_type*' {aka 'const wchar_t*'} [-Wformat=]
  455 |         fprintf(fileOut, "#embed \"%s\"\n", boost::filesystem::path(files.at(2)).c_str());
      |                                    ~^       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                     |                                                 |
      |                                     char*                                             const boost::filesystem::path::value_type* {aka const wchar_t*}
      |                                    %ls
```

I made a patch for the "convert from UTF-16 to UTF-8" functionality that seems to work properly:

```diff
diff --git a/bba/main.cc b/bba/main.cc
index 735dc499..0bed723f 100644
--- a/bba/main.cc
+++ b/bba/main.cc
@@ -30,6 +30,14 @@
 #include <string>
 #include <vector>
 
+#ifdef _WIN32
+    #include <stringapiset.h>
+    /* Somehow, TokenType is used in stringapiset.h, at least when compiling
+       with g++. Move our identifier out of Windows' way. */
+    #define TokenType TokenType_
+#endif
+
+
 enum TokenType : int8_t
 {
     TOK_LABEL,
@@ -452,7 +460,24 @@ int main(int argc, char **argv)
             fprintf(fileOut, "%s\n", s.c_str());
 
         fprintf(fileOut, "const uint8_t %s[%d] = {\n", streams[0].name.c_str(), int(data.size()) + 1);
[...]

PR #1484, opened 2025-05-14
bba: use `std::filesystem` instead of `boost::filesystem`

whitequark:
Also, convert paths to UTF-8 for Windows builds. See #1479.


gatecat:
Thanks for sorting this!

PR #1485, opened 2025-05-14
Set CXX standard for BBA  and remove boost lib

mmicko:
This fixes use case (as in oss-cad-suite) where BBA tool is built as separate project out of this CMakeList.txt file.
It also removes not needed boost library dependency.

### DIFF
--- a/bba/CMakeLists.txt
+++ b/bba/CMakeLists.txt
@@ -1,15 +1,15 @@
 cmake_minimum_required(VERSION 3.25)
 project(bba CXX)
 
+set(CMAKE_CXX_STANDARD 17)
+
 find_package(Boost REQUIRED COMPONENTS
     program_options
-    filesystem
     system)
 
 add_executable(bbasm
     main.cc)
 target_link_libraries(bbasm LINK_PRIVATE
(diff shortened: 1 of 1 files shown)


mmicko:
This is what you get on Ubuntu 22.04 for example, even if you change code to use experimental one, still fails linking. 
Compiling it as C++17 fixes both of the problems, and our main CMakeList.txt is already enforcing C++17

```
/work/_builds/linux-x64/nextpnr-bba/nextpnr/bba/main.cc: In function 'int main(int, char**)':
/work/_builds/linux-x64/nextpnr-bba/nextpnr/bba/main.cc:455:50: error: 'std::filesystem' has not been declared
  455 |         fprintf(fileOut, "#embed \"%s\"\n", std::filesystem::path(files.at(2)).u8string().c_str());
      |                                                  ^~~~~~~~~~
make[2]: *** [CMakeFiles/bbasm.dir/build.make:79: CMakeFiles/bbasm.dir/main.cc.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:87: CMakeFiles/bbasm.dir/all] Error 2
make: *** [Makefile:91: all] Error 2

```

PR #1489, opened 2025-05-15
timing: Add `TimingAnalysis::get_clock_skew_paths`

rowanG077:
Which is useful if you want to optimize clock skews and need the FFs which are effected if you change
a clock delay by re-routing

### DIFF
--- a/common/kernel/timing.cc
+++ b/common/kernel/timing.cc
@@ -788,6 +788,88 @@ void TimingAnalyser::compute_criticality()
     }
 }
 
+std::vector<ClockSkewPaths> TimingAnalyser::get_clock_skew_paths()
+{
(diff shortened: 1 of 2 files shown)


rowanG077:
@Ravenslofty Please check if you are oke with this API. You still have to calculate the skew your self per pair. But since if you have let's say 100 startpoints and 100 endpoints then there are 100*100 clock skew values. So I don't think it makes sense to store this in the struct itself. You can recover the skew pretty easily by just subtracting the delays.

Ravenslofty:
I'm gonna have to think on this a bit; it's a little too late in the day for me to picture how this will work.

rowanG077:
I don't think this is required anymore so I will close it

PR #1495, opened 2025-05-23
timeout for router1

KelvinChung2000:
Adding a timeout option for router1


gatecat:
Timeouts based on wall time are a bad idea, in my opinion, because they are inherently non-deterministic. Maybe you could look at an iteration limit instead?

KelvinChung2000:
I have updated to an iteration-based approach. Is this looking good for you?

PR #1502, opened 2025-06-17
Add sources to detailed timing report

TheZoq2:
I want to do some experiments with printing timing reports in a different format, and noticed that the source information for nets is not available in the json report.

For parsing the sources, I just lifted the code out of `common/kernel/timing_log.cc`


gatecat:
Thanks for adding this!

PR #1503, opened 2025-06-23
Use improved CPE model

mmicko:
- This one updates pack so LUT tree part,FF part and RAMIO part of CPE halfs are now split.
- We do joining of CPE LUT tree parts into one, so simulation for MX4, L2T5 is possible
- Renaming of parameters is now done in bitstream write

NOTE: Still some work to do, but comments are welcome

### DIFF
--- a/himbaechel/uarch/gatemate/pack.cc
+++ b/himbaechel/uarch/gatemate/pack.cc
@@ -302,7 +302,7 @@ void GateMateImpl::pack()
     packer.pack_misc();
     packer.pack_ram();
     packer.pack_serdes();
-    packer.pack_mult();
+    //packer.pack_mult();
     packer.pack_addf();
     packer.pack_cpe();
     packer.remove_constants();


gatecat:
How does splitting up the LUT and FF affect performance?

mmicko:
@gatecat It does perform bit slower but in range of 10%, so nothing major, but solves some other issues like ability to add better timing models and also to perform post pnr equivalence checking easier.

himbaechel/uarch/gatemate/bitstream.cc
```diff
@@ -223,7 +221,7 @@ struct BitstreamBackend
     {
         ChipConfig cc;
         cc.chip_name = device;
-        int bank[uarch->dies][9] = {0};
+        std::vector<std::vector<int>> bank(uarch->dies, std::vector<int>(9, 0));
```
gatecat: just to note, if you want something fixed size while still being more C++-ish, you can also use `std::array`

PR #1509, opened 2025-06-28
himbaechel: xilinx: misc `CMakeLists.txt` improvements

jmi2k:
This PR brings `himbaechel/uarch/xilinx/CMakeLists.txt` closer to its counterparts.

1. Replace hardcoded `/usr/bin/python3` with the more common `${Python3_EXECUTABLE}`, which can be set when calling `cmake`.
2. Recognize `IMPORT_BBA_FILES`, making it possible to use `.bba` files without setting `HIMBAECHEL_PRJXRAY_DB`.
3. Recognize "all" as a valid value for `HIMBAECHEL_XILINX_DEVICES`, defaulting to the list of devices that can be built from [f4pga/prjxray-db](https://github.com/f4pga/prjxray-db).

These changes were made while trying to enable support for Xilinx devices on a fork of `oss-cad-suite-build`. I've split them all in different commits to facilitate rejecting any of them separately, but I could also squash them together if that's desirable.

### DIFF
--- a/himbaechel/uarch/xilinx/CMakeLists.txt
+++ b/himbaechel/uarch/xilinx/CMakeLists.txt
@@ -35,7 +35,8 @@ message(STATUS "Enabled Himbaechel-Xilinx devices: ${HIMBAECHEL_XILINX_DEVICES}"
 foreach (device ${HIMBAECHEL_XILINX_DEVICES})
     add_bba_produce_command(
         TARGET  nextpnr-himbaechel-xilinx-bba
-        COMMAND /usr/bin/pypy3 ${CMAKE_CURRENT_SOURCE_DIR}/gen/xilinx_gen.py
+        COMMAND ${Python3_EXECUTABLE}
+            ${CMAKE_CURRENT_SOURCE_DIR}/gen/xilinx_gen.py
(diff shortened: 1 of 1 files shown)


omixann:
I can change the default value to use when `Python3_EXECUTABLE` is not overridden. Should I do the same for all other occurrences of `Python3_EXECUTABLE` or just for this one?

gatecat:
Only this one

omixann:
Turns out `Python3_EXECUTABLE` is provided by `pybind11`, so there is no way of both allowing overriding it while making it default to `pypy3` _only_ here (or, at least, it's not something I know how to, my experience with CMake is limited).

I've made an exception here to use `PYTHON` instead. It's not ideal but it's an improvement.

gatecat:
You could also create a new variable like `XilinxChipdb_Python3_EXECUTABLE`, anyone that doesn't want to use pypy3 or wants to override the path to it can override that?

omixann:
In theory `PYTHON` is that new variable you're describing already, but I see how a more descriptive name is a better idea than just `PYTHON` for this specific purpose. I'll make the change.

omixann:
Are the current changes OK?

gatecat:
Yep, thanks!

PR #1513, opened 2025-07-01
himbaechel: gatemate: replace VLA with C++ features

jmi2k:
Apparently `clang` doesn't like VLAs that much, and neither do I or C++ in general (they are not in the language spec). Using vectors is not exactly identical but it should work the same.

After these changes, [`nextpnr-himbaechel` can be build for `darwin-arm64`](https://github.com/jmi2k/oss-cad-suite-build/actions/runs/16005189044).

Unfortunately, I don't own a device to test these changes (neither an Apple device nor a Gatemate FPGA) so these changes are not tested.

### DIFF
--- a/himbaechel/uarch/gatemate/bitstream.cc
+++ b/himbaechel/uarch/gatemate/bitstream.cc
@@ -223,7 +223,7 @@ struct BitstreamBackend
     {
         ChipConfig cc;
         cc.chip_name = device;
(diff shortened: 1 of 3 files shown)


mmicko (approved):

mmicko:
Thanks, this is what we actually did on other branch. Will merge this and make code more like this change

PR #1516, opened 2025-07-10
machxo3d: minor fixes for I3C pins and RPLL

via:
When using nextpnr and trying to build a design that uses the pins that have I3C functionality, for example `LCMXO3D-9400HC-6SG72I`, I would get an error:
```
Terminate called after throwing an instance of 'nextpnr_machxo2::assertion_failure'
  what():  Assertion failure: no tile at (0, 29) with type in set (/home/via/dev/nextpnr/machxo2/arch.h:1026)
Aborted
```

I get a similar error when using RPLL, and after debugging I saw that the GPLL_R0 location was close but didn't match. I'm assuming this is just a typo and the second PLL hasn't been tested, but I don't know much about these things and don't know if this fix is appropriate (or might break other machxo(2) chips)


mmicko:
Thanks for looking into this. I3C change is fine for sure, but I am thinking that you are right about GPLL_R0 as well and that it was my typo

PR #1517, opened 2025-07-11
gatemate: optimizations and cleanups

mmicko:
This adds couple of optimizations and cleanups for GateMate architecture.

Minor: For CC_LUT1/2 we do constant or signal propagation in case of lut function is set like that. This is handled by Yosys but in case some parts are manually instantiated it can be optimized. Similar is done for MX2/MX4.

Similar to above DFFs and Latches are also optimized in case they are in special configuration.

Major:
To get better packing LUT2s are merged in ADDF/ADDF2 and FFs as well.
For L2T4s we are merging two DFFs where possible as well.

Replaced some magic numbers with enumerations and also gnd and vcc nets  now have separate variable holding their value.

Note: multipliers are not touched.

### DIFF
--- a/himbaechel/uarch/gatemate/pack.cc
+++ b/himbaechel/uarch/gatemate/pack.cc
@@ -26,13 +26,13 @@
 
 NEXTPNR_NAMESPACE_BEGIN
 
-void GateMatePacker::flush_cells(bool unbind)
+void GateMatePacker::flush_cells()
 {
     for (auto pcell : packed_cells) {
         for (auto &port : ctx->cells[pcell]->ports) {
             ctx->cells[pcell]->disconnectPort(port.first);
         }
-        if (unbind)
+        if (ctx->cells[pcell]->bel != BelId())
             ctx->unbindBel(ctx->cells[pcell]->bel);
         ctx->cells.erase(pcell);
     }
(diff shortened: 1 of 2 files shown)


himbaechel/uarch/gatemate/pack.cc
```diff
@@ -26,196 +26,28 @@
...
 NEXTPNR_NAMESPACE_BEGIN
 
-void GateMatePacker::flush_cells()
+void GateMatePacker::flush_cells(bool unbind)
 {
     for (auto pcell : packed_cells) {
         for (auto &port : ctx->cells[pcell]->ports) {
             ctx->cells[pcell]->disconnectPort(port.first);
         }
+        if (unbind)
```
gatecat: is there any reason to make `unbind` optional? I would just make this conditional on `ctx->cells[pcell]->bel != BelId()`, erasing a cell without unbinding it is always going to be a problem...

himbaechel/uarch/gatemate/pack.cc
```diff
@@ -283,6 +115,272 @@ void GateMatePacker::copy_constraint(NetInfo *in_net, NetInfo *out_net)
...
+        usr.cell->disconnectPort(port);
+        usr.cell->connectPort(port, to_net);
+    }
+}
+
+void GateMatePacker::count_cell(CellInfo &ci)
+{
+    packed_cells.insert(ci.name);
+    if (!count_per_type.count(ci.type))
+        count_per_type[ci.type] = 1;
```
gatecat: I don't think there's any need for this whole conditional - `count_per_type[ci.type]` will create an entry with default (0) value in the dictionary if `ci.type` isn't yet in there, so just `count_per_type[ci.type]++;` unconditionally will always work

mmicko:
@gatecat @Ravenslofty is it possible that one of you take care of these changes ? I will be away for longer period and do not have computer here. Thanks.

Issue #1515, opened 2025-07-08
[Reature Request] Automatic assignment of 'regular' clock signal to a global clock net.


zapta:
EDIT: This issue is now a feature request to have an automatic assignment of a clock signal to a global clock net even if it's a 'regular' input pin.

Transferring from this discussion https://github.com/YosysHQ/yosys/discussions/5214

nextpnr-himbaechel fails with an heavily skewed clock that violates min hold time.

```
⦗OSS CAD Suite⦘ ~/work4$ yosys -p "synth_gowin -top main -json hardware.json" -q main.v

⦗OSS CAD Suite⦘ ~/work4$ nextpnr-himbaechel --device GW2AR-LV18QN88C8/I7 --json hardware.json --write hardware.pnr.json --report hardware.pnr --vopt family=GW2A-18C --vopt cst=pinout.cst -q --freq 1
ERROR: Hold/min time violation for clock 'posedge CLK_IBUF_I_O':
ERROR: Hold/min time violation for clock 'posedge CLK_IBUF_I_O':
0 warnings, 2 errors

⦗OSS CAD Suite⦘ ~/work4$ 
```

main.v
```
module main (
    input  CLK,  // System clock
    output LED   // Output
);

  localparam WIDTH = 256;

  reg [WIDTH-1:0] counter = 0;

  assign LED = counter[WIDTH-1];

  always @(posedge CLK) begin
    counter <= counter + 7;
  end

endmodule
```

pinout.cst
```
IO_LOC "CLK"  4;
IO_PORT "CLK" PULL_MODE=UP BANK_VCCIO=1.8;;

IO_LOC "LED"  15; // Active low
IO_PORT "LED" PULL_MODE=UP DRIVE=8 BANK_VCCIO=1.8;
```

yrabbit:
If you look at Tangnano20k's schematic, pin 4 is not a dedicated clock pin - yes, it says SYS_CLK next to it, but it's nothing more than the name of a wire on the board. It's a general purpose pin, among its alternative functions is connection to the PLL input and that's the end of it.

<img width="800" height="600" alt="Image" src="https://github.com/user-attachments/assets/1db0215e-c380-46e4-a0b9-dc9c0b2cbf3d" />

Therefore, a general purpose router is used, which will result in many effects if applied to a clock. 

There are several ways out of this situation of varying degrees of badness:

[...]

PR #1518, opened 2025-07-12
Gowin. Allow clock network routing from GP pins.

yrabbit:
Adds automatic connection of a general-purpose pin to the global clock network.

The old behaviour, where such networks have to be explicitly specified, can be activated with the command line key
"--vopt disable_gp_clock_routing".

Addresses https://github.com/YosysHQ/nextpnr/issues/1515


gatecat:
Thanks for dealing with this one!

PR #1520, opened 2025-07-22
Gowin. Preparing to support the 5A series.

yrabbit:
Family recognition is added, as well as minor fixes, but base generation itself is not allowed for GW5A - this gives the ability to test the next Apicula release and still not break installations for those who simply specify `HIMBAECHEL_GOWIN_DEVICES = "all"`.

### DIFF
--- a/himbaechel/uarch/gowin/gowin.cc
+++ b/himbaechel/uarch/gowin/gowin.cc
@@ -129,14 +129,19 @@ void GowinImpl::init_database(Arch *arch)
         if (GW2) {
             log_error("For the GW2A series you need to specify --vopt family=GW2A-18 or --vopt family=GW2A-18C\n");
         } else {
-            std::regex devicere = std::regex("GW1N([SZ]?)[A-Z]*-(LV|UV|UX)([0-9])(C?).*");
-            std::smatch match;
-            if (!std::regex_match(args.device, match, devicere)) {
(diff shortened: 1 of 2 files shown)


himbaechel/uarch/gowin/gowin.cc
```diff
@@ -129,14 +129,19 @@ void GowinImpl::init_database(Arch *arch)
...
-            std::regex devicere = std::regex("GW1N([SZ]?)[A-Z]*-(LV|UV|UX)([0-9])(C?).*");
-            std::smatch match;
-            if (!std::regex_match(args.device, match, devicere)) {
-                log_error("Invalid device %s\n", args.device.c_str());
-            }
-            family = stringf("GW1N%s-%s", match[1].str().c_str(), match[3].str().c_str());
-            if (family.rfind("GW1N-9", 0) == 0) {
-                log_error("For the GW1N-9 series you need to specify --vopt family=GW1N-9 or --vopt family=GW1N-9C\n");
+            if (args.device.rfind("GW5A", 0) == 0) {
+                family = "GW5A-25A";
```
gatecat: might be a good idea to have a more thorough check here, even if the other GW5A parts aren't supported, an error would be better than it silently becoming a -25A

yrabbit: Fixed.

PR #1524, opened 2025-07-31
gatemate: multiplier router

Ravenslofty:
I am not the proudest of this code, but, it's a very important step towards getting working multipliers (again) as it removes routing variability, ensuring inversion consistency across runs.

EDIT: and apparently it fixes the multipliers 🎉


himbaechel/uarch/gatemate/gatemate.h
```diff
@@ -22,6 +22,7 @@
 
 #include "extra_data.h"
 #include "himbaechel_api.h"
+#include "idstring.h"
```
gatecat: is this actually needed? it should come from `nextpnr.h` anyway

himbaechel/uarch/gatemate/route_mult.cc
```diff
@@ -0,0 +1,826 @@
...
+        find_and_bind_downhill_pip(ctx, sb_big_ydiag, in_mux, net);
+    }
+
+    route_mult_diag(ctx, net, Loc{loc.x + 1, loc.y + 1, 0}, in_mux, 8);
+}
+} // namespace
+
+NEXTPNR_NAMESPACE_BEGIN
+
+void GateMateImpl::route_mult()
```
gatecat, line 721: It would be really nice to have some brief rationale behind this (fixing the routing so the inversion is correct) and ideally sketch ASCII art diagrams or something, so this code isn't quite as opaque to whoever stumbles across it in the future

himbaechel/uarch/gatemate/route_mult.cc
```diff
@@ -0,0 +1,826 @@
...
+namespace {
+USING_NEXTPNR_NAMESPACE;
+
+void find_and_bind_downhill_pip(Context *ctx, WireId from, WireId to, NetInfo *net)
+{
+    NPNR_ASSERT(from != WireId());
+    NPNR_ASSERT(to != WireId());
+    for (auto pip : ctx->getPipsDownhill(from)) {
+        if (ctx->getPipDstWire(pip) == to) {
+            log_info("    pip %s: %s -> %s\n", ctx->nameOfPip(pip), ctx->nameOfWire(from), ctx->nameOfWire(to));
```
gatecat: this feels like the kind of thing that should be gated behind `ctx->debug` or `ctx->verbose` ?

himbaechel/uarch/gatemate/gatemate.h
Ravenslofty: clangd helpfully inserts header includes for me; I'll try to clean that up.

himbaechel/uarch/gatemate/route_mult.cc
Ravenslofty: will do, although it's as much about routing consistency as anything; the x2y2 routes are all inverting, so some fix-up has to be done anyway.

himbaechel/uarch/gatemate/route_mult.cc
Ravenslofty: yeah this is presently incredibly noisy and so most of the debug prints will be gated.

mmicko:
@gatecat cleaned as much as I could. We may see to improve that route_mult but it is working fine.  whole point of this router is that ZERO and A_PASSTHRU signals are routed to MULT non-inverted so it requires special route to do so, and then diagonal connections to next levels, that invert.

PR #1532, opened 2025-08-14
Gowin. Use two additional DFFs.

yrabbit:
The GW-5A series has 8 flip-flops in a cell instead of 6. These additional flip-flops can be used if the control network matches that for the 4th and 5th DFFs in this cell.

### DIFF
--- a/himbaechel/uarch/gowin/gowin.cc
+++ b/himbaechel/uarch/gowin/gowin.cc
@@ -1060,7 +1060,7 @@ bool GowinImpl::slice_valid(int x, int y, int z) const
             }
         }
         // Check whether the current architecture allows 6 and 7 DFFs
-        if (z > 3 && ctx->getBelByLocation(Loc(x, y, 6 * 2 + 1)) != BelId()) {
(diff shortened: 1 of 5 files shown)


himbaechel/uarch/gowin/gowin.cc
```diff
@@ -1059,6 +1059,28 @@ bool GowinImpl::slice_valid(int x, int y, int z) const
                 return false;
             }
         }
+        // Check whether the current architecture allows 6 and 7 DFFs
+        if (z > 3 && ctx->getBelByLocation(Loc(x, y, 6 * 2 + 1)) != BelId()) {
```
Ravenslofty: I feel like there has to be a better way of doing this, personally. In the gatemate arch we have an "extra data" type associated with logic, and you would set a flag that says "this arch has 6/7 DFFs" or so. But, if it works, why not I suppose.

yrabbit: Well, in principle, it can be changed :)

Ravenslofty (approved):
ping me on matrix when CI is green.

PR #1533, opened 2025-08-14
gatemate: add IOSEL as separate primitive

mmicko:
Separate IO selection from rest of GPIO, to simplify code


mmicko:
Requires https://github.com/YosysHQ/prjpeppercorn/pull/6

PR #1537, opened 2025-08-21
gatemate: add iopath delays

mmicko:
Thing that is missing in PR is proper RAM clock assignment for IOPATHs. This takes data already available in prjpeppercorn so no chip database data version is required to be changed.

More work is pending for RAM so thinking of adding that info when RAM is handled as halfs instead of here.


himbaechel/uarch/gatemate/gatemate.cc
```diff
@@ -36,8 +36,55 @@ void GateMateImpl::init_database(Arch *arch)
     init_uarch_constids(arch);
     arch->load_chipdb(stringf("gatemate/chipdb-%s.bin", args.device.c_str()));
     arch->set_package("FBGA324");
-    arch->set_speed_grade("DEFAULT");
     dies = std::stoi(args.device.substr(6));
+    fpga_mode = 3;
+    timing_mode = 3;
+    if (args.options.count("fpga_mode"))
+        fpga_mode = std::stoi(args.options.at("fpga_mode"));
+    if (args.options.count("om"))
```
Ravenslofty: do we *need* to use the old and kind of obscure parameter here?

mmicko: Removed that one now, we can easy add back if needed for compatibility in future

Ravenslofty: I also feel like it might be useful to also have human-readable strings as well as the numbers.

Ravenslofty (approved):

PR #1538, opened 2025-08-21
gatemate: use CPE bridge

mmicko:
TODO: 
- split net that is actually used by CPE_BRIDGE and connect input and output pins, so we can use it for post pnr simulation.
- create separate PR for router2 changes
- bump required chipdb version

### DIFF
--- a/himbaechel/uarch/gatemate/gatemate.cc
+++ b/himbaechel/uarch/gatemate/gatemate.cc
@@ -368,7 +368,6 @@ void GateMateImpl::reassign_bridges(NetInfo* ni, const dict<WireId, PipMap>& net
 
 void GateMateImpl::postRoute()
 {
-#if 1
     int num = 0;
 
     pool<IdString> nets_with_bridges;
@@ -437,26 +436,6 @@ void GateMateImpl::postRoute()
     }
 
     log_info("%d MUX8s used for routing (should be %d)\n", num, int(cpe_bridges.size()));
-#else
-    for (auto &net : ctx->nets) {
-        NetInfo *ni = net.second.get();
-        for (auto &w : ni->wires) {
-            if (w.second.pip != PipId()) {
-                const auto &extra_data = *reinterpret_cast<const GateMatePipExtraDataPOD *>(
-                        chip_pip_info(ctx->chip_info, w.second.pip).extra_data.get());
-                if (extra_data.type == PipExtra::PIP_EXTRA_MUX && (extra_data.flags & MUX_ROUTING)) {
-                    IdStringList id = ctx->getPipName(w.second.pip);
-                    Loc loc = ctx->getPipLocation(w.second.pip);
-                    BelId bel = ctx->getBelByLocation({loc.x, loc.y, CPE_BRIDGE_Z});
-                    CellInfo *cell = ctx->createCell(ctx->id(ctx->nameOfBel(bel)), id_CPE_BRIDGE);
-                    ctx->bindBel(bel, cell, PlaceStrength::STRENGTH_FIXED);
-                    cell->params[id_C_BR] = Property(Property::State::S1, 1);
-                    cell->params[id_C_SN] = Property(extra_data.value, 3);
-                }
-            }
-        }
-    }
-#endif
     ctx->assignArchInfo();
 
     const ArchArgs &args = ctx->args;
--- a/himbaechel/uarch/gatemate/gatemate.cc
+++ b/himbaechel/uarch/gatemate/gatemate.cc
@@ -278,6 +278,22 @@ void GateMateImpl::postPlace()
 {
     repack();
     ctx->assignArchInfo();
+    for (auto &cell : ctx->cells) {
+        Loc loc = ctx->getBelLocation(cell.second.get()->bel);
+        if (loc.z <= CPE_BRIDGE_Z)
+            used_cpes.emplace(ctx->getBelName(cell.second.get()->bel)[0]);
+    }
+}
+bool GateMateImpl::checkPipAvail(PipId pip) const
+{
+    const auto &extra_data =
+            *reinterpret_cast<const GateMatePipExtraDataPOD *>(chip_pip_info(ctx->chip_info, pip).extra_data.get());
+    if (extra_data.type != PipExtra::PIP_EXTRA_MUX || (extra_data.flags & MUX_ROUTING) == 0)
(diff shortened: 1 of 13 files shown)


himbaechel/uarch/gatemate/gatemate.cc
```diff
@@ -181,7 +181,31 @@ bool GateMateImpl::getClusterPlacement(ClusterId cluster, BelId root_bel,
...
+bool GateMateImpl::checkPipAvail(PipId pip) const
+{
+    const auto &extra_data =
+            *reinterpret_cast<const GateMatePipExtraDataPOD *>(chip_pip_info(ctx->chip_info, pip).extra_data.get());
+    if (extra_data.type != PipExtra::PIP_EXTRA_MUX || (extra_data.flags & MUX_ROUTING) == 0)
+        return true;
+    IdStringList id = ctx->getPipName(pip);
+    if (used_cpes.count(id[0]))
+        return false;
+    return true;
```
Ravenslofty, line 207:
```suggestion
    return BaseArch::checkPipAvail(pip);
```

I think this is safer to avoid accidentally marking pips available where they wouldn't be before.

himbaechel/uarch/gatemate/gatemate.cc
```diff
@@ -261,6 +263,45 @@ void GateMateImpl::postPlace()
...
     ctx->assignArchInfo();
+    for (auto &cell : ctx->cells) {
+        // We need to skip CPE_MULT since using CP outputs is mandatory
+        // even if output is actually not connected
+        bool marked_used = cell.second.get()->type == id_CPE_MULT;
+        // Can not use FF for OUT2 if CPE is used in bridge mode
+        if (cell.second.get()->type == id_CPE_FF && ctx->getBelLocation(cell.second.get()->bel).z == CPE_FF_U_Z)
+            marked_used = true;
+        if (marked_used)
+            used_cpes.emplace(ctx->getBelName(cell.second.get()->bel)[0]);
```
gatecat: couldn't this be simplified by just having a pool of tile indices instead and using `bel.tile` in place of `ctx->getBelName(cell.second.get()->bel)[0]` ?

himbaechel/uarch/gatemate/gatemate.cc
```diff
@@ -261,6 +263,45 @@ void GateMateImpl::postPlace()
...
+        chip_pip_info(ctx->chip_info, pip).extra_data.get());
+
+    if (extra_data.type != PipExtra::PIP_EXTRA_MUX || !(extra_data.flags & MUX_ROUTING))
+        return;
+
+    auto found = cpe_bridges.count(pip) > 0;
+
+    NPNR_ASSERT((!found && net != nullptr) || (found && net == nullptr));
+
+    if (found)
```
gatecat: if `cpe_bridges` is only used post-route it would be more efficient to only build it post-route, too (this is not really relevant for router2 but would be for router1 that binds pips as it goes)

Ravenslofty: (We've given up on router1 in gatemate; since r2 only binds pips once, this is a more efficient - and more elegant - solution, IMO.)

himbaechel/uarch/gatemate/gatemate.cc
```diff
@@ -278,6 +280,26 @@ void GateMateImpl::postPlace()
...
+        // Can not use FF for OUT2 if CPE is used in bridge mode
+        if (cell.second.get()->type == id_CPE_FF && ctx->getBelLocation(cell.second.get()->bel).z == CPE_FF_U_Z)
+            marked_used = true;
+        if (marked_used)
+            used_cpes.emplace(cell.second.get()->bel.tile);
+    }
+}
+bool GateMateImpl::checkPipAvail(PipId pip) const
+{
+    const auto &extra_data =
```
gatecat: perhaps a `pip_extra_data` helper function just containing the whole `reinterpret_cast` expression makes sense given this crops up in a few places

himbaechel/uarch/gatemate/gatemate.cc
```diff
@@ -287,8 +309,126 @@ void GateMateImpl::preRoute()
...
+
+        IdStringList id = ctx->getPipName(pip);
+        Loc loc = ctx->getPipLocation(pip);
+        BelId bel = ctx->getBelByLocation({loc.x, loc.y, CPE_BRIDGE_Z});
+        CellInfo *cell = ctx->createCell(name, id_CPE_BRIDGE);
+        ctx->bindBel(bel, cell, PlaceStrength::STRENGTH_FIXED);
+        cell->params[id_C_BR] = Property(Property::State::S1, 1);
+        cell->params[id_C_SN] = Property(extra_data.value, 3);
+
+        IdString new_net_name = ctx->id(name.str(ctx) + "$muxout");
```
gatecat: this can also be a `ctx->idf` to keep things consistent

himbaechel/uarch/gatemate/gatemate.cc
```diff
@@ -287,8 +309,126 @@ void GateMateImpl::preRoute()
...
+        NetInfo *new_net = ctx->createNet(new_net_name);
+        IdString in_port = ctx->idf("IN%d", extra_data.value + 1);
+
+        cell->ports[in_port].name = in_port;
```
gatecat: use `cell->addInput`/`cell->addOutput` for these

himbaechel/uarch/gatemate/gatemate.h
```diff
@@ -83,6 +85,8 @@ struct GateMateImpl : HimbaechelAPI
     pool<IdString> multiplier_a_passthru_uppers;
     pool<IdString> multiplier_zero_drivers;
     std::vector<CellInfo *> multipliers;
+    std::set<int32_t> used_cpes;
```
gatecat: any specific reason for `std::set` rather than `pool`? a flat vector of bool (one index per tile) might even be the most efficient

himbaechel/uarch/gatemate/gatemate.cc
mmicko: Fixed this as well now on multiple other places

mmicko:
All comments have been addressed.

gatecat (approved):
LGTM, squash and merge as soon as you're ready on your side

mmicko:
@gatecat thanks for review

PR #1539, opened 2025-08-21
router2: iteratively reserve arc driver wires, too

Ravenslofty:
When using CPE Bridge Mode in the gatemate uarch (upcoming PR), there can be routing conflicts between logic inside the CPE, and the router using the bridge through an output. Because there are several muxes to navigate between the logic output and the general routing, we iteratively reserve arc drivers until routing is reached to avoid these conflicts.

### DIFF
--- a/common/route/router2.cc
+++ b/common/route/router2.cc
@@ -430,6 +430,23 @@ struct Router2
         }
     }
 
+    // Returns true if a wire contains no sink port
+    bool is_wire_unusable(WireId wire, const NetInfo *net, const pool<WireId>& sink_wires, int iter_count = 0)
+    {
+        if (iter_count > 7)
+            return false; // heuristic to assume we've hit general routing
+        if (wire_data(wire).reserved_net != -1 && wire_data(wire).reserved_net != net->udata)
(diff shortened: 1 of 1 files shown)


gatecat:
If you have a big-ish ECP5 design lying around, could you do me a favour and save me some time by checking that nothing breaks on that?

Ravenslofty:
well, it breaks on the trellisboard design >.>

common/route/router2.cc
```diff
@@ -459,32 +476,61 @@ struct Router2
...
+            WireId cursor = src;
+            bool done = false;
+
+            if (ctx->debug) {
+                log("reserving wires for net %s\n", ctx->nameOf(net));
+                log("   with driver wire %s\n", ctx->nameOfWire(src));
+            }
+
+            while (!done) {
+                auto &wd = wire_data(cursor);
```
gatecat, line 494: it's not related to the problem, but I've spotted that this re-runs unnecessarily for every arc of a net, when the driver of the net is going to be the same for all arcs of a net - maybe reserving the driver should be a different function that only runs once per arc

PR #1542, opened 2025-08-27
Gowin. Add ROM16 primitive.

yrabbit:
The LUTRAM mode is added to all supported chips at once.

This is essentially an alias for LUT4, so the packaging is also moved before searching for LUT-DFF pairs for possible optimization.

In addition to being the only LUTRAM mode in the GW5A series, the addition of ROM16 eliminates the need to manually rename the primitive and its pins when working with files generated by Gowin IDE - a similar situation occurred with INV, which is essentially LUT1.


gatecat:
Are you 100% sure there's no LUTRAM support in the GW5A (beyond ROM16 which is just an alias for a normal LUT)?

yrabbit:
> Are you 100% sure there's no LUTRAM support in the GW5A (beyond ROM16 which is just an alias for a normal LUT)?

They are not available in the GW5A-25A I have or in the TangMega138k I plan to purchase.

However, they may be available in the 60k.

ssram

gatecat:
Thanks for confirming, that's an annoying omission...

PR #1543, opened 2025-08-28
Gowin. Optimize ALU wiring

yrabbit:
Interestingly, although VCC and GND sources are present in each cell, they cannot be connected directly to all LUT inputs. Instead, additional PIPs are used.

A very simple ALU optimization: once we detect that one of the inputs is a constant, we modify the main LUT that describes the ALU function so that this primitive input is ignored, and then disconnect it from the network, freeing up the PIP.
For example (unrealistic, since a real ALU LUT has a larger size and service bits in the middle, etc.), the addition function of A and B when A = 1 is converted from the general case (A isn't a constant and B isn't a constant) to a special case:
0110 -> 0011

The renaming of ALU ports for ADD and SUB modes has also been removed—this has already been done in the chip database as a fixed change to the ALU LUT.

### DIFF
--- a/himbaechel/uarch/gowin/pack.cc
+++ b/himbaechel/uarch/gowin/pack.cc
@@ -1991,7 +1991,7 @@ struct GowinPacker
                     optimized = true;
 
                     // fix the carry
-                    if (!i) {
+                    if (i == 0) {
                         if (inp_net_name == vcc_net_name) {
                             alu_lut |= 0xfU;
                         } else {


himbaechel/uarch/gowin/pack.cc
```diff
@@ -1941,6 +1941,98 @@ struct GowinPacker
...
+                    break;
+                }
+                IdString inp_name = ctx->idf("I%d", i);
+                inp_net_name = ci->getPort(inp_name)->name;
+                if (inp_net_name == vcc_net_name || inp_net_name == gnd_net_name) {
+                    ci->disconnectPort(inp_name);
+                    optimized = true;
+
+                    // fix the carry
+                    if (!i) {
```
gatecat: nitpick - can this be an `if (i == 0)` because that makes the logic here a bit clearer imo

gatecat:
Thanks!

PR #1545, opened 2025-08-31
Gowin. Implement on-chip oscillator.

yrabbit:
A programmable on-chip crystal oscillator has been implemented for the GW5A series.

A critical innovation in this series was the change in the nature of the OSC output pin—it now belongs to the clock wires, and therefore the routes must be made with a special global router, as there is no possibility of using routing through general-purpose PIPs.

At the same time, we are transferring the outputs of all previous generations of OSC to potential clock wires. At the moment, this will not affect the way they are routed - they will still end up as segments as before, but in the future we may optimize the mechanism.

### DIFF
--- a/himbaechel/uarch/gowin/globals.cc
+++ b/himbaechel/uarch/gowin/globals.cc
@@ -65,11 +65,12 @@ struct GowinGlobalRouter
         bool not_dcs_pip = dst_name != id_CLKOUT;
         IdString src_type = ctx->getWireType(src);
         IdString dst_type = ctx->getWireType(dst);
-        bool src_valid = not_dcs_pip && src_type.in(id_GLOBAL_CLK, id_IO_O, id_PLL_O, id_HCLK, id_DLLDLY, id_OSCOUT);
+        bool src_is_outpin = src_type.in(id_IO_O, id_PLL_O, id_HCLK, id_DLLDLY, id_OSCOUT);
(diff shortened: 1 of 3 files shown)


himbaechel/uarch/gowin/globals.cc
```diff
@@ -65,11 +65,11 @@ struct GowinGlobalRouter
         bool not_dcs_pip = dst_name != id_CLKOUT;
         IdString src_type = ctx->getWireType(src);
         IdString dst_type = ctx->getWireType(dst);
-        bool src_valid = not_dcs_pip && src_type.in(id_GLOBAL_CLK, id_IO_O, id_PLL_O, id_HCLK, id_DLLDLY);
+        bool src_valid = not_dcs_pip && src_type.in(id_GLOBAL_CLK, id_IO_O, id_PLL_O, id_HCLK, id_DLLDLY, id_OSCOUT);
         bool dst_valid = not_dcs_pip && dst_type.in(id_GLOBAL_CLK, id_TILE_CLK, id_PLL_I, id_IO_I, id_HCLK);
 
         bool res;
-        if (src == src_wire && (!src_type.in(id_IO_O, id_HCLK, id_DLLDLY_O))) {
+        if (src == src_wire && (!src_type.in(id_IO_O, id_HCLK, id_DLLDLY_O, id_OSCOUT))) {
```
Ravenslofty: small style thing, but maybe it's worth extracting the `src_type.in(...)` into a variable, since this appears in both `src_valid` and this if.

PR #1547, opened 2025-09-02
Gowin. BUGFIX. Add data about gate wires.

yrabbit:
Very rarely (about once a year), the dedicated clock router would malfunction, issuing an incorrect route.

The reason turned out to be the so-called gate wires to the global clock wire system from the logic. Among the PIPs for which these wires are sinks, there are PIPs where the sources are also clock wires.

This leads to the possibility of feeding the clock signal back into the gate and again into the global clock MUX.

If handled carelessly, this can lead to a complete loop.

But the loop option itself is particularly useful in the case of DCS (dynamic clock selection) - the fact is that because these primitives have four clock inputs and each of them could theoretically address all 56 clock sources, but in practice there are not enough wires and the DCS inputs cannot serve as sinks for all clock sources.

The simplest solution (and the one that currently works) is to use the gate to re-enter the clock system, but this time changing the clock source.

This commit explicitly marks wires as gates and removes the possibility of looping (however unlikely it may be) where a loop is not needed.

### DIFF
--- a/himbaechel/uarch/gowin/globals.cc
+++ b/himbaechel/uarch/gowin/globals.cc
@@ -44,6 +44,16 @@ struct GowinGlobalRouter
 
     bool segment_wire_filter(PipId pip) const { return !gwu.is_segment_pip(pip); }
 
+    // To avoid a cycle where we connect the clock wire to the gate in
+    // the global clock system and it ends up in the global clock MUX
+    // again, we only allow connections from general-purpose wires.
+    bool clock_gate_wire_filter(PipId pip) const
+    {
+        WireId dst = ctx->getPipDstWire(pip);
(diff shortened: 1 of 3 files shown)


yrabbit:
Illustration of how DCS inputs (X21Y9/P46B and X18Y9/P36B) receive a signal via a wire that they may not process (from X18Y0/CLK1) by passing it through the gate (X19Y0/CLK1) again.

gates

The costs are negligible since everything happens in global clock wires.

Of course, there is room for optimization here - one can adjust to the capabilities of the DCS inputs by selecting the appropriate gates for the clock signal, but the essence of this commit is to provide information about such wires and fix the bad looping.

gatecat:
Nice fix!

PR #1550, opened 2025-09-03
gatemate: fix fourgroup for multi die

mmicko:
This one use different way of determining in-four-group location, using data from chip database, fixes multiplier router when using CCGM1A2

### DIFF
--- a/himbaechel/uarch/gatemate/route_mult.cc
+++ b/himbaechel/uarch/gatemate/route_mult.cc
@@ -26,6 +26,15 @@
 namespace {
 USING_NEXTPNR_NAMESPACE;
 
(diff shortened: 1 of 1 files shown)


himbaechel/uarch/gatemate/route_mult.cc
```diff
@@ -26,6 +26,15 @@
 namespace {
 USING_NEXTPNR_NAMESPACE;
 
+bool get_fourgroup(const GateMateTileExtraDataPOD *ti, int &x_within_fourgroup, int &y_within_fourgroup)
```
Ravenslofty: I think I would call this `is_fourgroup_a`, because a bool is a lot less meaningful from a function called `get_fourgroup`.

PR #1556, opened 2025-09-16
Gowin. Optimize ALU.

yrabbit:
By replacing the operation of adding the input to itself with a specially formed LUT, we free up two PIPs.


uis246:
Wouldn't it be beneficial to move ALU LUT generation into yosys, where it is free to do more optimizations? For example inverting output or input.

yrabbit:
Maybe, but it's too high a level for me — I don't know my way around there at all :)

PR #1559, opened 2025-09-23
gatemate: Include and use connection timing data

mmicko:
This one includes pip timing data and also changes nodes to be represented as pips so we can use proper timing data for them as well. There is slight drop in performance since router is trying now to find faster route based on real timing info.

Clock router and multiplier router had to be updated to reflect changes in chip database, however multiplier router now actually do have all connections so it also makes more sense.

This one requires https://github.com/YosysHQ/prjpeppercorn/pull/11

### DIFF
--- a/himbaechel/uarch/gatemate/route_clock.cc
+++ b/himbaechel/uarch/gatemate/route_clock.cc
@@ -130,6 +130,7 @@ void GateMateImpl::route_clock()
 
         std::priority_queue<QueuedWire, std::vector<QueuedWire>, std::greater<QueuedWire>> visit;
         dict<WireId, PipId> backtrace;
+        dict<WireId, delay_t> delay_map;
 
         auto is_glb_clk = clk_net->driver.cell->type == id_GLBOUT;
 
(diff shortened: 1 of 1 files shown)


himbaechel/uarch/gatemate/gen/arch_gen.py
```diff
@@ -226,12 +222,30 @@ def main():
...
+        for conn in nodes:
+            if conn.endpoint:
+                t_name = dev.get_tile_type(conn.x,conn.y)
+                if t_name not in new_wires:
+                    new_wires[t_name] = set()
+                new_wires[t_name].add(conn.name)
+                # Check to confirm no duplicates
+                if conn.name in wire_delay:
+                    if wire_delay[conn.name]!=conn.delay:
+                        print(conn.name, conn.delay)
```
gatecat: I think this seems like it should be an assertion rather than a print at this point?

himbaechel/uarch/gatemate/route_clock.cc
```diff
@@ -107,125 +112,101 @@ void GateMateImpl::route_clock()
...
+                sink_wires_to_do.erase(curr.wire);
+                if (sink_wires_to_do.empty())
+                    break;
+            }
+
+            for (auto dh : ctx->getPipsDownhill(curr.wire)) {
+                if (!ctx->checkPipAvailForNet(dh, clk_net))
+                    continue;
+                WireId dst = ctx->getPipDstWire(dh);
+                if (backtrace.count(dst))
```
gatecat: this won't ever do revisits, just to check that this is intentional ?

himbaechel/uarch/gatemate/gen/arch_gen.py
mmicko: agree. updated

himbaechel/uarch/gatemate/route_clock.cc
Ravenslofty: It's Dijkstra without lookahead, so each node is always visited with its lowest-cost path, aside from ties or whatever, so I think revisits aren't necessary.

gatecat: I don't think that's necessarily guaranteed. Imagine you have two ways of going between two wires:

```
    0.5ns     2.5ns
A -------> X -------> B
     1ns     0.5ns     0.5ns
A -------> Y ------> Z ------>  B
```

`X` will be visited first, and then the higher-cost path to `B`, the lower-cost path via `Y` and `Z` would not be discovered.

Note also the `alt < dist[v]` section in the "using a priority queue" pseudocode [here](https://en.wikipedia.org/wiki/Dijkstra%27s_algorithm).

Ravenslofty: Well, I added revisits, but it turns out to be a relatively minor decrease in path delay compared to not doing them.

PR #1561, opened 2025-09-25
rust: rework ownership model

Ravenslofty:
Yes, this is another leaking API. The alternative is trying to implement hashlib in Rust, and I'm not really sure I feel like that right now.


gatecat:
Couldn't you make some kind of iterator wrapping which would probably only require some small-ish glue functions?

Ravenslofty:
Yes, I can, however I would need to think quite hard about the least-painful way of describing the lifetimes involved, or else `CellInfo` and `NetInfo` can only be pointers on the Rust side of things...

PR #1562, opened 2025-09-25
gatemate: cleanup of PLL and BUFG

mmicko:
This one search for better PLL/BUFG assignment when that is possible.

Currently limiting number to 4 PLLs and BUFGs independently of number of dies, this makes copy of all clocks possible, so they become available in all dies in same order, this makes use of A2 and A4 for lower clock number.

Also it fixes special case where CC_ODDR is having inverter on output, by moving that inverter to both inputs. 

Separate dies are now defined as regions, and cells can be assigned to be in desired one.  force_die option now sets region for all non-pre-placed cells to a given one. We are also making sure that pack procedure preserves region information ( for future use).

### DIFF
--- a/himbaechel/uarch/gatemate/pack_clocking.cc
+++ b/himbaechel/uarch/gatemate/pack_clocking.cc
@@ -64,7 +64,7 @@ void GateMatePacker::sort_bufg()
     }
 
     if (count > (uarch->dies * 4)) {
-        //log_error("More than %d BUFG used. Unable to place them.\n", uarch->dies * 4);
+        // log_error("More than %d BUFG used. Unable to place them.\n", uarch->dies * 4);
         log_warning("More than 4 BUFG used. Those with highest fan-out will be used.\n");
         std::sort(bufg.begin(), bufg.end(), [](const ItemBufG &a, const ItemBufG &b) { return a.fan_out > b.fan_out; });
         for (size_t i = (uarch->dies * 4); i < bufg.size(); i++) {
@@ -119,7 +119,8 @@ void GateMatePacker::pack_bufg()
     unsigned max_plls = 4;
     // Index vector for permutation
     std::vector<unsigned> indexes(max_plls);
-    for (unsigned i = 0; i < max_plls; ++i) indexes[i] = i;
+    for (unsigned i = 0; i < max_plls; ++i)
+        indexes[i] = i;
 
     CellInfo *bufg[4] = {nullptr};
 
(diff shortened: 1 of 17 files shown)


himbaechel/uarch/gatemate/pack_clocking.cc
```diff
@@ -376,42 +413,33 @@ void GateMatePacker::pack_pll()
...
-                clkin[die]->params[ctx->idf("REF%d", pll_index[die])] = Property(pad_info->flags - 1, 3);
-                clkin[die]->params[ctx->idf("REF%d_INV", pll_index[die])] = Property(Property::State::S0);
-                ci.movePortTo(id_CLK_REF, clkin[die], ctx->idf("CLK%d", pad_info->flags - 1));
             } else {
                 // SER_CLK
-                clkin[die]->params[ctx->idf("REF%d", pll_index[die])] = Property(0b100, 3);
-                clkin[die]->params[ctx->idf("REF%d_INV", pll_index[die])] = Property(Property::State::S0);
-                ci.movePortTo(id_CLK_REF, clkin[die], id_SER_CLK);
+                if (clk != net_SER_CLK)
+                    log_error("CLK_REF connected to uknown pin.\n");
```
gatecat: this error message should include some more context - like the name of the clock causing the problem

himbaechel/uarch/gatemate/pack_clocking.cc
```diff
@@ -615,7 +642,186 @@ void GateMatePacker::pack_pll()
...
+            uarch->glbout[new_die]->params = uarch->glbout[0]->params;
+
+            // Copy PLLs
+            for (int i = 0; i < 4; i++) {
+                Loc fixed_loc = uarch->locations[std::make_pair(ctx->idf("PLL%d", i), 0)];
+                BelId pll_bel = ctx->getBelByLocation(fixed_loc);
+                CellInfo *pll = ctx->getBoundBelCell(pll_bel);
+                if (pll) {
+                    // Create new PLL
+                    CellInfo *pll_new = create_cell_ptr(pll->type, ctx->idf("%s$die%d", pll->name.c_str(ctx), new_die));
```
gatecat, line 746: this definitely seems like it could cause some issues if the different PLLs in different dice lock at different times - probably `USR_LOCKED_STDY_RST` (assuming this is the "locked" status output) should be ANDed together between all the dice, and then the output of that AND tree used across all dice. This way all the dice can come out of reset at the same time, if they use the PLL locked output as a reset (as is common).

himbaechel/uarch/gatemate/pack_clocking.cc
mmicko: Have addressed all the PLL related errors/warnings

himbaechel/uarch/gatemate/pack_clocking.cc
mmicko: Agree on this, note that multi die support in this PR is partial and it is more preparation for continuation of work on this, so I have added this as a TODO

PR #1568, opened 2025-10-04
ice40: Initialize context pointer in tests

GyrosGeier:
If the chipdb is not found, the Setup() call throws, but GTest still calls TearDown, which then stumbles over the uninitialized pointer.

This makes the tests fail without valgrind errors or segfaults at least.


gatecat:
Thanks for spotting this!

PR #1569, opened 2025-10-05
placer_heap: fix clamping to region

mmicko:
With current code it was not able to place some of designs, and was constantly trying location out of region box, with this in place it is able to do so.
There could be better way to do it, but this is what came to my mind

### DIFF
--- a/common/place/placer_heap.cc
+++ b/common/place/placer_heap.cc
@@ -909,22 +909,25 @@ class HeAPPlacer
 
                 // Determine a search radius around the solver location (which increases over time) that is clamped to
                 // the region constraint for the cell (if applicable)
-                int rx = radius, ry = radius;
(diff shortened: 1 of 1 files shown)


marzoul:
Very interesting, I wasn't aware nextpnr was supposed to be able to do that.
I found no doc about it, would it be possible to know how to use that, and what target arch is supported ?

gatecat:
>  I found no doc about it, would it be possible to know how to use that, and what target arch is supported ?

It's a generic, but not particularly well exposed (only available via Python), API that in theory supports all target architectures. There's a small example for the iCE40 [here](https://raw.githubusercontent.com/YosysHQ/nextpnr/refs/heads/master/ice40/examples/floorplan/floorplan.py).

PR #1571, opened 2025-10-06
himbaechel/uarch/gowin/cst.cc: added support for IO_LOC with _p/_n separated by a comma

trabucayre:
Gowin tools allows to have in cst file lines like:
```script
IO_LOC "hdmi_data2_p" 39,40;
```
But according to [this comment](https://github.com/YosysHQ/nextpnr/blob/master/himbaechel/uarch/gowin/cst.cc#L86-L88) only the _p is kept, second index is discarded.

But the result is a failure like:
```script
Info: Pack IOBs...
ERROR: Unconstrained IO:hdmi_data2_n_OBUF_O
```

This PR is a PoC/RFC to find a way to allows comma separated constraints.


gatecat:
@yrabbit - does this seem reasonable to you?

yrabbit:
> @yrabbit - does this seem reasonable to you?

It does not break the existing methods of specifying the location and adds an option for diff pins, which is used in Gowin IDE, eliminating one inconvenience when compiling designs intended for the vendor's compiler. So yes, it seems acceptable.

PR #1574, opened 2025-10-10
gatemate: support multiple clock distribution strategies

mmicko:
Added a way to specify strategy of clock placement for multi die gatemate.

TODO
- [x] Add checks if it is possible to use specified strategy
- [x] Add strategy "full" that will not use max number of PLLs available

Misc:
- [x] clangformat
- [x] make sure we are not moving global clock multiple times

### DIFF
--- a/himbaechel/uarch/gatemate/pack.cc
+++ b/himbaechel/uarch/gatemate/pack.cc
@@ -400,11 +400,11 @@ void GateMatePacker::remove_double_constrained()
 void GateMatePacker::recursiveAddToRegion(CellInfo *root, IdString die)
 {
     if (root->region && root->region->name != die)
-        log_error("Trying to assing cell '%s' to multiple regions.\n", root->name.c_str(ctx));
+        log_error("Trying to assign cell '%s' to multiple regions.\n", root->name.c_str(ctx));
     ctx->constrainCellToRegion(root->name, die);
     for (auto cell : root->constr_children) {
         if (cell->region && cell->region->name != die)
-            log_error("Trying to assing cell '%s' to multiple regions.\n", cell->name.c_str(ctx));
+            log_error("Trying to assign cell '%s' to multiple regions.\n", cell->name.c_str(ctx));
         ctx->constrainCellToRegion(cell->name, die);
         recursiveAddToRegion(cell, die);
     }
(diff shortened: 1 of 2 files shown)


himbaechel/uarch/gatemate/pack.cc
```diff
@@ -397,15 +397,72 @@ void GateMatePacker::remove_double_constrained()
...
+
+void GateMatePacker::assign_clocks()
+{
+    log_info("Assign cells based on clock..\n");
+    for (auto &glob : uarch->global_signals) {
+        const NetInfo *net = glob.first;
+        for (auto &user : net->users) {
+            IdString die = uarch->index_to_die[uarch->tile_extra_data(net->driver.cell->bel.tile)->die];
+            if (user.cell->region && user.cell->region->name != die)
+                log_error("Trying to assing cell '%s' to multiple regions.\n", user.cell->name.c_str(ctx));
```
gatecat: nitpick: spelling

himbaechel/uarch/gatemate/pack_clocking.cc
```diff
@@ -682,7 +682,116 @@ void GateMatePacker::copy_clocks()
...
+
+            iosel->connectPorts(id_GPIO_OUT, gpio, id_A);
+            iosel->connectPorts(id_GPIO_EN, gpio, id_T);
+            gpio->connectPort(id_IO, new_clk1);
+        } else
+            gpio->connectPort(id_I, new_clk1);
+
+        NetInfo *new_signal = ctx->createNet(ctx->idf("%s$die%d", net->name.c_str(ctx), new_die));
+        uarch->glbout[new_die]->connectPort(ctx->id("GLB1"), new_signal);
+        copy_constraint(net, new_signal);
```
gatecat, line 787: note to myself as much as anything else - we really need support for clock constraints with phase shifts for timing analysis to work in this kind of situation properly

himbaechel/uarch/gatemate/pack_clocking.cc
```diff
@@ -682,7 +682,116 @@ void GateMatePacker::copy_clocks()
...
+
+        if (new_die == 0) {
+            iosel->connectPort(id_OUT1, ctx->getNetByAlias(uarch->global_signals.begin()->first->name));
+            CellInfo *cpe = move_ram_o_fixed(iosel, id_OUT1, loc).first;
+            uarch->ignore.emplace(cpe->name);
+
+            iosel->connectPorts(id_GPIO_OUT, gpio, id_A);
+            iosel->connectPorts(id_GPIO_EN, gpio, id_T);
+            gpio->connectPort(id_IO, new_clk1);
+        } else
```
gatecat: can you add braces to the `else` clause here too? looks weird without them

himbaechel/uarch/gatemate/pack.cc
mmicko: fixed

himbaechel/uarch/gatemate/pack_clocking.cc
mmicko: fixed

gatecat:
Merged, appreciate it's not so important while multi-die parts are not generally available, but as a follow-up it would be good to document somewhere what these different strategies actually mean.

mmicko:
Thanks. Think best place to document that would be in code so will add that.

Meanwhile I have pushed clangformat changes since those were missing in xo2 PR.

PR #1583, opened 2025-10-17
gatemate: additional region handling

mmicko:
This PR adds
- Assigning to cells to specific region from netlist (GATEMATE_DIE attribute)
- Assigning IO to specific die with DIE parameter if CCF file
- Assigning whole cluster to region (previously just executed for clock distribution)
- When merging cells take region for cells having them defined instead of ignoring information
- Adds penalty to estimated delay when crossing between dies

### DIFF
--- a/himbaechel/uarch/gatemate/delay.cc
+++ b/himbaechel/uarch/gatemate/delay.cc
@@ -39,6 +39,16 @@ delay_t GateMateImpl::estimateDelay(WireId src, WireId dst) const
     return 100 + 100 * (std::abs(dx - sx) + std::abs(dy - sy)) + d2d;
 }
 
+delay_t GateMateImpl::predictDelay(BelId src_bel, IdString src_pin, BelId dst_bel, IdString dst_pin) const
+{
+    int d2d = 0;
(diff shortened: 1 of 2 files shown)


himbaechel/uarch/gatemate/delay.cc
```diff
@@ -30,10 +30,13 @@ NEXTPNR_NAMESPACE_BEGIN
...
 {
     int sx, sy, dx, dy;
+    int d2d = 0;
+    if (tile_extra_data(src.tile)->die != tile_extra_data(dst.tile)->die)
+        d2d += 2000;
     tile_xy(ctx->chip_info, src.tile, sx, sy);
     tile_xy(ctx->chip_info, dst.tile, dx, dy);
 
-    return 100 + 100 * (std::abs(dx - sx) + std::abs(dy - sy));
+    return 100 + 100 * (std::abs(dx - sx) + std::abs(dy - sy)) + d2d;
```
gatecat, line 39: probably worth adding this extra delay to `predictDelay` too so the placer will always take into account as well

mmicko: actually had this one pending, but was not sure since default calculation does not add this constant 100 delay, but adding it seems not to affect it in any problematic way

PR #1589, opened 2025-10-21
Gate Boost `system` component behind version check

### LINKED ISSUE #1588: Build broken on boost 1.89.0 due to removal of `libboost_system` library stubs
As of recently, boost removed the library stubs for `libboost_system` which means the cmake component lookup for it will fail causing the build to error out, even through the library is still present.

See this thread: https://github.com/boostorg/system/issues/132

The fix would be to check the boost version and then only add the `system` component if it's below `1.89.0`, otherwise to omit it from the lookups

This also applies to prjtrellis: https://github.com/YosysHQ/prjtrellis/issues/251

lethalbit:
This PR adds a version check for the Boost `system` component to only require the stub libraries if the version of boost is older than `1.89.0`, as it was removed in that version

This closes #1588

See Also: https://github.com/YosysHQ/prjtrellis/pull/258

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -169,11 +169,18 @@ else()
     endif()
 endif()
 
-set(boost_libs filesystem program_options iostreams system)
+set(boost_libs filesystem program_options iostreams)
 if (Threads_FOUND)
     list(APPEND boost_libs thread)
 endif()
 find_package(Boost REQUIRED COMPONENTS ${boost_libs})
+
+# The `boost_system` library stubs were removed in 1.89.0, so only add them if
+# boost is older than that
+if (Boost_VERSION VERSION_LESS "1.89.0")
+    list(APPEND boost_libs system)
+    find_package(Boost REQUIRED COMPONENTS ${boost_libs})
+endif()
 
 if (BUILD_PYTHON)
     # TODO: sensible minimum Python version
--- a/bba/CMakeLists.txt
+++ b/bba/CMakeLists.txt
@@ -3,13 +3,24 @@ project(bba CXX)
 
 set(CMAKE_CXX_STANDARD 17)
 
-find_package(Boost REQUIRED COMPONENTS
-    program_options
-    system)
+set(boost_libs program_options iostreams)
+
+find_package(Boost REQUIRED COMPONENTS ${boost_libs})
+
+if (Boost_VERSION VERSION_LESS "1.89.0")
+    list(APPEND boost_libs system)
+    find_package(Boost REQUIRED COMPONENTS ${boost_libs})
+endif()
+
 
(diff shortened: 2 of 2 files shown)


lethalbit:
I have no idea why this broke the build on Ubuntu, i'll figure it out in a bit, going to try to also get a PR for prjtrellis to fix the same issue together as well in parallel

whitequark:
What is nextpnr actually using from boost_system?

lethalbit:
I have no idea, and honestly if C++11 is the lowest version supported then we don't need it at all.

So i'm not against just outright removing it if wanted. But i'm just fixing the build issue as it stands.

whitequark:
I'd strongly prefer to get rid of it entirely rather than adding complex build system logic for something we seemingly don't even need. Same goes for prjtrellis.

lethalbit:
I don't disagree, but to be fair it's not *that* much more complex, but if C++17 is the min target and we're not taking advantage of the other things that the Boost `system` library provides then I would also vote for cutting it out.

I feel that @mmicko or @gatecat should weigh in prior to that decision being made though.

whitequark:
While the logic itself may not look complex, it does add significant future maintainer workload: when making changes to the boost-related bits, one would have to stop, consider whether they need to try rebuilding the project with 1.88.0, then potentially spend time on that build and looking at the changelogs. (Essentially what I just had to do when reviewing this PR--and I still don't really know where it's _used_.) This is what I want to spare future maintainers from.

lethalbit:
Again, I don't disagree, and honestly I would be in favor of gutting Boost altogether if not just the `system` dep.

But, again, I think one of the other code owners of the project(s) should weigh in prior to that decision being made, and if the decision is that it's fine to just cut it out then we can do that.

I rather not jump the gun here, because maybe there *is* a good reason to use the Boost `system` headers over the C++ STL ones, as it does have some added things such as a `result<T>` and other general improvements over the standard ones.

If we're not using any of those, then yeah, toss it out

gatecat:
From what I can tell nothing in nextpnr is using `boost::system` unless it was somehow required for `boost::filesystem`. In any case, as nextpnr is now using C++17, I've removed both in https://github.com/YosysHQ/nextpnr/pull/1591

gatecat:
As far as removing boost altogether goes, the biggest remaining issue would be `program_options` as that's quite heavily baked into nextpnr at the moment. Any thoughts on this would be welcome, I'll have significantly more time to work on nextpnr-related things in the near future.

mmicko:
@gatecat for `program_options` replacement, quite simple, header only and doing what we need is https://github.com/jarro2783/cxxopts, that is btw added time ago to be used for yosys as well

PR #1591, opened 2025-10-22
Remove use of boost system and filesystem

gatecat:


mmicko:
Note `bba/CMakeLists.txt` is also using it as Boost_SYSTEM_LIBRARY variable

gatecat:
Thanks, good spot, fixed!

mmicko (approved):

PR #1594, opened 2025-10-24
Gowin. Second part of SDP division.

yrabbit:
SemiDual Port BSRAM in the GW5A series cannot function with a data width of 32/36 bits. This is not specified in the documentation; it was discovered during research into bit streams, where one SDP is split into two with a lower bit width.

The situation is complicated by the fact that the bit width is set separately for port A and port B. Nextpnr already knows how to correct the simplest case, when both ports have a width of 32/36 bits.

Here we take the next step: we correct the situation when port A has a bit width of 32/36 bits and port B has an arbitrary bit width.

With a narrow port B, all address lines and control signals are shared between the old and new cells without change. The old and new cells operate in parallel and simultaneously output two versions of port B data.

However, the DOx outputs of the cells are connected to a MUX, which selects which cell's data will be the official output.  The MUX is controlled by an address decoder for the ADB4 line - since port A writes half of the data to different cells in 16/18 bits, it is the ADB4 bit that selects the correct cell for reading.

The MUX responds too quickly, so we add one delay register, an additional output register if necessary, and BLKSELB signal decoding.


whitequark (commented):
This transformation doesn't seem like it should be in a P&R tool.

yrabbit:
Why? This is a correction of the implementation features of the primitive in one specific type of chip, not observed in other series, and perhaps in more powerful models of the same series.

Nextpnr knows the specific chip down to the pinout, yosys does not need such parameters. Yes, we already specify the series there, but we are still far from specifying the PartNO for synthesis :)

gatecat:
Yeah, I agree it would probably be nicer to implement this in Yosys (if I understand right, just disabling the type of RAM that doesn't work and letting memory_libmap do the rest?) than having a bunch of code generating logic in nextpnr, that always ends up quite messy.

PR #1603, opened 2025-11-21
gatemate: pack multiplier output registers

Ravenslofty:
Given the potentially very long propagation delay of the gatemate multipliers, we attempt to pack registers directly next to the multiplier P output to minimise any further routing delay.

### DIFF
--- a/himbaechel/uarch/gatemate/pack_mult.cc
+++ b/himbaechel/uarch/gatemate/pack_mult.cc
@@ -927,9 +927,9 @@ void GateMatePacker::pack_mult()
                     auto *p_passthru =
                             create_cell_ptr(id_CPE_L2T4, ctx->idf("%s$p[%d]_passthru", cpe_half->name.c_str(ctx), p));
 
-                    p_passthru->params[id_INIT_L00] = Property(LUT_D0, 4);   // IN5
-                    p_passthru->params[id_INIT_L01] = Property(LUT_ZERO, 4); // (unused)
-                    p_passthru->params[id_INIT_L10] = Property(LUT_D0, 4);   // L02
+                    p_passthru->params[id_INIT_L00] = Property(LUT_D0, 4);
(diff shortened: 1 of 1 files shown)


himbaechel/uarch/gatemate/pack_mult.cc
```diff
@@ -916,6 +916,43 @@ void GateMatePacker::pack_mult()
...
+            if (cpe_half_cpout && cpe_half_cpout->users.entries() == 1) {
+                auto cpe_half_cpout_user = *cpe_half_cpout->users.begin();
+                auto *dff = cpe_half_cpout_user.cell;
+                NPNR_ASSERT(dff != nullptr);
+                if (dff->type == id_CC_DFF) {
+                    // Instantiate a P passthrough L2T4 for the flop.
+                    auto *p_passthru =
+                            create_cell_ptr(id_CPE_L2T4, ctx->idf("%s$p[%d]_passthru", cpe_half->name.c_str(ctx), p));
+
+                    p_passthru->params[id_INIT_L00] = Property(LUT_D0, 4);   // IN5
```
mmicko: maybe better get rid of these comments in these 3 lines (for id_INIT_L* params), since those are not correct for all cases

rowanG077:
Hmmm shouldn't this be a yosys optimization?

Ravenslofty:
> Hmmm shouldn't this be a yosys optimization?

No, because the GateMate multipliers work very differently to how other architectures implement multiplication:

msrdc_eo7BeiQHhR

Highlighted in red are the output registers that this PR results in - the rhombus shape is the multiplier, made up of 2x2 signed multiplier cells. It's not like a more conventional FPGA architecture where register packing amounts to setting a bitstream configuration, we need to physically constrain the registers ourselves.

rowanG077:
I see, yes I would have expected it would be a parameter on a primitive that you flip to 1 or "TRUE" or whatever. If that's not the case and it amounts to custom placement then I agree this should be in nextpnr.

PR #1607, opened 2025-11-27
gatemate: floorplanning [sc-168]

mmicko:
- [x] add CCF file parsing compatible with test p_r
- [x] add basic constraining to specified region

Improvements
- [x] add wildcard support for sub scopes
- [x] verbose info about placement
- [x] clean CCF format to include only valuable information (EDIT: we can ignore unused info, but support full format from p_r)


ShinyKate:
[sc-168]

PR #1608, opened 2025-12-01
gatemate: fix output register packing

Ravenslofty:
As it turns out, I was only packing the "diagonal" section of the P output registers, not the "vertical" section. This fixes that.

While I was here, I also added logic to hopefully avoid packing P output registers when those registers have mismatched control sets.


mmicko:
@Ravenslofty Sorry it took a while

PR #1611, opened 2025-12-02
ice40: Explicitly connect CIN when legalizing carry chain

### LINKED ISSUE #1605: ice40: Maybe a connection is missing in carry_passout in the generated netlist
Hi,

We have sometimes issue simulating ice40 post-pr netlists generated by nextpnr (while correctly behaving on the device).

We have tried to track the issue and I think it is related to the way carry cells and lut are packed in logic cells when both the carry  and lut output are used.

More precisely, we found that adding a connection between the I3 input and the CIN input here:

https://github.com/YosysHQ/nextpnr/blob/cfa5f77dd992909a9af2fc59493c1c340f4efdae/ice40/chains.cc#L126

solves the issues for me.

Something like:

```diff
         // If COUT also connects to a CIN; preserve the carry chain
         if (cin_cell) {
+            lc->connectPort(id_CIN, co_i3_net);
             NetInfo *co_cin_net = ctx->createNet(ctx->id(lc->name.str(ctx) + "$COUT"));
 
             // Connect I1 to 1 to preserve carry chain
```

Maybe someone with better knowledge of ICE40 arch and nextpnr internals can take a  look at this/give us some hints.

tarikgraba:
This connection is implicit as it is hardwired in the hardware. This commit makes the connection explicit and thus appearing in the generated netlist allowing post-rout simulation.

Resolves #1605

### DIFF
--- a/ice40/chains.cc
+++ b/ice40/chains.cc
@@ -122,6 +122,8 @@ class ChainConstrainer
         o_r.cell = lc.get();
         cout_port.net->driver = o_r;
         cout_port.net = co_i3_net;
+        // CIN is connected to COUT in hardware, make the connection explicit in the netlist
+        lc->connectPort(id_CIN, co_i3_net);
 
         // If COUT also connects to a CIN; preserve the carry chain
         if (cin_cell) {


ice40/chains.cc
```diff
@@ -122,6 +122,10 @@ class ChainConstrainer
         o_r.cell = lc.get();
         cout_port.net->driver = o_r;
         cout_port.net = co_i3_net;
+        // Also connect the created net to the CIN input
```
wanda-phi: I don't think we need three lines of comment for a line that should have always been here in the first place.

However, if you do write a comment, at least please keep the proper indentation.

tarikgraba: My bad.

Commit changed.

ice40/chains.cc
```diff
@@ -122,6 +122,8 @@ class ChainConstrainer
         o_r.cell = lc.get();
         cout_port.net->driver = o_r;
         cout_port.net = co_i3_net;
+        // CIN is connected to COUT in hardware, explicit the connection in the netlist
```
wanda-phi:
```suggestion
        // CIN is connected to COUT in hardware, make the connection explicit in the netlist
```

PR #1612, opened 2025-12-09
gatemate: input register packing

Ravenslofty:
Still to do:
- it seems like Yosys sometimes merges input registers together, resulting in these input registers having multiple fanout.
  - this normally inhibits register packing (e.g. output register packing is inhibited in this case); is this desirable?
  - or do we duplicate the register in question so we can pack it while preserving placement quality for other users?
- odd A input register packing
- B input register packing
- do we want to merge L2T4s into the otherwise-passthrough LUT? what do we do about "synthesis bad luck" where the input logic is a L2T5? (this also applies for output register packing, but is perhaps a bit less critical)

### DIFF
--- a/himbaechel/uarch/gatemate/pack_mult.cc
+++ b/himbaechel/uarch/gatemate/pack_mult.cc
@@ -46,12 +46,14 @@ struct ZeroDriver
 // Propagate A0 through OUT1 and A1 through OUT2; zero COUTX and POUTX.
 struct APassThroughCell
 {
-    APassThroughCell(CellInfo *lower, CellInfo *upper, CellInfo *comp, CellInfo *cplines, IdString name);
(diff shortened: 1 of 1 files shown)


Lunaphied:
Is this still being worked on?

Ravenslofty:
Since you're so determined to get it closed, there you go.

Lunaphied:
Not determined to get it closed just trying to help with the backlog ^^;

Ravenslofty:
Having someone close one PR as stale and claim another is stale and worth closing without saying a word to me sure comes off as "determined to get it closed".

Lunaphied:
Honestly sorry about that. I was working on a triage script and I had a live run when I meant for a dry run. These should've just been questions about current status

Issue #1613, opened 2025-12-11
`nextpnr-himbaechel` for `gatemate` devices fails when built on top of the `IIC-OSIC-TOOLS` container


thommythomaso:
In our [VLSI 2 Course at ETH](vlsi.ethz.ch), we use JKU's [IIC-OSIC-TOOLS](https://github.com/iic-jku/IIC-OSIC-TOOLS) v2025.11 container to provide the required open-source EDA tools. The container currently does not include nextpnr, so we are trying to build the nextpnr toolchain on top of the upstream OSIC container.

We can successfully build nextpnr using the following commands inspired by the [OSS CAD Tools](https://github.com/YosysHQ/oss-cad-suite-build):

```bash
# work dir
mkdir -p /foss/nextpnr-build
cd /foss/nextpnr-build

# get dependencies
apt update
apt install -y libboost-all-dev qtbase5-dev qtchooser qt5-qmake qtbase5-dev-tools

# get and prepare nextpnr
cd /foss/nextpnr-build
git clone https://github.com/YosysHQ/nextpnr.git
cd nextpnr
git checkout c46f2bbd86e73f9254051dabc573380a7dc157fd
git submodule update --init --recursive

# get and install peppercorn
cd /foss/nextpnr-build
git clone https://github.com/YosysHQ/prjpeppercorn.git
cd prjpeppercorn
git checkout 5a43fa5fc0a7316170103c2a7186ef6ff9be5dc9
cd libgm
sed -i 's,cmake_policy(SET CMP0167 NEW),,g' CMakeLists.txt
cmake -DCMAKE_INSTALL_PREFIX=/foss/tools/nextpnr .
make -j$(nproc) install

# build and export nextpnr-gatemate-bba
cd /foss/nextpnr-build/nextpnr
rm -rf build && mkdir -p build && cd build
cmake .. \
    -DARCH="himbaechel" \
    -DBUILD_GUI=ON \
    -DUSE_IPO=OFF \
    -DHIMBAECHEL_UARCH="gatemate" \
    -DHIMBAECHEL_GATEMATE_DEVICES=all \
    -DHIMBAECHEL_PEPPERCORN_PATH=/foss/nextpnr-build/prjpeppercorn \
    -DEXPORT_BBA_FILES=/foss/tools/nextpnr/bba-files
make -j$(nproc) nextpnr-all-bba

# build and install nextpnr-himbaechel
cd /foss/nextpnr-build/nextpnr
[...]

PR #1614, opened 2025-12-12
himbaechel: fix parsing vopt memory issue

mmicko:
This fixes https://github.com/YosysHQ/nextpnr/issues/1613


gatecat:
Thanks for fixing this!

PR #1679, opened 2026-03-27
qt: remove qt5 as it's no longer relevant and clean up some warnings

Lunaphied:
The latest Ubuntu LTS (24.05) has Qt6 and by removing Qt5 compatibility we can clean up some warnings that appear when building under it.

This also cleans up a tiny warning from a no longer required eigen3 init method.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -227,14 +227,7 @@ else()
 endif()
 
 if (BUILD_GUI)
-    # Find the Qt6/5 libraries
     find_package(Qt6 COMPONENTS Core Widgets OpenGL OpenGLWidgets QUIET)
-    if (Qt6_FOUND)
-        message(STATUS "Using Qt6")
-    else()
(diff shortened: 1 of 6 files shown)


gatecat:
@mmicko at a glance looks like oss-cad-suite is still using Qt5. Will there be any problems upgrading this to Qt6?

mmicko:
@gatecat for oss-cad-suite it would be problem for now, plan is to move to Qt6, but it would take me some time, since I have to build libraries for macOS manually due to way how things are working for cross compiling.

GyrosGeier:
I'm using it for backported packages to Debian oldstable, and for m68k builds, I think all of these can be dropped.

For everything else, Qt6 appears to be a drop-in replacement.

mmicko:
@gatecat we can merge this if you are ok with it, oss-cad-suite now builds with  qt6 anyway

PR #1697, opened 2026-04-09
gatemate: removing recursion in GateMateImpl::reassign_bridges

sylefeb:
This PR changes the algorithm structure of `GateMateImpl::reassign_bridges` in `gatemate.cc` to remove the recursive calls. This resolves a stack overflow crash when using a wasi+wasm build (yowasp, pypi) on all but smallest designs. This modified version is equivalent in functionality.

Changes:
- implementation in `gatemate.cc` now uses a vector to record next items to be processed, recursion is implemented by looping on the vector while non-empty, recursive calls are replaced by a push of a record on the vector,
- method signature in `gatemate.h` adjusted for parameter name change.

### DIFF
--- a/himbaechel/uarch/gatemate/gatemate.cc
+++ b/himbaechel/uarch/gatemate/gatemate.cc
@@ -425,21 +425,23 @@ void GateMateImpl::preRoute()
     }
 }
 
-void GateMateImpl::reassign_bridges(NetInfo *start_nfo, const dict<WireId, PipMap> &net_wires,
+void GateMateImpl::reassign_bridges(NetInfo *start_net, const dict<WireId, PipMap> &net_wires,
                                     WireId start_wire, dict<WireId, IdString> &wire_to_net, int &num)
 {
     // Processing list, holds parameters to implement the equivalent of recursive calls.
-    struct record { NetInfo *nfo; WireId wire; };
+    // This avoids a stack overflow when recursion becomes deep, as the function
+    // has a relatively large stack footprint.
+    struct record { NetInfo *net; WireId wire; };
     std::vector<record> to_process;
     // Insert start record.
-    to_process.push_back({start_nfo, start_wire});
+    to_process.push_back({start_net, start_wire});
     // For as long as there are pending records, process them.
     while (!to_process.empty()) {
-	// Get the next record to process.
+        // Get the next record to process.
         record cur = to_process.back();
         to_process.pop_back();
 
-        wire_to_net.insert({cur.wire, cur.nfo->name});
+        wire_to_net.insert({cur.wire, cur.net->name});
 
         for (auto pip : ctx->getPipsDownhill(cur.wire)) {
             auto dst = ctx->getPipDstWire(pip);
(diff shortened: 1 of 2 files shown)


himbaechel/uarch/gatemate/gatemate.cc
```diff
@@ -425,62 +425,75 @@ void GateMateImpl::preRoute()
     }
 }
 
-void GateMateImpl::reassign_bridges(NetInfo *ni, const dict<WireId, PipMap> &net_wires, WireId wire,
-                                    dict<WireId, IdString> &wire_to_net, int &num)
+void GateMateImpl::reassign_bridges(NetInfo *start_nfo, const dict<WireId, PipMap> &net_wires,
```
gatecat: it's a nitpick but I prefer start_net to start_nfo, more consistent with other parts of nextpnr

himbaechel/uarch/gatemate/gatemate.cc
```diff
@@ -425,62 +425,75 @@ void GateMateImpl::preRoute()
...
+                                    WireId start_wire, dict<WireId, IdString> &wire_to_net, int &num)
 {
-    wire_to_net.insert({wire, ni->name});
-
-    for (auto pip : ctx->getPipsDownhill(wire)) {
-        auto dst = ctx->getPipDstWire(pip);
+    // Processing list, holds parameters to implement the equivalent of recursive calls.
```
gatecat, line 431: can you also add a comment explaining that recursion here was causing stack overflow issues and that's why we're avoiding it with this?

sylefeb: Thanks for the feedback, good points => fixing

gatecat:
This looks good to me know. any comments from @Ravenslofty / @mmicko as gatemate owners before this gets merged?

Ravenslofty:
I did take a look; I remember this being kind of a pain to ensure it was right, which is why prjpeppercorn-test-cases has `make check-pnr` to try to prove equivalence. @sylefeb did you try this on some designs?

sylefeb:
I did run it on a few VGA designs without problems (three from Silice, relatively complete in terms of features) but I certainly don't want to break anything! I'll try to strictly compare the bitstreams binary (fixing seeds) and report back. Are there specific tests that could be done beyond that? (=> will look at prjpeppercorn-test-cases check-pnr)

mmicko:
@sylefeb To check all, best is to clone https://github.com/YosysHQ/prjpeppercorn-test-cases and then run `./check_pnr.py` in root of it, that will confirm that output netlist is equivalent with input one. And also you can go into `verification` sub directory and run `./check_pnr.py` there as well.
Note that both of these will check just a bit smaller designs, so for complex I would recommend trying them on hardware at the end as you did for some VGA examples.

sylefeb:
I tested both `./check_pnr.py` and `verification/check_pny.pr`, all tests succeeded. I successfully tried several of the VGA demos in `prjpeppercorn-test-cases`, and also compared bitstream before/after change, all those I tested were identical.

PR #1700, opened 2026-04-14
Improve file open error messages

mrcmry:
This PR follows my proposition in #1682.

It improve all file opening error checks:

- add when missing
- quote 'filename'
- add error cause to easier troubleshoot
- use existing consistent string style

When missing, I added it only where it seems there was use of log_error in this context.

Note that checking `if (!file.is_open())` is the same as `if (!file)` just after opening the file and make more sense when using `std::generic_category()`.

I formatted the modified files with ``clang-format -style=file -i `git ls-files --modified` ``

I tested with a gatemate build and different cases (file as root, file missing, file is a dir, ...) and ubuntu 24.04. It should works fine on other plateforms and build.

### DIFF
--- a/nexus/main.cc
+++ b/nexus/main.cc
@@ -62,9 +62,7 @@ void NexusCommandHandler::customBitstream(Context *ctx)
 {
     if (vm.count("fasm")) {
         std::string filename = vm["fasm"].as<std::string>();
-        std::ofstream out(filename);
-        if (!out)
-            log_error("Failed to open output FASM file %s.\n", filename.c_str());
+        auto out = open_ofstream_and_log_error(filename, "output FASM file");
         ctx->write_fasm(out);
     }
 }
@@ -101,9 +99,7 @@ void NexusCommandHandler::customAfterLoad(Context *ctx)
 {
     if (vm.count("pdc")) {
         std::string filename = vm["pdc"].as<std::string>();
-        std::ifstream in(filename);
-        if (!in)
-            log_error("Failed to open input PDC file %s.\n", filename.c_str());
+        auto in = open_ifstream_and_log_error(filename, "input PDC file");
         ctx->read_pdc(in);
     }
 }
--- a/common/kernel/command.cc
+++ b/common/kernel/command.cc
@@ -609,13 +609,13 @@ int CommandHandler::executeMain(std::unique_ptr<Context> ctx)
         try {
             if (vm.count("json")) {
                 std::string filename = vm["json"].as<std::string>();
-                std::ifstream f(filename);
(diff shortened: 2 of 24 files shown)


gatecat:
Do you know if this will cause any problems on the Windows build? I'm not 100% sure if errno will be available there?

If special handling for Windows is needed we should probably wrap this all into some kind of helper function.

mrcmry:
Good point. 

`errno` is [available since at least Visual Studio 2015](https://learn.microsoft.com/en-us/cpp/standard-library/cerrno?view=msvc-140). `error_code` [as well](https://learn.microsoft.com/en-us/cpp/standard-library/error-code-class?view=msvc-140#error_code). So I think it's portable on windows.

Now there is the problem of the imports. I was thinking to write functions like this:
```C++
std::ifstream open_ifstream_and_log_error(std::string filename, const char* file_description)
{
    std::ifstream file(filename);
    if (!file.is_open()) {
        log_error("Failed to open %s '%s': %s.\n", filename.c_str(),
                    std::error_code(errno, std::generic_category()).message().c_str());
    }

    return std::move(file);
}
```

and the same for `ostream` with "for writting" in the string and add it to `log.h` and `log.cc`. What do you think ? It's probably more readable and also easier to fix problems with Windows if any. I think `_and_log` is better than `or_log` because even with `log_error` always rising an exception conceptually without exception it always open and return.

gatecat:
That approach makes sense to me!

mrcmry:
So I refactored the changes to use 2 functions in log.h. It is also rebased on the latest main.

nexus/main.cc
```diff
@@ -30,7 +30,7 @@ class NexusCommandHandler : public CommandHandler
 {
   public:
     NexusCommandHandler(int argc, char **argv);
-    virtual ~NexusCommandHandler() {};
+    virtual ~NexusCommandHandler(){};
```
gatecat, line 33: there are a few format changes like this that have crept in maybe because the clang-format version being used is different. would be great if you could remove them from this patch set

mrcmry: Of course, no problem. Which version of clang-format do you use ? My Ubuntu 24.04 has apt package for clang-format from version 14 to 20. The default is 18.

I had problem with clang-format version differences in the past in other projects. Perhaps it would be good to precise which version to use in the README.md.

gatecat: Apparently I'm on 22.1.1

mrcmry:
I ran "clang-format-20 -style=file -i `git diff --name-only HEAD~1 HEAD`" and rebased it to a single commit. Now there is only changes to the fstream part of the code.

gatecat:
Great, thanks for sorting this!

PR #1715, opened 2026-05-09
WASI: fix thread presence detection

whitequark:
I've also fixed a few outdated checks for the WASI system, as a part of a shift towards using upstream toolchain files from WASI SDK. I will also send follow-up PRs to all the individual projects with the same fix.


gatecat (approved):

whitequark:
(Waiting for confirmation that this works in YoWASP, then I'll merge)

PR #1722, opened 2026-05-12
fix visual studio build

moffel:
These changes were necessary to make nextpnr-himbaechel build with gui on windows using visual studio 2022/2026.

The only general change is the addition of proc_path + "../share" to the list of share search locations. This is the natural path where the share folder is located after build. Allowing visual studio builds to "just run" out of the box.

Note that windows does not require backslashes as folder separators. Therefore, I removed the win32 specific version.

### DIFF
--- a/3rdparty/QtPropertyBrowser/src/CMakeLists.txt
+++ b/3rdparty/QtPropertyBrowser/src/CMakeLists.txt
@@ -39,7 +39,9 @@ add_library(${TARGET_NAME} STATIC
   ${_QRC_SRCS}
   )
 
-target_compile_options(${TARGET_NAME} PRIVATE -Wno-deprecated-declarations)
+if (!MSVC)
+  target_compile_options(${TARGET_NAME} PRIVATE -Wno-deprecated-declarations)
+endif()
 
 if (MSVC)
   target_compile_options(${TARGET_NAME} PRIVATE /wd4457 /wd4718)
--- a/3rdparty/QtPropertyBrowser/src/CMakeLists.txt
+++ b/3rdparty/QtPropertyBrowser/src/CMakeLists.txt
@@ -39,7 +39,7 @@ add_library(${TARGET_NAME} STATIC
   ${_QRC_SRCS}
   )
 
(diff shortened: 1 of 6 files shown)


3rdparty/QtPropertyBrowser/src/CMakeLists.txt
```diff
@@ -39,7 +39,9 @@ add_library(${TARGET_NAME} STATIC
   ${_QRC_SRCS}
   )
 
-target_compile_options(${TARGET_NAME} PRIVATE -Wno-deprecated-declarations)
+if (!MSVC)
```
gatecat: looking at the CI failure I think this needs to be `if (NOT MSVC)`

common/kernel/command.cc
```diff
@@ -191,36 +191,21 @@ void init_share_dirname() { npnr_share_dirname = "/share/"; }
 void init_share_dirname()
 {
     std::string proc_self_path = proc_self_dirname();
-#if defined(_WIN32) && !defined(nextpnr_WIN32_UNIX_DIR)
```
gatecat, line 194: are you certain  that removing this is okay in all cases?

moffel: Yes. Is there any case in particular you worry about? 

Anyway, the main change is the addition of one more default search path. The rest is just unnecessary refactoring I could as well revert.

3rdparty/QtPropertyBrowser/src/CMakeLists.txt
moffel: done

common/kernel/command.cc
gatecat: would be nice if @mmicko can confirm that this won't break the oss-cad-suite build

mmicko: Creating local build right now with oss-cad-suite tools to try this

mmicko: @gatecat @moffel can confirm that it is working fine with oss-cad-suite made builds

PR #1723, opened 2026-05-12
merge L2T5 and upper L2T4 int a single CPE if possible

moffel:
This implements the merge of CC_L2T5 and CC_L2T4 macros into a single GateMate CPE as described in [UG1001](https://www.colognechip.com/docs/ug1001-gatemate1-primitives-library-latest.pdf) 4.3.

### DIFF
--- a/himbaechel/uarch/gatemate/pack_cpe.cc
+++ b/himbaechel/uarch/gatemate/pack_cpe.cc
@@ -181,7 +181,7 @@ void GateMatePacker::override_region(CellInfo *first, CellInfo *second)
 void GateMatePacker::pack_cpe()
 {
     log_info("Packing CPEs..\n");
-    std::vector<CellInfo *> l2t5_list;
+    pool<IdString> l2t5_list;
 
     auto merge_dff = [&](CellInfo &ci, CellInfo *dff) {
         dff->cluster = ci.name;
@@ -203,7 +203,7 @@ void GateMatePacker::pack_cpe()
             continue;
         bool is_l2t5 = false;
         if (ci.type == id_CC_L2T5) {
-            l2t5_list.push_back(&ci);
+            l2t5_list.insert(ci.name);
             ci.renamePort(id_I0, id_D0_00);
             ci.renamePort(id_I1, id_D1_00);
             ci.renamePort(id_I2, id_D0_01);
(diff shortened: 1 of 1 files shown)


himbaechel/uarch/gatemate/pack_cpe.cc
```diff
@@ -294,20 +294,55 @@ void GateMatePacker::pack_cpe()
             }
         }
     }
+
+    std::sort(l2t5_list.begin(), l2t5_list.end());
```
gatecat: sorting a list of pointers is going to introduce non-determinism

gatecat: looks like this is being sorted in order to use `lower_bound`. a dict or pool of IdStrings would be the idiomatic nextpnr solution here

himbaechel/uarch/gatemate/pack_cpe.cc
```diff
@@ -294,20 +294,55 @@ void GateMatePacker::pack_cpe()
...
+            upperNet->driver.cell->cluster.empty()) {
+            auto it = std::lower_bound(l2t5_list.begin(), l2t5_list.end(), upperNet->driver.cell);
+            merge = it == l2t5_list.end() || *it != upperNet->driver.cell;
+        }
+
+        if (merge) {
+            CellInfo *upper = upperNet->driver.cell;
+
+            // rename cell
+            IdString newName = ctx->idf("%s$upper", ci->name.c_str(ctx));
```
gatecat: is there any actual need to rename the cell?

himbaechel/uarch/gatemate/pack_cpe.cc
moffel: Good catch. I changed this to use dict of IdString instead.

himbaechel/uarch/gatemate/pack_cpe.cc
moffel: Indeed not. I thought it was necessary because I had issues with this change at first. But it turns out not to be required with the proper cell filter. I also removed the net renaming.

himbaechel/uarch/gatemate/pack_cpe.cc
```diff
@@ -181,7 +181,7 @@ void GateMatePacker::override_region(CellInfo *first, CellInfo *second)
 void GateMatePacker::pack_cpe()
 {
     log_info("Packing CPEs..\n");
-    std::vector<CellInfo *> l2t5_list;
+    dict<IdString, CellInfo*> l2t5_list;
```
gatecat: the CellInfo* here is redundant. you can use a pool of IdString and then do `ctx->cells.at(name).get()` to get the CellInfo pointer

himbaechel/uarch/gatemate/pack_cpe.cc
```diff
@@ -294,20 +294,42 @@ void GateMatePacker::pack_cpe()
...
+
+    for (auto& it : l2t5_list) {
+        CellInfo *ci = it.second;
+        NetInfo *upperNet = ci->getPort(id_I4);
+
+        if (upperNet && 
+            upperNet->driver.cell && 
+            upperNet->driver.cell->type == id_CPE_L2T4 &&
+            upperNet->driver.cell->cluster.empty() &&
+            l2t5_list.find(upperNet->driver.cell->name) == l2t5_list.end()) {
```
gatecat: neater would probably just be `!l2t5_list.count(upperNet->driver.cell->name)`

himbaechel/uarch/gatemate/pack_cpe.cc
```diff
@@ -294,20 +294,42 @@ void GateMatePacker::pack_cpe()
...
+            upperNet->driver.cell && 
+            upperNet->driver.cell->type == id_CPE_L2T4 &&
+            upperNet->driver.cell->cluster.empty() &&
+            !l2t5_list.count(upperNet->driver.cell->name)) {
+
+            CellInfo *upper = upperNet->driver.cell;
+
+            // cluster into single CPE
+            upper->cluster = ci->name;
+            upper->region = ci->region;
```
moffel, line 312: Can you please elaborate on what this region pointer is for? I feel like the merge should only be done if both cells are within the same region? But I'm not sure what the region pointer is used for.

gatecat: Yes, this is for constraining cells to specific placement regions (like on the multi die gatemate parts). This isn't always 100% followed in the packers even if it should be. So if the rest of the code is merging cells with different regions you probably can get away without doing different here... Otherwise yes you should be checking they match before clustering.

PR #1724, opened 2026-05-12
gatemate: Fix GPIO output-enable inversion when CC_LUT1 drives T input

### LINKED ISSUE #1713: GateMate inverting CC_IOBUF disable has no effect
Inverting the input to the disable port of the CC_IOBUF macro using an inverting LUT1 has no effect on the buffer. While it should invert the T port of the IO Cell. See the minimal repro case:

```
	scl::Counter timer(5_b);
	Bit testPusle = timer.value().msb() | name("test_pulse");
	pinOut(testPusle, "TEST_ENABLE");

	{
		scl::arch::colognechip::CC_IOBUF buf;
		buf.pin("TEST0");
		buf.delayOut(15);
		buf.disable() = !testPusle;
		buf.pulldown(true);
		buf.O() = '1';
	}
	{
		scl::arch::colognechip::CC_IOBUF buf;
		buf.pin("TEST1");
		buf.delayOut(15);
		buf.disable() = testPusle;
		buf.pulldown(true);
		buf.O() = '1';
	}
```
Both buffers are instantiated equal except for the inverted disable input. On the oscilloscope we can observe that both pins output an identical signal.
Yellow is the testPulse signal, magenta is the non inverted output and cyan is the inverted output. The pulldown resistor is used to clearly identify the tristated and driven phases.

Image

The full vhdl repro project, including bitfile:

[HyperRAM_gatemate_io_disable_repro_case.zip](https://github.com/user-attachments/files/27394632/HyperRAM_gatemate_io_disable_repro_case.zip)

pu-cc:
In `pack_io_sel()`, the `is_inverted[2]` flag unconditionally overwrites `INIT_L10` with `LUT_INV_D0`, discarding any logic from an absorbed CC_LUT1 that drives the T (output-enable) port.

Fixes #1713.

### DIFF
--- a/himbaechel/uarch/gatemate/pack_io.cc
+++ b/himbaechel/uarch/gatemate/pack_io.cc
@@ -602,8 +602,11 @@ void GateMatePacker::pack_io_sel()
         Loc root_loc = ctx->getBelLocation(ci.bel);
         for (int i = 0; i < 4; i++) {
             CellInfo *cpe = move_ram_o_fixed(&ci, ctx->idf("OUT%d", i + 1), root_loc).first;
-            if (cpe && is_inverted[i])
-                cpe->params[id_INIT_L10] = Property(LUT_INV_D0, 4);
+            if (cpe && is_inverted[i]) {
+                uint8_t existing = int_or_default(cpe->params, id_INIT_L10, 0);
+                uint8_t inverted = (existing ^ 0xF);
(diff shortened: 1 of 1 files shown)


gatecat:
Thanks for fixing this one!

mmicko:
Thanks @pu-cc for fix, sorry for not taking care of this earlier, was busy creating releases

PR #1725, opened 2026-05-13
gowin: Enable driving FF XD from SEL

gatecat:
This depends on https://github.com/YosysHQ/apicula/pull/505 which also expands on the background a bit more


gatecat:
oops "DFFs have been given the ability to connect independently of the neighbouring LUT. Segment routing must be corrected." guess this needs to be fixed too

PR #1734, opened 2026-06-13
mistral: emit compressed bitstreams by default

jfabienke:
## Problem

`nextpnr-mistral` emits **uncompressed** RBF bitstreams by default (`--compress-rbf` opted into compression). An uncompressed bitstream *configures* the device — the socfpga FPGA manager reports `operating` — but on the FPGA-manager load path it leaves user IO non-functional.

Verified on a DE10-Nano (Cyclone V `5CSEBA6U23I7`, MiSTer loader):

- trivial fabric designs built with stock `nextpnr-mistral` (a `cnt[23]`→LED blinker, and a constant `assign LED = 8'hFF`) load but the LEDs stay **dark**;
- the **byte-identical** design re-emitted *compressed* lights/blinks the LEDs;
- Quartus also emits compressed bitstreams.

## Root cause

Entirely in `mistral/bitstream.cc::options()`. The default (uncompressed) branch set `COMPRESSION_DIS` **and** the `OPT_B` header word for the uncompressed format; the `--compress-rbf` branch already set the `OPT_B` value Quartus uses. So a working bitstream was always one flag away — the default was simply the non-functional one.

## Fix

Flip the default to compressed:

- new `--uncompressed-rbf` opt-out (kept for debugging / faster generation; documented that it may not load on all boards);
- `--compress-rbf` retained as a deprecated no-op so existing invocations don't break.

`OPT_B` already tracked the compression choice in-tree, so this is purely a default flip — no new magic values.

## Verification

Regenerating a design through the patched `nextpnr-mistral` with **no flags** produces bytes **identical** to a bitstream confirmed working on silicon (what the old `--compress-rbf` path emitted). `clang-format` clean; no functional change to the compressed path.

The deeper question of *why* the uncompressed encoding fails on the FPGA-manager path (a libmistral framing detail vs. a loader requirement) is still open, but the practical fix is unambiguous: compressed output matches Quartus and works on hardware, so it should be the default.


gatecat:
My understanding is that there are other situations where specifically non-compressed bitstreams are required (e.g. JTAG programming). In general this also seems like a better default for now (other arches bitstream tools also generally default to uncompressed), so I'm not going to merge this.

PR #1737, opened 2026-06-16
router2: add stall detection, iteration cap, and periodic validity checks

nkapreTT:
The negotiated-congestion router can spin without converging on
designs where (a) the architecturally optimal wire is genuinely
unavailable, (b) the cost gradient between two near-equivalent
candidates is too shallow to settle, or (c) a placer choice has
fenced off the legal binding space and the failure won't surface
until bind_and_check_all() actually runs.  Today the main loop has
no upper bound on iterations and bind_and_check_all() is only ever
called on full convergence, so a stuck design loops forever while
logging "archfail=NA" every iteration.

Three small, opt-in safety hatches:

1. Hard iteration ceiling -- `cfg.max_router_iters` (default 2000).
   When reached, log a warning, bind what we have, and break the
   loop instead of spinning.

2. Stall detection -- `cfg.stall_iter_limit` (default 50).  Watch
   (overused_wires, total_wire_overuse) across consecutive
   iterations.  When both stay unchanged for stall_iter_limit
   iterations, first try resetting `curr_cong_weight` back to
   `cfg.init_curr_cong_weight` (once) to escape a local minimum.
   If the stall persists after the reset, break with a
   best-effort binding rather than continuing to loop.

3. Periodic validity check -- `cfg.bind_check_interval` (default
   100).  Even when `overused_wires > 0`, call
   bind_and_check_all() every `bind_check_interval` iterations so
   architectural impossibilities surface early instead of being
   silently masked by `archfail=NA`.

All three are fully configurable via context settings
(`router2/maxRouterIters`, `router2/stallIterLimit`,
`router2/bindCheckInterval`).  Defaults preserve existing
behaviour for designs that converge promptly (well under 2000
iterations and never stall for 50 in a row), and the new break
paths only fire on pathological cases that would otherwise loop
forever.

No new options exposed at the CLI; defaults are sized for typical
designs and out-of-the-box behaviour is unchanged.

### DIFF
--- a/common/route/router2.cc
+++ b/common/route/router2.cc
@@ -1710,6 +1710,15 @@ struct Router2
         log_info("Running main router loop...\n");
         if (timing_driven)
             tmg.run(true);
+
+        // Stall-detection / hard-limit state for the main router loop.
+        // Configurable via cfg.stall_iter_limit / cfg.max_router_iters /
+        // cfg.bind_check_interval below.
+        int stall_count = 0;
+        int prev_overused = -1;
+        int prev_overuse = -1;
+        bool did_cong_reset = false;
+
         do {
             ctx->sorted_shuffle(route_queue);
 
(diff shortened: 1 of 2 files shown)


gatecat:
I'm curious, where are you seeing archfails? In most cases these are very inefficient and should be replaced with the use of resources anyway

PR #1740, opened 2026-06-19
gatemate: Initial hypergraph partitioner concept

gatecat:
This is nowhere near production ready, but it's a first step that opens up some interesting possibilities. It uses hypergraph partitioning to split the design to minimise cross-die routing before placement.

Currently there are several issues:

 - neither placer heap nor static handle the floorplans it produces very well. HeAP doesn't take into region constraints before strict legalisation, so strict legalisation becomes very slow and QoR collapses; static just sometimes doesn't converge with them but produces good enough results when it does
 - the current multilevel FM implementation is somewhat basic and doesn't give anywhere near optimum results for every seed (coarsening is likely a significant issue here)
 - the area constraints for the partitions are not correct and would cause problems if running close to the maximum
 - there's no ability for the placer to tweak the partitioning at all, which might sometimes make sense? 

Still, it opens up the possibility of doing further placer development to improve how floorplans are supported, so it's an interesting experiment.

### DIFF
--- a/common/place/placer_static.cc
+++ b/common/place/placer_static.cc
@@ -505,13 +505,33 @@ class StaticPlacer
         }
     }
 
-    RealPair limit_to_reg(Region *reg, RealPair val)
+    RealPair get_region_gradient(Region *reg, RealPair pos)
     {
+        RealPair val(0, 0);
         if (reg == nullptr)
             return val;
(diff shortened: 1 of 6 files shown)


Ravenslofty:
If this isn't production-ready, maybe it should be converted to a draft?

PR #1743, opened 2026-06-30
himbaechel: Merge all test entrypoints into one

gitRaiku:
This pull request fixes the build failure that happened when trying to build a unified himbaechel architecture with ng-ultra and gatemate enabled.
This was because the tests for the two architectures both had different main.cc files, which were then linked into the single unified test binary. Which of course failed.

This pulls the main.cc file up out of the uarch folder structure keeping the same one for every uarch.


gitRaiku:
And as an aside, as mentioned in https://github.com/YosysHQ/nextpnr/issues/1676, running the gatemate tests fail with
```
./himbaechel/uarch/gatemate/tests/lut.cc:107: Failure
Expected equality of these values:
  ctx->cells.size()
    Which is: 8
  9LU
    Which is: 9

[  FAILED  ] GateMateTest.remove_lut1_inv (30 ms)
```
Which i would also want to fix to finalise packing an updated version for nix, but i am not sure what the actual problem even is for the test.

gitRaiku:
Bump for this thread as it has been inactive for a month.

mmicko (approved):
LGTM

PR #1745, opened 2026-07-10
himbaechel/gatemate: fix multiplier-routing double-bind (w2n_entry assert) on dense designs

aignermax:
## Problem

On the Himbächel GateMate uarch, `nextpnr-himbaechel` aborts during the **"Routing multipliers"** pass with:

```
Assertion failure: w2n_entry == nullptr (common/kernel/base_arch.h:259)
```

i.e. `bindPip` is asked to bind a pip whose destination wire is already bound to another net. It is **seed-dependent** and shows up on **denser designs that use several `CC_MULT` blocks**; small designs (and a single multiplier in isolation) route fine, which is why it's easy to miss.

## Root cause

`himbaechel/uarch/gatemate/route_mult.cc` routes each multiplier by binding a **dedicated, hand-picked, `STRENGTH_LOCKED`** chain of pips whose wires are computed purely from the multiplier's placed `loc`. Some of those wires form a "halo" that reaches **outside the cluster body** — e.g. the A-passthrough column sits at `root.x - 1` and its switch-box routes (`SB_BIG/SB_SML.P0x…`) reach a tile or two further, and the `x4y1/x4y2` variants reach `loc.x + 3`.

Nothing reserves those halo wires during placement, and there is **no spacing rule between multiplier clusters**. So once the placer packs two multiplier clusters close enough that their halos overlap, the second cluster's `find_and_bind_downhill_pip` tries to bind a wire the first cluster already owns → the `bindPip` assertion.

Concretely, I reproduced a double-bind on wire `X24Y64/SB_SML.P05.Y1_int` between the `a_passthru_lower` routes of **two different** multiplier clusters.

## Fix

Two commits:

**1. `gatemate: prevent multiplier clusters from overlapping route_mult halos`** — the actual fix.
Add a placement-legality rule in `GateMateImpl::isBelLocationValid`: compute a multiplier cluster's halo bounding box (its placed cells' bbox, expanded by a margin that covers `route_mult`'s reach) and reject a placement whose halo overlaps another multiplier cluster's halo. The multiplier-cluster roots are cached once in `assign_cell_info` (via the existing `multiplier_*` pools) so the check stays cheap. This guarantees `route_mult`'s hand-routes are disjoint by construction — the same routes that already work for an isolated multiplier.

**2. `gatemate: diagnose multiplier routing conflicts instead of asserting`** — hardening.
Make `find_and_bind_downhill_pip` a no-op when the destination wire is already part of *this* net's route (a harmless revisit of a shared start wire, previously able to trip the same assert), and, if a *different* net holds the wire, emit an actionable `log_error` naming the wire and both nets instead of the opaque `NPNR_ASSERT`. This never silently drops a connection, and turns any residual/unforeseen collision into a diagnosable message.

## Verification

On a design (raycore compute core + monitor + UDP stack on a ULX5M-GS / CCGM1A1) that reproducibly hit the assert:

- Seeds that previously aborted in "Routing multipliers" (e.g. seed 4, seed 8) now **place and route** cleanly (core clock 26–28 MHz, timing met).
- A seed that already routed still routes — the rule is **not over-constraining** the placement.
- No seed hits the assert or the new `log_error` any more.

## Notes / open questions for maintainers

- The halo margin is a small fixed value (3 tiles), chosen conservatively to cover the reaches I could see in `route_mult.cc`. It could instead be derived exactly from each `route_mult_*` variant's reach if you'd prefer a tighter bound; I kept it simple and slightly generous (a too-large margin only costs a little fabric, a too-small one would let the crash slip through). Happy to adjust.
- The check recomputes cluster bboxes per call rather than caching positions (cells move during placement); multiplier clusters are few so this is cheap, but I can memoize if you'd like.

Found and fixed while bringing up a design on hardware; happy to iterate on the approach.

### DIFF
--- a/himbaechel/uarch/gatemate/gatemate.cc
+++ b/himbaechel/uarch/gatemate/gatemate.cc
@@ -208,6 +208,47 @@ bool GateMateImpl::isBelLocationValid(BelId bel, bool explain_invalid) const
         return true;
     }
 
+    // Multiplier clusters lay down dedicated, hand-picked LOCKED routing in
+    // preRoute (route_mult) that reaches into a halo of tiles around the placed
+    // cluster body. Nothing reserves those wires during placement and there is no
+    // spacing rule between clusters, so if two multiplier clusters are packed close
+    // enough their halos overlap and route_mult double-binds a shared wire (the
+    // base_arch.h bindPip assertion). Forbid any placement where the halo of this
+    // cell's multiplier cluster would overlap another placed multiplier cluster's
+    // halo. Placed first so it triggers for every cluster cell type (incl. the
+    // packed P-register FFs), not just the CPE_LT/L2T4 cells below.
+    if (!mult_cluster_roots.empty() && cell->cluster != ClusterId()) {
+        CellInfo *root = get_cluster_root(ctx, cell->cluster);
+        bool is_mult_cluster = false;
+        for (auto *r : mult_cluster_roots) {
+            if (r == root) {
+                is_mult_cluster = true;
+                break;
+            }
+        }
+        if (is_mult_cluster) {
+            int ax0, ay0, ax1, ay1;
+            if (get_mult_cluster_halo(root, ax0, ay0, ax1, ay1)) {
+                for (auto *other : mult_cluster_roots) {
+                    if (other == root)
+                        continue;
+                    int bx0, by0, bx1, by1;
+                    // Skip clusters not (fully) placed yet.
+                    if (!get_mult_cluster_halo(other, bx0, by0, bx1, by1))
+                        continue;
+                    bool disjoint = ax1 < bx0 || bx1 < ax0 || ay1 < by0 || by1 < ay0;
+                    if (!disjoint) {
+                        if (explain_invalid)
+                            log_info("Bel '%s' invalid: multiplier cluster halo overlaps another "
+                                     "multiplier cluster.\n",
+                                     ctx->nameOfBel(bel));
+                        return false;
+                    }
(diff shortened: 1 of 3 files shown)


whitequark:
I am strongly opposed to merging any PR generated by Claude.

himbaechel/uarch/gatemate/gatemate.cc
```diff
@@ -208,6 +208,47 @@ bool GateMateImpl::isBelLocationValid(BelId bel, bool explain_invalid) const
...
+    // enough their halos overlap and route_mult double-binds a shared wire (the
+    // base_arch.h bindPip assertion). Forbid any placement where the halo of this
+    // cell's multiplier cluster would overlap another placed multiplier cluster's
+    // halo. Placed first so it triggers for every cluster cell type (incl. the
+    // packed P-register FFs), not just the CPE_LT/L2T4 cells below.
+    if (!mult_cluster_roots.empty() && cell->cluster != ClusterId()) {
+        CellInfo *root = get_cluster_root(ctx, cell->cluster);
+        bool is_mult_cluster = false;
+        for (auto *r : mult_cluster_roots) {
+            if (r == root) {
```
gatecat, line 224: ignoring anything else this seems like a very poor way of determining if a cluster is a multiplier

gatecat:
yeah, the overall code quality here is pretty terrible too. maybe there is a genuine issue for @Ravenslofty as code owner of the gatemate multiplier support to look at but this PR isn't going anywhere towards a good quality, performant fix

PR #1750, opened 2026-07-16
ecp5: Improve use of dedicated ECLK routing

gatecat:
- Add an (overrideable) error if an ECLK is routed through fabric
 - Factor in ECLK sides when automatically placing PLLs


whitequark:
Thank you!

PR #1751, opened 2026-07-21
fabulous: add generic way to set BEL timings and setting placer/router tunables

EverythingElseWasAlreadyTaken:
Add bel.v3.txt, which carries per-arc BEL timing (Delay/SetupHold/ ClkToOut/Clock) with config-bit conditions, on top of bel.v2's structural blocks. Timing is applied per instance: a representative per-type estimate in prePlace, then each placed BEL's own arcs in postPlace (bel_timing_by_bel, keyed by BelId).

Add placement_estimate.txt carrying BEL timing estimates and nextpnr placer/router tunables (delayScale/delayOffset/delayEpsilon/ripupPenalty/ carryPredictDelay, each defaulting to nextpnr's historical hardcoded value).

A more detailed description of the FABulous model files can be found here: 
https://fabulous.readthedocs.io/en/latest/user_guide/using_doc/pnr/nextpnr.html#bel-v3-txt-bel-timing

bel.v3.txt and placement_estimate.txt were added as separate files to keep compatibility with older nextpnr versions and existing fabulous projects.


EverythingElseWasAlreadyTaken:
@KelvinChung2000 @mole99
Would be nice if you could also have a look on this.

generic/viaduct/fabulous/fabulous.cc
```diff
@@ -444,6 +591,178 @@ struct FabulousImpl : ViaductAPI
...
+    {
+        std::string s(v.m_ptr, v.m_length);
+        try {
+            return std::stof(s);
+        } catch (const std::exception &) {
+            log_error("invalid bel.v3 timing value '%s'\n", s.c_str());
+        }
+    }
+
+    static double parse_double(parser_view v)
```
KelvinChung2000: Maybe move some of these parse helper to an extra `helper.cc`?

KelvinChung2000 (commented):
I am just thinking is our bel timing is really that non-uniform that one cell is significantly differ from the other?

EverythingElseWasAlreadyTaken:
> I am just thinking is our bel timing is really that non-uniform that one cell is significantly differ from the other?

I don't know, I'd guess that back then, they just played around with the numbers until they got something that works well with the default `8` for pip timing.
But we'll figure out soon when the timing model generation is updated. :)

generic/viaduct/fabulous/fabulous.cc
EverythingElseWasAlreadyTaken: Actually decided to add a bel_timing.h/cc and move out all timing related functions from fabulous.cc, similar to how its done in fasm.cc and so on.

generic/viaduct/fabulous/bel_timing.h
```diff
@@ -0,0 +1,81 @@
+/*
+ *  nextpnr -- Next Generation Place and Route
+ *
+ *  Copyright (C) 2021-22  gatecat <gatecat@ds0.me>
```
KelvinChung2000: Header clean up

mole99 (approved):
Looks generally good to me. Thanks for the work!

KelvinChung2000 (approved):

EverythingElseWasAlreadyTaken:
@gatecat I would appreciate it if you or someone else from the nextpnr team could have a look on this PR. :)

PR #1754, opened 2026-07-24
ice40: Add noglobal attribute and raise promotion threshold

gatecat:
The iCE40 is generally generously provisioned in terms of routing resources, so factoring in insertion delay, there's no real benefit in such a low fanout threshold for globals. Also add the `noglobal` attribute to force disable promotion for a specific net.


whitequark:
Thank you! Will test in a moment.

whitequark:
The new default behavior solves the issue I was having independently of the seed. Setting a `noglobal` attribute deep in a design is actually somewhat tricky now that I'm considering doing it in practice, if we have to do that regularly in Glasgow, I'd have to come up with a way for applets to mark an inner signal with `noglobal` somehow.

Could we add an option like `--no-promote-global-ce`? It seems like CE is the most problematic one, seeing as high-fanout resets are usually domain resets (at least with Amaranth), but enables are inferred by Yosys so it's largely out of the frontend's control.

Issue #1617, opened 2025-12-15
[GateMate] `No Fmax available; no interior timing paths found in design.` when implementing multiplier


Jaspi10:
On the current build of oss-cad-suite (2025-12-15) the Verilog input:

```Verilog
module mult_test
    #(parameter N = 16)
    (
        input clk_i,
        input signed[N-1:0] a_i,
        input signed[N-1:0] b_i,
        output signed[2*N-1:0] p_o
    );

    reg signed[N-1:0] a_reg;
    reg signed[N-1:0] b_reg;

    wire signed[2*N-1:0] p_int;
    reg signed[2*N-1:0] p_reg;

    always @(posedge clk_i) begin
        a_reg <= a_i;
        b_reg <= b_i;
    end

    assign p_int = a_reg * b_reg;

    always @(posedge clk_i) begin
        p_reg <= p_int;
    end

    assign p_o = p_reg;

endmodule
```

When synthesized with the yosys like so:
```sh
yosys -p "read_verilog mult_test.v; synth_gatemate -json out/synth.json -luttree -nomx8; write_verilog out/synth.v"
```

And implemented with nexpnr like so:
```sh
nextpnr-himbaechel --device CCGM1A1 --json out/synth.json --write out/impl.json --vopt allow-unconstrained --router router2 --report out/report.json --detailed-timing-report
```

Shows
```
Info: Critical path report for cross-domain path '<async>' -> 'posedge $iopadmap$clk_i':
Info:       type curr  total name
Info:     source  0.00  0.00 Source $iopadmap$mult_test.a_i_6.Y
Info:    routing  0.00  0.00 Net $iopadmap$mult_test.a_i_6$conn$Y (61,133) -> (61,133)
Info:                          Sink $iopadmap$mult_test.a_i_6$iosel.GPIO_IN
Info:      logic  0.38  0.38 Source $iopadmap$mult_test.a_i_6$iosel.IN1
Info:    routing  8.58  8.96 Net $iopadmap$a_i[6] (61,133) -> (4,51)
Info:                          Sink $auto$ff.cc:337:slice$107$lt.IN1
Info:      logic  0.55  9.51 Source $auto$ff.cc:337:slice$107$lt.OUT
Info:    routing  0.00  9.51 Net $auto$ff.cc:337:slice$107$di (4,51) -> (4,51)
Info:                          Sink $auto$ff.cc:337:slice$107.DIN
[...]

PR #1755, opened 2026-07-24
gatemate: Fix DFF pin names when packing mult p registers

Jaspi10:
This fixes a bug similar to #1617.
My understanding is that b8a6559 switched `CLK` to `CLK_INT` and `EN` to `EN_INT` but missed the output register packing in `pack_mult.cc`. This results in the timing analyzer missing the paths to those FFs.


Ravenslofty (approved):
Makes me wonder how it worked in the first place...

PR #1764, opened 2026-07-31
himbaechel/xilinx: make depth-cascaded RAMB36E1 memories work

0xa000:
**A note on authorship**: the changes in this PR and their commit messages were written by an AI assistant (Claude), working in an interactive debugging session that I directed. I followed the analysis and have verified the results on hardware (a MEGA65 core port that boots to BASIC on a QMTech Wukong xc7a100t), but I am not deeply familiar with this codebase — please review accordingly rather than assuming expert authorship. Happy to answer questions, re-test on hardware, or provide additional evidence.

## Summary

Deep, narrow inferred memories (e.g. 64K x 1, which yosys maps to a depth-cascaded pair of RAMB36E1s, and 32K x 1 single-cell mode) were broken in several independent ways. This PR fixes them in five self-contained commits:

1. **fasm: emit `BRAM36_{READ,WRITE}_WIDTH_{A,B}_1`** — width-1 ports need the RAMB36-level feature in addition to the per-half RAMB18 bits; without it the two 16K halves alias (address LSB ignored).
2. **fasm: emit `RAM_EXTENSION_{A,B}_LOWER`** — the cascade configuration of the LOWER cell was never written (UPPER shares the NONE encoding, so only LOWER needs a bit).
3. **pack: don't override driven `ADDRARDADDRL15`/`ADDRBWRADDRL15`** — these were unconditionally tied to VCC, but in a cascaded pair they carry the address MSB selecting the half; tying them pinned all accesses to the upper 32K. Tie only when undriven.
4. **pack/arch: place cascade pairs as clusters** — CASCADEOUT→CASCADEIN are dedicated wires reaching only the vertically adjacent RAMB36 site. Bel z indices are not uniform across BRAM tiles, so relative (x,y,z) constraints can't express "next RAMB36 up the column"; a precomputed next-in-column map plus a `getClusterPlacement` override does.
5. **arch: give RAMB36 bels a nonzero area in the static placer** — `bel_area` was 0x0 while cells claim 1x6, so BRAM-heavy designs couldn't converge to a legalisable placement.

## Repro

Infer a 64K x 1 (cascaded pair) or 32K x 1 (single cell, commits 1-2) memory, e.g. `reg mem[0:65535]` with registered read, initialised with a known pattern; read it back. Before this PR: placement may fail outright (commit 4), and the read data aliases or comes from the wrong half.

## Evidence

Verified against Vivado bitstreams for the same configurations, and on xc7a100t hardware: a 128K ROM built from cascaded RAMB36E1 pairs passes a full content fingerprint check (32/32), as part of a MEGA65 core port (93% BRAM utilisation) that boots to BASIC.

### DIFF
--- a/himbaechel/uarch/xilinx/pack.cc
+++ b/himbaechel/uarch/xilinx/pack.cc
@@ -754,14 +754,19 @@ void XC7Packer::pack_bram()
                 ci->connectPort(p, ctx->nets[ctx->id("$PACKER_VCC_NET")].get());
             }
         } else if (ci->type == id_RAMB36E1_RAMB36E1) {
+            // ADDRARDADDRL15/ADDRBWRADDRL15 must be tied high in non-cascaded
+            // modes, but in a depth-cascaded pair (RAM_EXTENSION_A/B = LOWER/
+            // UPPER) they carry the address MSB that selects the half; the
+            // multixform has already wired ADDRARDADDR[15] onto them, and
+            // overriding that with VCC pins both reads and writes to the upper
+            // 32K half. Only tie when not driven.
             for (auto p : {id_ADDRARDADDRL15, id_ADDRBWRADDRL15}) {
                 if (!ci->ports.count(p)) {
                     ci->ports[p].name = p;
                     ci->ports[p].type = PORT_IN;
-                } else {
-                    ci->disconnectPort(p);
                 }
-                ci->connectPort(p, ctx->nets[ctx->id("$PACKER_VCC_NET")].get());
+                if (ci->getPort(p) == nullptr)
+                    ci->connectPort(p, ctx->nets[ctx->id("$PACKER_VCC_NET")].get());
             }
             if (int_or_default(ci->params, id_WRITE_WIDTH_A, 0) == 1) {
                 ci->disconnectPort(id_DIADI1);
--- a/himbaechel/uarch/xilinx/pack.cc
+++ b/himbaechel/uarch/xilinx/pack.cc
@@ -808,6 +808,42 @@ void XC7Packer::pack_bram()
             }
         }
     }
+
+    // Cluster cascaded RAMB36 pairs: CASCADEOUT->CASCADEIN are dedicated
+    // wires that only reach the vertically adjacent RAMB36 site, so the
+    // pair must be placed together (cf. carry chains). BRAM tiles are 5
+    // grid rows apart within a column.
+    int cascade_pairs = 0;
+    for (auto &cell : ctx->cells) {
+        CellInfo *lower = cell.second.get();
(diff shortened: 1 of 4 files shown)


himbaechel/uarch/xilinx/pack.cc
```diff
@@ -803,6 +808,42 @@ void XC7Packer::pack_bram()
...
+    int cascade_pairs = 0;
+    for (auto &cell : ctx->cells) {
+        CellInfo *lower = cell.second.get();
+        if (lower->type != id_RAMB36E1_RAMB36E1)
+            continue;
+        for (auto pn : {ctx->id("CASCADEOUTA"), ctx->id("CASCADEOUTB")}) {
+            NetInfo *cn = lower->getPort(pn);
+            if (cn == nullptr)
+                continue;
+            for (auto &usr : cn->users) {
```
gatecat: this should also check that the port only has one user, and that user is the RAM being cascaded, otherwise things are illegal

himbaechel/uarch/xilinx/xilinx.cc
```diff
@@ -397,7 +440,7 @@ void XilinxImpl::configurePlacerStatic(PlacerStaticCfg &cfg)
         comb.cell_area[id_RAMB18E1_RAMB18E1] = StaticRect(1.0f, 3.0f);
         comb.bel_area[id_RAMB18E1_RAMB18E1] = StaticRect(1.0f, 3.0f);
         comb.cell_area[id_RAMB36E1_RAMB36E1] = StaticRect(1.0f, 6.0f);
-        comb.bel_area[id_RAMB36E1_RAMB36E1] = StaticRect(0.0f, 0.0f);
+        comb.bel_area[id_RAMB36E1_RAMB36E1] = StaticRect(1.0f, 6.0f);
```
gatecat: this part of the PR is definitely not correct. placeable area is dealt with per-group, so this would double-count the RAMs as the 18s and 36s overlap (equally, it's not needed as sufficient area is given by the 18s already). probably something else is going wrong in the placer if you need to tell it that twice the RAM area as there actually is exists

whitequark:

0xa000:
Both points addressed, thanks.

On the port check: the clustering now validates the cascade topology and errors out instead of silently mis-clustering. A CASCADEOUT net must have exactly one load, that load must be the corresponding CASCADEIN of another RAMB36E1, and chains deeper than two (which cannot exist on xc7, RAM_EXTENSION is only LOWER/UPPER) are rejected.

On the bel area: you're right, and I've dropped that commit. I re-ran the design that motivated it (a MEGA65 core port, ~93% BRAM utilisation, 48 cascaded pairs) on current master plus the remaining commits of this PR: placement legalises promptly and the design routes without the area change.

PR #1765, opened 2026-07-31
himbaechel/xilinx: fix LUTRAM initialisation (INIT_A..INIT_D)

0xa000:

## Symptom

All distributed RAM (RAM32M/RAM64M etc.) is configured with INIT=0 regardless of the design's initial contents.

## Root cause and fix

The primitives carry per-port initialisation in parameters named `INIT_A..INIT_D`, but `pack_dram` looked up `INITA..INITD` — the lookup never matched (one-character fix). Fixing the name exposed a second, previously unreachable bug in the wide (dbits>1) path: the 64-bit INIT slice was converted with `as_bits()`, which yields raw bool values rather than the '0'/'1' characters a `Property` string requires, producing a corrupt property. The commit de-interleaves the canonical bit string directly and rebuilds the Property with `update_intval()`.

## Repro

A RAM64M/RAM32M with nonzero `INIT_A..D` and its outputs on LEDs (WE driven by a live net, held low). Before: all outputs read 0; after: the INIT contents.

## Evidence

Hardware-verified on xc7a100t: the repro above, plus LUTRAM-based tables in a MEGA65 core port that boots to BASIC.

### DIFF
--- a/himbaechel/uarch/xilinx/pack_dram.cc
+++ b/himbaechel/uarch/xilinx/pack_dram.cc
@@ -495,8 +495,8 @@ void XilinxPacker::pack_dram()
                                                      dout, zoffset + i);
                     if (base == nullptr)
                         base = dram;
-                    if (ci->params.count(ctx->idf("INIT%c", 'A' + i)))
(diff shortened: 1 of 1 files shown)


whitequark:

PR #1766, opened 2026-07-31
himbaechel/xilinx: do not set OUT_DIFF for TMDS_33 outputs

0xa000:

## Symptom

HDMI/DVI output via TMDS_33 differential pairs is completely dead; the attached monitor reports "no transmitter present".

## Root cause and fix

TMDS_33 outputs use the Y0 current-mode driver alone (`TMDS_33.OUT` plus `DRIVE.I_FIXED`, with the Y1 pad set IN_ONLY). Vivado bitstreams for the same design do not set the tile-level `OUT_DIFF` feature for TMDS pairs, and setting it disables the driver stage entirely. The fasm backend set `OUT_DIFF` for every differential output; guard it so TMDS_33 is excluded (other diff standards, e.g. LVDS_25, keep it).

## Evidence

A/B on a 640x480 DVI design for the QMTech Wukong (xc7a100t): without the fix the emitted fasm contains `OUT_DIFF` on all four TMDS tiles and the screen stays dark; with it, no `OUT_DIFF` (matching Vivado) and the monitor shows a perfect test card. Also exercised by the HDMI path of a MEGA65 core port on the same board.

### DIFF
--- a/himbaechel/uarch/xilinx/fasm.cc
+++ b/himbaechel/uarch/xilinx/fasm.cc
@@ -1004,7 +1004,11 @@ struct FasmBackend
         else
             inv = uarch->get_site_bel(pad_bel_site, ctx->id("IOB33S.O_ININV"));
 
-        if (inv != BelId() && ctx->getBoundBelCell(inv) != nullptr)
+        // TMDS_33 outputs are driven by the Y0 current-mode driver alone
+        // (TMDS_33.OUT + DRIVE.I_FIXED, Y1 IN_ONLY); Vivado does not set
+        // OUT_DIFF for them and setting it disables the driver stage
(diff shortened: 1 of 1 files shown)


whitequark:

gatecat:
This might be an Artix-7 vs Kintex-7 difference, as TMDS output was definitely working for me on a Genesys2.

Do you have the input design, bitstream and FASM from your Vivado test to look at?

0xa000:
This seems to have been a false positive. Several issues where in play at once. Re-doing the A/B test shows stable test card with and without OUT_DIFF.  Its functionally harmless here, even though Vivado never emits it for TMDS_33 on this part.

Do you want me to close the PR, or keep it for fasm parity with Vivado?

Lunaphied:
Closing for now

PR #1767, opened 2026-07-31
himbaechel/xilinx: fix MMCM loop filter and power register programming

0xa000:

## Symptom

MMCME2_ADV never asserts LOCKED on hardware; all derived clocks stay dead. (PLLE2 was unaffected.)

## Root cause and fix

The fasm writer put the 10-bit BANDWIDTH loop-filter word into `FILTREG1_RESERVED[11:0]` and wrote a hardcoded `0x3d4` into `TABLE[9:0]`. In prjxray's MMCM model, `TABLE[9:0]` *is* the loop-filter table word; with the filter word misplaced the analog loop is misconfigured and never locks. The fix swaps them (`TABLE = filter_lookup[CLKFBOUT_MULT-1]`, `FILTREG1_RESERVED = 0x8`) and programs `POWER_REG` from the actual configuration (`0x0100` integer-only, `0x9900` when CLKOUT0/CLKFBOUT use fractional divide) instead of `0xffff` — all matching Vivado bitstreams for the same configurations (verified for integer and `CLKOUT0_DIVIDE_F=6.5` fractional testcases).

## Evidence

Hardware-validated on a QMTech Wukong (xc7a100t): an MMCM blinky that never locked before locks and blinks after; the fix also carries a 640x480 DVI design and a MEGA65 core port (both MMCM-clocked) on the same board. A/B fasm diff shows exactly the TABLE/FILTREG1_RESERVED/POWER_REG features changing, with the same 10-bit filter word moving from FILTREG1_RESERVED to TABLE.

### DIFF
--- a/himbaechel/uarch/xilinx/fasm.cc
+++ b/himbaechel/uarch/xilinx/fasm.cc
@@ -1691,14 +1691,23 @@ struct FasmBackend
             filter_lookup = Xc7MMCM::filter_lookup_high;
         else
             filter_lookup = Xc7MMCM::filter_lookup_optimized;
-        write_int_vector("FILTREG1_RESERVED[11:0]", filter_lookup[clkfbout_mult - 1], 12);
+        // prjxray's TABLE[9:0] IS the 10-bit loop-filter word; putting the
+        // filter word in FILTREG1_RESERVED with a hardcoded TABLE=0x3d4
+        // (as before) yields an MMCM that never locks on hardware.
+        // Values verified against Vivado bitstreams (integer and
+        // CLKOUT0_DIVIDE_F=6.5 fractional testcases) and hardware-validated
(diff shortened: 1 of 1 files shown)


whitequark:

gatecat:
I'd prefer leaving the power stuff out of here, just in case it happens there's a case that needs another bit enabled that we've missed (as I understand that's just an optimisation found on the side, it's only the table part that's needed to get the design to lock?). otherwise, once again, give this a less verbose comment and it's good to go.

0xa000:
Dropped the POWER_REG change and reworded the comment. I re-verified on hardware: the MMCM blinky on the Wukong board (xc7a100t) locks and blinks at the expected 1 Hz with the TABLE/FILTREG1 fix applied and POWER_REG left at 0xffff.

PR #1772, opened 2026-08-05
Accept UTF-8 filenames and terminal I/O on Windows

whitequark:
See https://github.com/YosysHQ/yosys/pull/6087.

DO NOT MERGE until the Yosys PR is in.


whitequark:
The PR does work for both console IO and command line parsing, but `src` attribute printing is broken somewhere. I didn't debug this at this time, as I have higher priorities to tackle.

whitequark:
Ready and well tested now.

PR #1773, opened 2026-08-07
himbaechel/gui: Fix segfault when using "Open json" in gui

Jaspi10:
This PR fixes a segfault occuring when using "Open json" in the gui.
The PR is split into three parts for easier review, but should probably be squashed before merging.

The three commits fix three separate (but related) issues:
c7990e8 fixes a null pointer deref in `isPipInverting()` due to `ctx->uarch` never being initialized when creating a new context during json load. This is also the issue in the stack trace posted issue #1625.

8b1976b fixes a double free error due to concurrent access to a hashmap by both the `FPGAViewWidget` and `import_module`.

6b583cb fixes another null pointer deref on `ctx->uarch`. Due to `FPGAViewWidget` holding a copy of the context, which is deleted by the `unique_ptr` assignment.

Please review carefully as I am not very familiar with the GUI code. I have also not tested this on any architecture except himbaechel + gatemate.

Fixes #1625

Implemented design I used to test:  [impl.json](https://github.com/user-attachments/files/30827244/impl.json)

### DIFF
--- a/frontend/frontend_base.h
+++ b/frontend/frontend_base.h
@@ -132,6 +132,9 @@ template <typename FrontendType> struct GenericFrontend
         m.is_toplevel = true;
         m.prefix = "";
         m.path = top;
+
+        std::lock_guard<std::mutex> lock(ctx->mutex);
+
         ctx->top_module = top;
         // Do the actual import, starting from the top level module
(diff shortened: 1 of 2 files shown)


mmicko (approved):
LGTM. Confirmed working fine on non-himbaechel archs as well.

PR #1776, opened 2026-08-12
gatemate: Also use clock router for RAM clocks

gatecat:
While poking around I discovered that router2 was regularly taking a significant proportion of time (sometimes 10 seconds out of 30) to route these, and it seems like there's no harm getting the clock router to do it much quicker instead.

### DIFF
--- a/himbaechel/uarch/gatemate/route_clock.cc
+++ b/himbaechel/uarch/gatemate/route_clock.cc
@@ -56,6 +56,11 @@ void GateMateImpl::route_clock()
         return (ctx->getBelBucketForCellType(port.cell->type) == id_CPE_FF) && port.port.in(id_CLK_INT);
     };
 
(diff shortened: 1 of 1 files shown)


gatecat:
example routing, I believe this is the same as router2 was doing

```
Info:         routing arc to storage_7.0.0$full.CLOCK1 (wire X131Y115/RAM.CLOCK1):
Info:             bind pip X131Y115/RAM.CLOCK1/RAM.CLOCK1_n --> X131Y115/RAM.CLOCK1 (plane 0)
```

mmicko (approved):

mmicko:
Please do check variant where there are multiple clocks, and when it is used on A2 as well. Just to be on safe side.

Running complete set of prjpeppercorn-test-cases should cover most of use cases I guess.

PR #1777, opened 2026-08-13
gatemate: Reserve all wires one level further uphill if needed

gatecat:
Previously, this was reserving the path via CINY2 (as the first pip it happened to see) but not the path via CLK and plane 9 that's actually used. This meant that the clock router could use plane 9 in a tile as part of the routing for a previous clock, causing a later routed clock in that tile to be unrouteable.


Ravenslofty (approved):
prjpeppercorn-test-cases run looks good; stress on 098 goes from 21/100 to 94/100.

mmicko (approved):

PR #1783, opened 2026-08-21
fabulous: Drop legacy project support, add per flop arrival times, make pip delay scale settable

EverythingElseWasAlreadyTaken:
- Drop the legacy project format and bel.v1.
- Read each flop's arrival time from an optional delay field on the bel.v3 GlobalClk command
- Warn when a bel.v3 times only some clock pins
- Make pips.txt delay scale settable via a new pipDelayScale key


EverythingElseWasAlreadyTaken:
@KelvinChung2000 please have a look on this.

KelvinChung2000 (approved):

PR #1791, opened 2026-09-04
#1628 ice40, gowin: handle singleton vector [0] in PCF and CST constraints

Shriyans406:
In HDL designs, 1-bit ports are frequently declared as singleton vectors like output wire [0:0] LEDS. Yosys synthesizes these into scalar cell names like LEDS, but user constraint files commonly specify indexed pin names like set_io LEDS[0] 45 in .pcf or IO_LOC "LEDS[0]" 15; in .cst.

Previously, nextpnr-ice40 and nextpnr-himbaechel (gowin) failed to match LEDS[0] against scalar LEDS, emitting unmatched constraint warnings or unconstrained IO errors.

This PR applies the same fallback resolution pattern already implemented for ECP5 and MachXO2 to both ice40/pcf.cc and himbaechel/uarch/gowin/cst.cc.

### DIFF
--- a/himbaechel/uarch/gowin/cst.cc
+++ b/himbaechel/uarch/gowin/cst.cc
@@ -145,8 +145,16 @@ struct GowinCstReader
                     }
                 }
 
-                IdString net = ctx->id(match[1]);
+                std::string net_name = match[1];
+                IdString net = ctx->id(net_name);
                 auto it = ctx->cells.find(net);
+                // 1-bit wires are treated as scalar by nextpnr.
+                // In HDL they might have been a singleton vector.
+                if (it == ctx->cells.end() && net_name.size() >= 3 && net_name.substr(net_name.size() - 3) == "[0]") {
+                    net_name = net_name.substr(0, net_name.size() - 3);
+                    net = ctx->id(net_name);
+                    it = ctx->cells.find(net);
+                }
                 if (cst_type != clock && cst_type != adc && it == ctx->cells.end()) {
                     log_info("Cell %s not found\n", net.c_str(ctx));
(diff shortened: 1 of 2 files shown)


bba/CMakeLists.txt
```diff
@@ -1,7 +1,17 @@
...
+set(CMAKE_CXX_COMPILER_VERSION "16.2.0") 
+set(CMAKE_C_COMPILER_VERSION "16.2.0") 
+
+# Intercept and stub the target_compile_features command specifically for bba 
+macro(target_compile_features) 
+    # Do nothing here so it won't crash when looking for "cxx_std_11" or "cxx_std_17" 
+endmacro() 
+
+# Globally force C++17 flags since CMake doesn't know how to map them for GCC 16 
+add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-std=c++17>)
```
whitequark: I don't think any of this code should be added in this PR (or, most likely, at all).

whitequark (changes requested):
Unrelated changes in `bba/CMakeLists.txt`.

Shriyans406:
Respected Sir/Madam

I have perfomed the requested changes on bba/CMakeLists.txt.
Please do check it.

Regards

Lunaphied (changes requested):
I appreciate the fix! But with having this replicated in so many locations, could you please abstract this out to common logic on `Context`, specifically `BaseCtx`? I think something like `getNetForConstraintName` would work. You can add the definition to `common/kernel/basectx.cc` and the declaration in it's header.

EDIT: updated based on discussion

PR #1795, opened 2026-09-08
Himbaechel Xilinx: Fix inaccuracy at print of VCO frequency

marzoul:
Print of VCO frequency was wrong due to implicit cast from double to int (delay_t).
For a VCO period of 810.81 ps this was rounded to 810, thus printing 1234.6 instead of 1233.3 MHz.

Additional change : print 2 decimals instead of just 1 for PLL frequencies, for consistency with the rest of nextpnr, and also because this detail really matters when dealing with PLLs.


gatecat:
Thanks!

PR #1805, opened 2026-09-21
Himbaechel Xilinx : Fix packing dram reconnecting net driver

marzoul:
Fix assert failure duxing Himbaechel Xilinx packing of DRAM (LUTRAM).
Reconnection of net driver was missing checks for non-null net and also was assigning a new driver without prior disconnecting previous driver.


gatecat:
Do you have the specific case this was failing on? I just want to make sure it's not hiding any other issues

marzoul:
Yes this is a design stuffed with memories, for board Zedboard, synthesized by Ghdl then Yosys.
Public loud link : https://cloud.univ-grenoble-alpes.fr/s/xwGyEjYfdWAKnGe
(json.gz file, 33MB)

xdc contents if necessary :
```
create_clock -name fclk0_clock -period 10.00 [get_nets {FCLKCLK[0]}]
set_property PACKAGE_PIN Y9 [get_ports {gclk}]
set_property IOSTANDARD LVCMOS33 [get_ports {gclk}]
create_clock -name gclk_clock -period 10.00 [get_ports gclk]
```

Command line :
```
nextpnr-himbaechel -l nextpnr.log --device xc7z020clg484-1 -o xdc=xdc/zedboard.xdc --json zedboard_nnawaq_axi3.json -o fasm=zedboard_nnawaq_axi3.fasm --router router2
```

I investigated based on experienced crashes.

My intrepretation is that It is not illegal to have unconnected outputs of RAM macros, at least in these 2 cases :
- muti-port memory with addresses different than Write address, so LUTD is used exclusively as Write port and the Read is unused,
- 32-line memories that use the 2 LUT outputs, and the output is used is not a multiple of 2 or does not fully use the slice.

marzoul:
Currently that design can't fit the device because it exceeds the number of DSPs, it's not a problem here because the placer fails legitimately after the packer.
(I will have to fight these pesky DSPs separately with Yosys)

gatecat:
Thanks for the testcase! `connectPort` will already skip on nullptr nets, and I decided to fix the reconnecting driver issue by making RAM64X1S packing consistent with the other paths that already disconnect the output pins before creating the new LUT cell: https://github.com/YosysHQ/nextpnr/pull/1809

Can you confirm that this works for you?

marzoul:
@gatecat Your fix is enough for my design to pass !

PR #1806, opened 2026-09-21
heap: Use a vector for cell_locs

gatecat:
Only a 3-4% improvement or so but still worth it


Ravenslofty (commented):
I don't understand heap well enough to approve this, even without present COVID brain, but I do smell the router2 influence in this patch :p

